PR #24, opened 2014-11-07
ARM GPU modifications

marcino239:
Add function to check if local memory is present.  Without this check the OpenCL code generation fails on systems without local memory.


abergeron:
This has been on here for over 6 months and I heard nothing back.  Closing for now.

PR #33, opened 2015-01-15
Maintenance changes

abergeron:


nouiz:
I'm good with that, but I didn't run tests. Can you confirm you ran it? If so, you can merge.

Issue #34, opened 2015-02-16
pygpu.gpuarray.GpuArrayException: Invalid value


pannous:
pygpu.init can't find Radeon GPU card:

>>> pygpu.init('opencl0:0').devname 
u'Intel(R) Core(TM) i7-2600 CPU @ 3.40GHz' // CPU ok
>>> pygpu.init('opencl0:1').devname 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "gpuarray.pyx", line 587, in pygpu.gpuarray.init (pygpu/gpuarray.c:7113)
  File "gpuarray.pyx", line 558, in pygpu.gpuarray.pygpu_init (pygpu/gpuarray.c:7051)
  File "gpuarray.pyx", line 962, in pygpu.gpuarray.GpuContext.__cinit__ (pygpu/gpuarray.c:10398)
pygpu.gpuarray.GpuArrayException: Invalid value

opencl0:1 is valid in pyopencl:
pyopencl.Device 'ATI Radeon HD 6970M' on 'Apple' at 0x1021b00

>>> pygpu.init('opencl0:2').devname
ValueError: No device 2  // OK, only 0 and 1 are valid devices

Is this card not supported or did something go wrong during installation/compilation?

abergeron:
Can you try the master now?  It should be fixed, by working around AMD not following the standard.

abergeron:
This should be fixed by #36

PR #36, opened 2015-02-19
Fix OpenCL CL_DRIVER_VERSION query from causing GPU selection to fail.

paolodedios:
On some OpenCL GPUs, an exception, "pygpu.gpuarray.GpuArrayException: Invalid value", is thrown when initializing a GpuContext.

$ python -c "import pygpu; print pygpu.init('opencl0:1').devname"
  File "<string>", line 1, in <module>
  File "pygpu/gpuarray.pyx", line 587, in pygpu.gpuarray.init (pygpu/gpuarray.c:7506)
  File "pygpu/gpuarray.pyx", line 558, in pygpu.gpuarray.pygpu_init (pygpu/gpuarray.c:7435)
  File "pygpu/gpuarray.pyx", line 962, in pygpu.gpuarray.GpuContext.__cinit__ (pygpu/gpuarray.c:11098)
pygpu.gpuarray.GpuArrayException: Invalid value

This seems to occur because the driver_version buffer is not big enough to accommodate long device driver string signatures. 

When the following OpenCL function is called

```C
  err = clGetDeviceInfo(id, CL_DRIVER_VERSION, sizeof(driver_version),
                        driver_version, NULL);
  if (err != CL_SUCCESS)
    return NULL;
```

It fails and exits the entire GpuContext initialization routine.

With my GPU, an AMD Radeon HD 7950, the driver version is listed as "1.2 (Aug 17 2014 20:27:52)", which is a combination of the driver version and device version strings, and it overflows the current 16 byte buffer.

This PR is a simple fix that increases the size of the driver_version string and it should fix problems reported in issue #34.

### DIFF
--- a/src/gpuarray_buffer_opencl.c
+++ b/src/gpuarray_buffer_opencl.c
@@ -49,7 +49,7 @@ cl_ctx *cl_make_ctx(cl_context ctx) {
   cl_device_id id;
   cl_command_queue_properties qprop;
   char vendor[32];
-  char driver_version[16];
+  char driver_version[64];
   cl_uint vendor_id;
   size_t len;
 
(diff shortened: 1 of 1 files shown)


abergeron:
God damn it.  This is out of spec.  But I'll still merge because that's not a big change.

PR #41, opened 2015-04-24
Make extcopy work properly with float16 data.

abergeron:


src/gpuarray_buffer_cuda.c
```diff
@@ -1085,13 +1085,21 @@ static inline int gen_extcopy_kernel(const cache_key_t *a,
...
-  const char *out_t;
+  const char *in_t, *in_ld_t;
+  const char *out_t, *out_ld_t;
   const char *rmod;
   char arch[6]; /* Must be at least 6, see detect_arch() */
 
   in_t = map_t(a->itype);
   out_t = map_t(a->otype);
+  if (a->itype == GA_HALF)
+    in_ld_t = "b16";
```
nouiz: Can you put a comment why this is needed? Why f16 do not work and what b16 mean?

abergeron: Yes, but the general thing is that f16 is not a real type and can't be used for most operations, so we have to fake it with b16 (which is like uint16). I have to do this explicitly since we are using PTX directly. Normally the cuda compiler would do that for us.

PR #42, opened 2015-04-29
Make .gpudata actually work as expected

abergeron:


abergeron:
I've updated the doc and added a comment.

I also remove the scan.py file since it's old compyte code that doesn't work with libgpuarray anymore.

PR #47, opened 2015-05-21
Fix incorrect calculation of axis length in reshape

ejls:


nouiz:
Normally we need tests. Do this make an existing test fail? Can you add one in pygpu/tests/test_gpu_ndarray.py. There is already tests for reshape there, can you add a case that cause the old code to fail?

ejls:
I've amended the commit with a test case on which the old code fail.

PR #49, opened 2015-05-21
Fix some tests with newer NumPy and import problem

nouiz:


abergeron:
I'll trust you on the flags change and the rest seems ok.

PR #51, opened 2015-05-27
Fix problem with extcopy on Titan X

abergeron:
There are also a couple of changes to make the code more amenable to running on python 3.


abergeron:
I'm not against documenting that somewhere, but the installation document is not the right place.

You don't even need a GPU in the system to install it so it's not an installation requirement.

nouiz:
I'm running tests, I'm waiting for them to merge, but otherwise, I'm ok with the PR.

Issue #55, opened 2015-06-17
gcc 5.1.1 : max_align_t also defined in stddef.h


mdda:
There's a conflicting redefinition going on for ``max_align_t`` between : 

* ``/home/andrewsm/env/libgpuarray/src/util/halloc.c`` ; and
* ``/usr/lib/gcc/x86_64-redhat-linux/5.1.1/include/stddef.h``

It looks like it could be resolved by having ``#ifndef _GCC_MAX_ALIGN_T" around the definitions in our ``halloc.c``.

Does this make sense?

nouiz:
That seem good. Do you want to make a PR?

By the way, which OS do you use?

thanks

mdda:
I'm on Fedora (which tends to be kind of bleeding edge - on FC22 now).

I'll do a PR this weekend, since I'm on a bit of a PyCon-talk prep deadline - and now find that Nvidia doesn't like the gcc version either : So the whole "run some GPU demo" thing is circling the drain...

I just spent a bunch of time playing with blocks inside iPython (with Bokeh) - so facing a compiler message quickly brought me back down to earth :-)

PR #56, opened 2015-06-21
Prevent redefinition of max_align_t

mdda:
This addresses issue #55 (and fixes a couple of typos, and cures a spurious warning).

### DIFF
--- a/pygpu/gpuarray.pyx
+++ b/pygpu/gpuarray.pyx
@@ -594,7 +594,7 @@ def init(dev):
     For opencl the device id is the platform number, a colon (:) and
     the device number.  There are no widespread and/or easy way to
     list available platforms and devices.  You can experiement with
-    the values, unavaiable ones will just raise an error, and there
+    the values, unavailable ones will just raise an error, and there
     are no gaps in the valid numbers.
     """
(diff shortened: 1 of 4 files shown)


rmsc:
I just proposed a similar pull, failed to notice this one.

First of all, I would avoid mixing unrelated changes in the same commit (that is, the max_align_t thing and the unrelated typos).

Second, after the ifndef you should insert a define, just as it's done on the stddef.h header. Otherwise if the file is included more than once, you'll get an error.

mdda:
I'll add the extra define if you like, but note that this is in a .c file, so no-one is including it anywhere...

If you don't like the typo fixes, then I'll resubmit from a blank slate (after all, this is a 2 line fix for the max_align_t issue).  The typos would then have to wait (which, apparently, is a thing).

abergeron:
I think this is all fine, but the PR needs a rebase (for some reason).

abergeron:
Actually it seems that PR #58 already fixes the max_align_t problem and that is the conflict.

abergeron:
The max_align_t issue has been fixed for a while, so I'll close this PR.

PR #58, opened 2015-07-08
Fix compilation with C11 or C++11 enabled (default on GCC 5).

rmsc:
GCC 5 enables C11 by default, and stddef already creates struct max_align and max_align_t.

In order to compile libgpuarray you can either disable C11/C++11 on the CMakeLists (undesirable IMHO), or avoid redeclaring the struct if it has already been declared. This patch does just that. I decided not to remove the struct declaration altogether, in order to keep the support for older versions of the standard.


nouiz:
thanks for the commit and explanation.

PR #59, opened 2015-07-10
Two small fixes

davidweichiang:
Hi, here are two small changes:
- in gpuarray.pyx, allow arrays to be created like empty(shape, dtype=numpy.float32).
- in tools.py, fix a bug that was throwing an exception if a binary function had a scalar and a vector argument.


abergeron:
This looks good, especially the fix in tools.py.

PR #60, opened 2015-07-26
remove warning about unused variable tmp

davidweichiang:
Hi, another small fix -- one of the kernels was causing nvcc to give a warning about tmp being declared but not used. Thanks!


abergeron:
Looks good, thanks!

PR #61, opened 2015-07-27
Two fixes and a suggestion for check_args()

davidweichiang:
Hi, the first two commits fix bugs in check_args that are described in the commit messages. The third commit is just a suggestion: this function computes information that never gets used in the contiguous case (which is presumably also the most common), so a simplified version check_contig only computes what is needed, and is faster.

### DIFF
--- a/pygpu/elemwise.py
+++ b/pygpu/elemwise.py
@@ -2,7 +2,7 @@ from mako.template import Template
 
 import numpy
 
-from .tools import ScalarArg, ArrayArg, as_argument, check_args, lfu_cache
+from .tools import ScalarArg, ArrayArg, as_argument, check_contig, check_args, lfu_cache
 from .dtypes import (parse_c_arg_backend, dtype_to_ctype, get_np_obj,
                      get_common_dtype)
 from . import gpuarray
@@ -428,11 +428,13 @@ class ElemwiseKernel(object):
         return k, args
 
     def select_kernel(self, args, collapse=None, broadcast=False):
+        n, offsets, contig = check_contig(args)
+        if contig:
+            return (self.contig_k, self.prepare_args_contig(args, n, offsets)), n
+
         n, nd, dims, strs, offsets, contig = check_args(args,
                                                         collapse=collapse,
                                                         broadcast=broadcast)
-        if contig:
-            return (self.contig_k, self.prepare_args_contig(args, n, offsets)), n
 
         try:
             return self.try_specialized(args, n, nd, dims, strs, offsets), n
@@ -480,8 +482,7 @@ class ElemwiseKernel(object):
             k(*args, n=n)
 
     def call_contig(self, *args):
-        n, nd, dims, strs, offsets, contig = check_args(args, collapse=False,
-                                                        broadcast=False)
+        n, offsets, contig = check_contig(args)
         if not contig:
             raise ValueError("Can't call contig on non-contiguous data")
         if n != 0:
--- a/pygpu/tests/test_tools.py
+++ b/pygpu/tests/test_tools.py
@@ -1,9 +1,28 @@
 from pygpu.tools import (as_argument, Argument, ArrayArg, ScalarArg,
-                         check_args, Counter, lfu_cache)
+                         check_contig, check_args, Counter, lfu_cache)
 
 
 from .support import (guard_devsup, rand, check_flags, check_meta, check_all,
                       context, gen_gpuarray, dtypes_no_complex)
+
+
+def test_check_contig_1():
+    ac, ag = gen_gpuarray((50, 1, 20), 'float32', ctx=context)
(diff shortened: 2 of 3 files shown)


davidweichiang:
I forgot to add, if check_contig is kept, then the contiguity check in check_args could be removed. But I wasn't sure whether that would mess up the collapse=None case.

abergeron:
I lke check_contig so ok for that.

If you are interested, this function could use some tests that make sure it works correctly.  The cases you mentioned here could make a good starting point.  The only exception is that the (1, 1) case should reduce to (1,), not ().

davidweichiang:
OK, done. Thanks!

Do you have any thoughts on whether the contiguity check can be removed from check_args so as not to repeat work?

abergeron:
Yes it does "duplicate" work, but it's still useful to know that the inputs are contiguous in each case since it supports the `collapse=None` option.  If you can find a way to pass in the information then it may be useful to remove that check but I feel that it is too much effort for little gain.

davidweichiang:
Thanks! In `select_kernel`, `check_args` is called only in the discontiguous case, so `collapse=True` can be used instead. But if {`call_basic`, `call_dimspec`, `call_specialized`} are called directly, this is not true. However, I'm uncertain whether it makes sense to call `call_basic(..., collapse=None)`, etc. My hunch is that it does not, in which case I think that the `collapse=None` option could be removed. Is that right?

Or, is there a reason the following wouldn't work: in `check_args`,if `collapse=True`, and the arguments are detected to be contiguous, return `size, 1, size, (itemsize,)*len(args), offsets, True` as a special case. Then I think neither `check_contig` nor `collapse=None` would be needed.

abergeron:
Yes it does make sense to use collapse=None since you can reuse a basic kernel when contiguous (to avoid recompiling too often), but still get massive speedup by working on a vector from the point of view of the kernel.  The indexing code is quite costly.

Your second case could also work, yes, but I would make that `if collapse or collapse is None` since I don't want to drop support for that argument unexpectedly.  It doesn't cost much to keep it in.  Feel free to make a new PR with that.

davidweichiang:
Sorry, I don't quite understand yet -- as far as I can see, `basic_kernel` doesn't have a codepath to treat contiguous arrays specially. So I'm not seeing where `call_basic(..., collapse=None)` would get the speedup.

abergeron:
The codepath that does this is inside `check_args()`.  The speedup comes from the fact that after collapsing you only have a 1D array which is much faster to work with than 2D or 3D (or more).

davidweichiang:
I tried idea number 2 (exit `check_args` early if contiguous) and it was not as fast as the current version, so I like the current version better.

However, I'm still trying to understand the `collapse=None` option. I gather that the purpose of this option was to collapse a contiguous multiway array into a vector without actually going through the costly collapsing logic. But this optimization requires the caller to check the return value and select the appropriate kernel. Thus,

-  `ElementwiseKernel.__call__` now checks contiguity using `check_contig` and calls `check_args` only if discontiguous, in which case `collapse=None` and `collapse=True` are equivalent. So this case performs well, but doesn't need `collapse=None`.
- `ElementwiseKernel.call_basic` and friends call `check_args` unconditionally and ignore the `contig` flag returned by it. So if `k` is an `ElementwiseKernel` and `a` is a 10-dimensional contiguous array, and I call `k.call_basic(a, collapse=None)`, it will do slow 10-dimensional indexing -- but if I call `k.call_basic(a[...,0], collapse=None)`, it will collapse all 9 dimensions and do 1-dimensional indexiing! I can't think of any reason why that would be desired. It would seem faster to use `collapse=True` instead.

Thanks for your patience!

abergeron:
Yes it does ignore the contig flag, but not the returned dimensions and strides which is what the collapsing code modifies.

Also, `collapse=None` was introduce as a sort of "collapse-light" mode that would only run the expansive collapsing code if it can guarantee significant benefits from it (i.e the array is contiguous so everything will be collapsed).  It may not be useful in all cases and certainly not if you know exactly what your data is and how much you have to gain.

I don't see why there would be collapsing for `a[...,0]`, but not for `a` in your example though.

davidweichiang:
For `collapse=None`, collapsing occurs iff the array is _not_ contiguous (right?), so since `a` is contiguous but `a[...,0]` is not, the former will not be collapsed but the latter will.

abergeron:
No it only occurs iff the array is contiguous.

davidweichiang:
But...the docstring says: If `collapse` is None dimension collapsing will be performed if some arguments are non-contiguous.

And the code says:

    if not contig and collapse is None:
        ...
        collapse = True

abergeron:
Oh, you're right.  So yeah, this means `collapse=None` should map to `collapse=True` in all cases, so we can probably remove the contiguous check.  I somehow remember it being the opposite way ...

Sorry about that.

PR #63, opened 2015-07-29
Advanced indexing

abergeron:
This introduces GpuArray_take1() which performs an operation similar to numpy.take().

I also got rid of halloc since it brings very little benefit and a lot of complexity.


nouiz:
I didn't review in detail the take1 code. It will need tests. Tests from Theano Op that use it would be enough I think. WIth maybe minimal tests here just to make sure basic case work.

nouiz:
I have a segfault when I try to run the tests:


```
#0  0x00007ffff158f7f0 in cuda_init () from /u/bastienf/.local/lib64/libgpuarray.so
#1  0x00007ffff17aeeb6 in __pyx_pf_5pygpu_8gpuarray_10GpuContext_2__cinit__ (__pyx_v_kind=<value optimized out>, __pyx_v_devno=0x6191a0, __pyx_v_self=<value optimized out>) at pygpu/gpuarray.c:11343
#2  0x00007ffff17af222 in __pyx_pw_5pygpu_8gpuarray_10GpuContext_3__cinit__ (t=<value optimized out>, a=0x7fffe6dd1998, k=0x0) at pygpu/gpuarray.c:11294
#3  __pyx_tp_new_5pygpu_8gpuarray_GpuContext (t=<value optimized out>, a=0x7fffe6dd1998, k=0x0) at pygpu/gpuarray.c:25497
#4  0x00007ffff7cc4223 in type_call (type=0x7ffff19d9ae0, args=0x7fffe6dd1998, kwds=0x0) at Objects/typeobject.c:729

```

This is the gdb output if I run with gdb the equivalent of `DEVICE=cuda0 python -c "import pygpu;pygpu.test()"`

abergeron:
I've addressed the comments and fix a couple of small bugs.  All that remains is to add tests.

abergeron:
I've created some tests and fixed some bugs with OpenCL support that were discovered while running the tests.

src/gpuarray_array.c
```diff
@@ -305,6 +308,203 @@ int GpuArray_index(GpuArray *r, const GpuArray *a, const ssize_t *starts,
...
+
+  err = GpuKernel_call(&k, 2, ls, gs, 0, args);
+  free(args);
+  if (err == GA_NO_ERROR && check_error) {
+    err = v->ops->buffer_read(&kerr, errbuf, 0, sizeof(int));
+    if (err == GA_NO_ERROR) {
+      if (kerr != 0) err = GA_VALUE_ERROR;
+      kerr = 0;
+      /* We suppose this will succeed. */
+      v->ops->buffer_write(errbuf, 0, &kerr, sizeof(int));
```
nouiz: here write to errbuf only if needed. (kerr was not 0)

abergeron: done.

nouiz:
After having merged this PR, I get may of those warnings:

/usr/include/bits/nan.h:29:0: note: this is the location of the previous definition
./Tmp/bastienf/gpuarray.cuda.XXqMkijF:9:0: warning: "NAN" redefined

when I run:

DEVICE=cuda0 python -c "import pygpu;pygpu.test()"

abergeron:
The NAN problem is fixed by https://github.com/Theano/libgpuarray/pull/69

PR #65, opened 2015-08-05
round reduction size up to nearest power of two to avoid overloading cache

davidweichiang:
Hi, in `reduction.py`, `ReductionKernel._get_basic_kernel` is cached according to its arguments `maxls` and `nd`, where `maxls` is the smaller of the reduction size and `self.init_local_size` (= 1024 on my machine). If there are a lot of small reductions, there will be many different versions of the basic kernel in the cache, too many to fit in the cache. However, one of the first things that the basic kernel does is to round `maxls` up to the nearest power of two. This patch does the rounding up before caching so that there are at most lg(self.init_local_size) different versions of the basic kernel for each value of `nd`. (It also does the rounding again afterwards, to avoid breaking anything.) I think this fixes the problem with small reductions without costing anything for larger reductions.

### DIFF
--- a/pygpu/reduction.py
+++ b/pygpu/reduction.py
@@ -8,6 +8,12 @@ from . import gpuarray
 from .tools import ArrayArg, check_args, prod, lfu_cache
 from .elemwise import parse_c_args, massage_op
 
+def _ceil_log2(x):
+    # nearest power of 2 (going up)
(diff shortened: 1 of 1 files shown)


abergeron:
Good catch, this is certainly ok.

Issue #66, opened 2015-08-07
lfu_cache purges most-recently added items


davidweichiang:
Hi, even after patch #65, I find that `lfu_cache` is still causing too many kernel recompiles. The reason seems to be that when an item is ejected from the cache, its count is also reset. When an item is added, its count is therefore 1, and is liable to get ejected again right away. However, I don't know the most appropriate fix -- perhaps an LRU policy instead?

nouiz:
Or a combination of minimum life time + lfu?

I think LRU would be better the just lfu for the reason you gave and it
would be probably simpler then combination.

davidweichiang:
Please see #67, thanks!

I'm also wondering if you consider it an issue that ndgpuarray's `__add__` and friends recompile their kernels every single time (because they create a new `ElemwiseKernel`/`ReductionKernel` object and `self` is part of the cache key).

nouiz:
That is not optimal for sure. But we don't have time shortly for this. Do
you want to try it? If not, make another issue.

abergeron:
The plan was/is to add a cache at the C level to try to avoid kernel recompiles.  But making a cache in C is much harder.

nouiz:
Just to be sure, that isn't in the short term plan, so if you need a quick
fix, you could make something at the python level and we would accept that.

PR #67, opened 2015-08-08
add lru_cache as alternative to lfu_cache

davidweichiang:
Follow-up to issue #66.

fix gh-66


nouiz:
Thanks. I think it is good, but I'll let @abergeron time to review it next week.

PR #71, opened 2015-08-25
Use nvrtc, instead of nvcc, to compile CUDA to PTX.

seanprime7:
These commits allow libgpuarray to use nvrtc (http://docs.nvidia.com/cuda/nvrtc/index.html) to compile CUDA source code to PTX, instead of nvcc, if nvrtc is found on the system.

It should reduce compilation overhead associated with nvcc subprocess creation, nvcc intermediate file disk I/O, etc.

If nvrtc is not found on the system, nvcc is used as before.

### DIFF
--- a/pygpu/elemwise.py
+++ b/pygpu/elemwise.py
@@ -42,8 +42,10 @@ KERNEL void ${name}(const unsigned int n
 % endfor
 
   for (i = idx; i < n; i += numThreads) {
+% if nd > 0:
     int ii = i;
     int pos;
+% endif
 % for arg in arguments:
     % if arg.isarray():
         GLOBAL_MEM char *${arg.name}_p = (GLOBAL_MEM char *)${arg.name}_data;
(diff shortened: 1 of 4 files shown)


abergeron:
I have another way of doing that in that was in the works and is #72

I don't think that executing ptxas is a good idea since we still waste time waiting for a forked process.  You can give PTX code directly to cuLoadModule so that is the route I've taken.

I think your idea of adding a prefix to detect_arch is better than my approach, so I switched to that.

Also I would like the fixes you made in the last two commits, perhaps in another PR.

seanprime7:
@abergeron, I guess you will merge PR #72 instead of this one.  If so, I will close this PR.  Otherwise, please let me know then I will update this PR.

abergeron:
Probably, but I still have to get someone else to review it.  So let's wait a bit.

seanprime7:
Okay.

abergeron:
If I'm not mistaken all the changes in this PR where merge in other PRs.  So I'll close this one.

PR #74, opened 2015-08-26
Match the data types used for dimensions, size, etc. with the ones in…

seanprime7:
As per @abergeron's request, creating a separate PR with this patch from PR #71.

### DIFF
--- a/pygpu/elemwise.py
+++ b/pygpu/elemwise.py
@@ -136,7 +136,7 @@ KERNEL void ${name}(
 contiguous_kernel = Template("""
 ${preamble}
 
-KERNEL void ${name}(const ga_size n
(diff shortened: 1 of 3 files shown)


pygpu/elemwise.py
```diff
@@ -136,11 +136,11 @@
 contiguous_kernel = Template("""
 ${preamble}
 
-KERNEL void ${name}(const unsigned int n
```
abergeron: This should also stay unsigned int (or ga_uint).

abergeron:
Ok for this except for my one comment.

pygpu/elemwise.py
seanprime7: Fixed it with f88b39f

PR #77, opened 2015-09-01
Hgemm

abergeron:
Add support for float16 data whenever the blas library supports it.


abergeron:
Done.

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -171,28 +240,53 @@ static int sgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
...
+    offA = offB;
+    offB = lt;
+  }
+
+  cuda_enter(ctx);
+
+  for (i = 0; i < batchCount; i++) {
+    cuda_wait(A[i], CUDA_WAIT_READ);
+    cuda_wait(B[i], CUDA_WAIT_READ);
+    cuda_wait(C[i], CUDA_WAIT_READ|CUDA_WAIT_WRITE);
```
nouiz: I think the main use case is that the list of A, B and C are different "view" of the same gpu tensor. Is it useful to have a param to tell this and have the wait once before the loop?

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -258,24 +381,239 @@ static int dgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
...
-      ARRAY_FINI(C[i]);
+static int sgemv(cb_order order, cb_transpose transA, size_t M, size_t N,
+                 float alpha, gpudata *A, size_t offA, size_t lda,
+                 gpudata *X, size_t offX, int incX,
+                 float beta, gpudata *Y, size_t offY, int incY) {
+  cuda_context *ctx = A->ctx;
+  cublasStatus_t err;
+  size_t t;
+
+  ASSERT_BUF(A);
```
nouiz: You don't do that check for *gemm and *batchedgemm. Be consistent.

nouiz: idem for ger.

src/gpuarray_blas_cuda_cublas.c
abergeron: It may be useful, but would be too complex to communicate in a non-annoying way through the interface.  Since I don't see any calls anywhere to this function I also fail to see any use case for now.

If we have a use case where A, B and C are essentially a 3D tensors that we perform gemm elementwise on, a different interface would be much faster.  Otherwise I can look at the problem and try to come up with a reasonable interface if possible.

src/gpuarray_blas_cuda_cublas.c
abergeron: I've added it everywhere.

src/gpuarray_blas_cuda_cublas.c
nouiz: In Theano, we support only that 3d case. So adding support for it make sence. I like your idea of another method for that case. This can be done later.

nouiz: https://github.com/Theano/libgpuarray/issues/78

PR #79, opened 2015-09-07
OpenCL Support for Theano's libgpuarray

Nirvik-D:
Starting work on the Theano OpenCL support. Initially studying the Blas files and figuring out the C code, before starting to work on the Ops.

### DIFF
--- a/src/gpuarray_buffer_opencl.c
+++ b/src/gpuarray_buffer_opencl.c
@@ -16,6 +16,7 @@
 #define strdup _strdup
 #endif
 
+#define _unused(x) ((void)x)
 #define SSIZE_MIN (-(SSIZE_MAX-1))
 
 static cl_int err;
@@ -639,6 +640,7 @@ static int cl_memset(gpudata *dst, size_t offset, int data) {
   }
   /* If this assert fires, increase the size of local_kern above. */
   assert(r <= sizeof(local_kern));
+  _unused(r);
 
   sz = strlen(local_kern);
   rlk[0] = local_kern;
--- a/src/gpuarray_blas_opencl_clblas.c
+++ b/src/gpuarray_blas_opencl_clblas.c
@@ -78,18 +78,18 @@ static int sgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
   cl_event evl[3];
   cl_event ev;
   size_t i;
-  cl_uint num_ev;
+  cl_uint num_ev = 0;
   clblasStatus err;
 
   for (i = 0; i < batchCount; i++) {
     ARRAY_INIT(A[i]);
     ARRAY_INIT(B[i]);
     ARRAY_INIT(C[i]);
-    err = clblasSgemm(convO(order) convT(transA), convT(transB), M, N, K,
+    err = clblasSgemm(convO(order), convT(transA), convT(transB), M, N, K,
                       alpha, A[i]->buf, offA[i], lda, B[i]->buf, offB[i], ldb,
-                      beta, C[i]->buf, offB[i], ldc, 1, &ctx-q,
+                      beta, C[i]->buf, offB[i], ldc, 1, &ctx->q,
                       num_ev, num_ev == 0 ? NULL : evl, &ev);
-    if (err = clblasSuccess)
+    if (err != clblasSuccess)
       return GA_BLAS_ERROR;
     ARRAY_FINI(A[i]);
     ARRAY_FINI(B[i]);
@@ -110,18 +110,18 @@ static int dgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
   cl_event evl[3];
   cl_event ev;
   size_t i;
-  cl_uint num_ev;
+  cl_uint num_ev = 0;
   clblasStatus err;
 
   for (i = 0; i < batchCount; i++) {
     ARRAY_INIT(A[i]);
     ARRAY_INIT(B[i]);
     ARRAY_INIT(C[i]);
-    err = clblasDgemm(convO(order) convT(transA), convT(transB), M, N, K,
+    err = clblasDgemm(convO(order), convT(transA), convT(transB), M, N, K,
                       alpha, A[i]->buf, offA[i], lda, B[i]->buf, offB[i], ldb,
-                      beta, C[i]->buf, offB[i], ldc, 1, &ctx-q,
+                      beta, C[i]->buf, offB[i], ldc, 1, &ctx->q,
                       num_ev, num_ev == 0 ? NULL : evl, &ev);
-    if (err = clblasSuccess)
+    if (err != clblasSuccess)
       return GA_BLAS_ERROR;
     ARRAY_FINI(A[i]);
     ARRAY_FINI(B[i]);
@@ -271,7 +271,36 @@ static int sger(cb_order order, size_t M, size_t N, float alpha,
   err = clblasSger(convO(order), M, N, alpha, X->buf, offX, incX,
                    Y->buf, offY, incY, A->buf, offA, lda, 1, &ctx->q,
                    num_ev, num_ev == 0 ? NULL : evl, &ev);
-  if (err = clblasSuccess)
+  if (err != clblasSuccess)
+    return GA_BLAS_ERROR;
+
+  ARRAY_FINI(X);
+  ARRAY_FINI(Y);
+  ARRAY_FINI(A);
+
+  clReleaseEvent(ev);
+
+  return GA_NO_ERROR;
+}
+
+static int dger(cb_order order, size_t M, size_t N, double alpha,
+                gpudata *X, size_t offX, int incX,
+                gpudata *Y, size_t offY, int incY,
+                gpudata *A, size_t offA, size_t lda) {
+  cl_ctx *ctx = X->ctx;
+  cl_event evl[3];
+  cl_event ev;
+  cl_uint num_ev = 0;
+  clblasStatus err;
+
+  ARRAY_INIT(X);
+  ARRAY_INIT(Y);
+  ARRAY_INIT(A);
+
+  err = clblasDger(convO(order), M, N, alpha, X->buf, offX, incX,
+                   Y->buf, offY, incY, A->buf, offA, lda, 1, &ctx->q,
+                   num_ev, num_ev == 0 ? NULL : evl, &ev);
+  if (err != clblasSuccess)
     return GA_BLAS_ERROR;
 
   ARRAY_FINI(X);


Nirvik-D:
@nouiz  @abergeron I see that the gpuarray.py module is gone. Are you trying to generate gpuarray in C?

nouiz:
Amend your commit to not include the file:  lib/libgpuarray-static.a

There was a name colision, the file gpuarray.py have been renamed _array.py. It not much used for now I think.

I let @abergeron make a good review of this PR, but a quick look seem ok.

abergeron:
There never was a gpuarray.py file.  There are gpuarray.pyx and .pxd files which are Cython code for this module.

_array.py was renamed from array.py

abergeron:
Also the changes look good for the opencl code, but please amend your commit as @nouiz said.

Nirvik-D:
@nouiz @abergeron I might have possibly built and installed the pygpu in a corrupted manner. I tried building blas.pyx and gpuarray.pyx exclusively using Cython which resulted in some .so files. After that pygpu installed successfully but I was receiving all sorts of `Memory allocation failure` from a broken gpuarray. But now it's fixed and working correctly.

Also I have removed the `libgpuarray-static.a` in the latest commit.

Nirvik-D:
@nouiz Is the plan for libgpuarray to have all the sandbox files ported over to libgpuarray? Since, I haven't seen the use of libgpuarray yet. What are the possible uses of libgpuarray in the future? Do we continue using the sandbox gpuarray for now?

nouiz:
You can use it in the sandbox for now without problem. Currently, we have
all the ops from the Deep Learning Tutorial. So it is useful.

At some point, we will move it outside the sandbox but in Theano. I think a
good time is when we will recommand that everybody use it. For now, we
can't as I think too many ops are missing.

I don't think we will port everything inside libgpuarray. But they can stay
in Theano with OpenCL version. All what is in numpy are good canditate to
move to libgpuarray, but not what is in nnet.py for example.

Nirvik-D:
@nouiz Ok Understood.

abergeron:
Could you merge your commits together so that the .a isn't is the history.  This bloats checkouts.

Nirvik-D:
@abergeron Done.

abergeron:
I think there was a miscommunication.  You currently have 4 commits in your branch.  There should be only 1, which should correspond to 9c9365e

Nirvik-D:
@abergeron Messed up the repository. Should work now. There should be 1 commit now.

src/gpuarray_buffer_opencl.c
```diff
@@ -638,6 +638,7 @@ static int cl_memset(gpudata *dst, size_t offset, int data) {
                  offset, n, val);
   }
   /* If this assert fires, increase the size of local_kern above. */
+  if (r != GA_NO_ERROR) return r;
```
abergeron: This is wrong.  r is not a gpuarray error code.

Nirvik-D:
@abergeron Done. Have a look and let me know if it requires any more changes.

abergeron:
All good, thank you.

Nirvik-D:
@abergeron @nouiz You are welcome. I am now trying to now work on the MLP.

Nirvik-D:
@abergeron @nouiz 
Facing new errors while carrying out tests today. Have a look.

```File "/usr/local/lib/python2.7/site-packages/pygpu/__init__.py", line 4, in get_include
    assert os.path.exists(os.path.join(p, 'gpuarray_api.h'))
AssertionError: ('The following error happened while compiling the node', HostFromGpu(gpuarray)(<GpuArrayType(float32, (False,))>), '\n')```

**Comment by abergeron** [comment]
The pygpu install is not done right since it's missing includes.  Have you run cython on the relevant .pyx files?  This should be done automatically when you run the setup.py.

**Comment by Nirvik-D** [comment]
I have been running `sudo python setup.py install`, but the error still exists. I haven't cythonized anything externally yet. During the installation of pygpu, it shows that `gpuarray_api.h , blas_api.h , numpy_compat.h` and `gpuarray.h` has been installed in a separate `pygpu` folder under `/usr/local/` , while the rest of the pygpu is installed at `/usr/local/lib/pytho2.7/site-packages/`

**Comment by Nirvik-D** [comment]
Removed and installed repeatedly, but to no avail. The same error keeps generating as pygpu fails to find the required headers.

**Comment by Nirvik-D** [comment]
@abergeron 
Been reinstalling for hours. But pygpu still can't find the headers. What could be the possible reasons? It was working fine till yesterday.

**Comment by abergeron** [comment]
It seems that there is a weirdness since those headers should go in the package directory not under /usr/local.

It may be the version of distutils/distribute that you have which has a bug.

**Comment by Nirvik-D** [comment]
@abergeron 
I found the header files going inside the `/usr/local/` directory extremely weird as well . I have never seen it happen before in my earlier installations of `pygpu`. When I tried re-installing from the `pygpu ` available in your repository, it kept showing the following:

```
byte-compiling /usr/local/lib/python2.7/site-packages/pygpu/elemwise.py to elemwise.pyc
byte-compiling /usr/local/lib/python2.7/site-packages/pygpu/__init__.py to __init__.pyc
byte-compiling /usr/local/lib/python2.7/site-packages/pygpu/_array.py to _array.pyc
byte-compiling /usr/local/lib/python2.7/site-packages/pygpu/tools.py to tools.pyc
byte-compiling /usr/local/lib/python2.7/site-packages/pygpu/reduction.py to reduction.pyc
running install_data
creating /usr/local/pygpu
copying pygpu/gpuarray.h -> /usr/local/pygpu
copying pygpu/gpuarray_api.h -> /usr/local/pygpu
copying pygpu/blas_api.h -> /usr/local/pygpu
copying pygpu/numpy_compat.h -> /usr/local/pygpu
running install_egg_info
Writing /usr/local/lib/python2.7/site-packages/pygpu-0.2.1-py2.7.egg-info
raptor@raptor-535U4C:~/Deep_Learning/libgpuarray$ 
```

**Comment by abergeron** [comment]
Since it work correctly for me (and I just tried again), I strongly suspect a bug in whatever ends up 
selected as the build tool in setup.py.

Can you post the output of `pip list | grep setuptools` or if that doesn't produce any output the output of `python -V`.  It is also possible that you have distribute install, in which case I recommend that you remove it since it is unsupported and often wrong.

For referrence I have version 18.1 of setuptools and it works fine.

**Comment by Nirvik-D** [comment]
@abergeron 

```
raptor@raptor-535U4C:~/Project_Index_CL$ pip list | grep setuptools
setuptools (18.3.1)
raptor@raptor-535U4C:~/Project_Index_CL$ python -V
Python 2.7.9
raptor@raptor-535U4C:~/Project_Index_CL$ 
```
Can you tell me where the include files are supposed to go? LIke should they be installed in /usr/local/include/pygpu? My gpuarray headers are all installed in /usr/local/include/gpuarray and they work just fine. I did a fresh installation and they are still responding the same way.

abergeron:
The gpuarray headers are supposed to go somewhere like /usr/local/include (which is seems they are).

The headers in the python package are supposed to go into the python package dir right next to the .py files.

Nirvik-D:
@abergeron 

Ok. I will just remove setuptools and reinstall and then try once more.

abergeron:
First I would check if you have distribute installed.  That could be the cause of the problem.

Nirvik-D:
@abergeron 

I just reinstalled distribute and setuptools. Checking once more.

abergeron:
You should uninstall distribute, it is bad.  setuptools is ok though.

Nirvik-D:
@abergeron 

Ok. I will remove it. But I think setuptools isn't working for me, since the warnings are coming from distutils.

abergeron:
setuptools uses distutils behind the scene and does a bit of monkey patching so you will see a lot of references to it.

Nirvik-D:
@abergeron 
It's still the same. The pygpu folder forms under /usr/local and the header files get copied over there.

abergeron:
Just to get it working you can probably manually copy the headers where they are supposed to go.

But otherwise, without a way to reproduce this myself, I can't really debug why this is happening.

Nirvik-D:
@abergeron 

Did that. Its bad practice, but still it works :+1: . So I guess it's all good for now. I will try to figure out what's going on later. Also, the corr_gemm porting is partially done. I will commit it for verification as soon as possible.

Nirvik-D:
@abergeron 

Apparently, data_setup has a bug, and it exists across all Ubuntu systems. The bug allows distutils to install extension packages break norms and mess up the installation path. This is specifically why, the pure python packages are getting installed into `site-packages` and the non-python packages are getting installed into `sys.exc_prefix`, which in my case is `usr/local/`

abergeron:
I do have Ubuntu, but I use python from anaconda so that may be why I don't get this problem.

I'll try to find a workaround for this, if possible.

Nirvik-D:
@abergeron 

That could be a possible reason. It's a reported bug at Ubuntu. Also, regarding the corr_gemm kernel, can you give a look at https://github.com/Theano/Theano/pull/3372 ? I am some problems with the GEMM implementation. @nouiz said about 3 new Ops. But I can't find them.

nouiz:
They are in sandbox/cuda/blas.py. It is the class GpuCorrMM
GpuCorrMM_gradWeights and GpuCorrMM_gradInputs

Nirvik-D:
@nouiz 

Ok, I will check it out.

Nirvik-D:
@nouiz 

The three Ops are based on BaseGpucorrMM class and that class makes use of C helper code containing CUDAndarray. Wouldn't that make the Ops exclusive to CUDA?

abergeron:
Yes, they currently are.  The point is to port the code so that it will work with libgpuarray and OpenCL

nouiz:
I think @abergeron mean with CUDA and OpenCL

Nirvik-D:
@abergeron @nouiz 

When you said that libgpuarray is being used as a wrapper for OpenCL, which files can I go through to invoke the OpenCL host code? Like getting the platform ID, creating the context, building the kernel and invoking etc. Since, I am only familiar with the C version of OpenCL, it's difficult to find it here in libgpuarray.

Nirvik-D:
@abergeron @nouiz 

And I think it would be better if I bring the two kernels im2col and col2im to libgpuarray. It's practically of no use to keep them their, I think. I plan on writing the 2 kernels in libgpuarray and invoke the kernels via the libgpuarray wrapper host code, since it can't be called directly in the kernel like CUDA. I have gone through it for hours and the only way possible to invoke the kernel function is through the host code in which I can try to inculcate the PyGpuarray code as well. What do you think?

abergeron:
Theano is also "host code" here. "host code" refers to things that run on the CPU.  I think it's a better idea to start in theano for this since I'm not sure I want to provide convolution in libgpuarray.

I have created a class called GpuKernelBase to help use custom kernels with theano.  You can see multiple examples of its use but the simplest is the GpuEye code.  This one has a single kernel, but you can have more than one if you need more.

Nirvik-D:
@abergeron 

Ok, then I will leave the im2col and col2im kernels under sandbox.gpuarray. Meanwhile, I will start working on the the GpuKernelBase.

PR #87, opened 2015-09-22
Add cache

abergeron:
This adds a cache to kernel compiles which will hopefully speed-up certain use cases.  This cache is per-process (so multiple contexts can reuse the same compiled kernel if their GPUs are compatible).


CMakeLists.txt
```diff
@@ -12,6 +12,11 @@ execute_process(COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_SOURCE_DIR}/l
 set(LIBRARY_OUTPUT_PATH ${CMAKE_SOURCE_DIR}/lib)
 
 set(CMAKE_OSX_ARCHITECTURES x86_64)
+set(CMAKE_C_VISIBILITY_PRESET hidden)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
+if(POLICY CMP0063)
```
nouiz: Add a comment to tell that this is.

abergeron: This is an unfortunate consequence of using cmake.  The only reason this is here is to make it not spam warnings when building with newer versions.

I could add a generic comment explaining that, but I won't explain what each policy does individually, that would be too much work and is already easily searchable on the internet for the curious.

nouiz: that is good.

abergeron:
Ok, done.

abergeron:
What is happening with this?

nouiz:
Just no time to review.

abergeron:
I would put a bump on this since it's killing performance to not have it for the multi-gpu (I'm talking about 90s vs 8s for each epoch) and that is with nvrtc.

I just encountered that while testing my other PR about caching allocations.

nouiz:
What is this 90s vs 8s?

abergeron:
Execution time for one epoch (in seconds)

abergeron:
The last commit here fixes a bunch of problems with compilation that somehow were still on the uploaded branch.  I blame a bad rebase (I think).

Apart from that it passes all tests.

abergeron:
Last commit should fix the undefined symbol 'assert' on linux.

src/gpuarray_buffer_cuda.c
```diff
@@ -910,6 +973,36 @@ static gpukernel *cuda_newkernel(void *c, unsigned int count,
           cuda_exit(ctx);
           return NULL;
         }
+        if (compile_cache == NULL)
```
nouiz: I would move this to before line 933, otherwise, sometimes the cache won't be defined.

abergeron: It's ok.  The check above is only to see if something is in the cache.  So if there is no cache, nothing's in there.

Issue #92, opened 2015-10-23
"Could NOT find clBLAS" error in CMake even when clBLAS is correctly installed


thebucc:
The FindclBLAS.cmake script is broken, it can't find clBLAS even if present

thebucc:
By the way, I'm seeing the problem in Windows 7, unverified in other OSes.
I have posted a fixed version here: https://groups.google.com/forum/#!topic/clmath/nsqv-WA5SH4, how do I commit the change here, what kind of rules do you guys follow?

abergeron:
If you have a fixed version of FindclBLAS.cmake, you can submit a PR and I'll check it out.

I'll just have to make sure it still works on linux and mac.

thebucc:
All done. Please let me know if I haven't followed the correct procedure.

abergeron:
Yes, the PR is good, I just have a comment to make it work on other platforms.

PR #93, opened 2015-10-24
Changes for issues #92, #94, #95

thebucc:
This fix works on Windows x64, with the default directory structure in clBLAS. It probably needs additional changes for win x32 and other OSes, particularly around line 26:

HINTS ${CLBLAS_ROOT_DIR}/lib64/import)

More in issue #92.


CMakeModules/FindclBLAS.cmake
```diff
@@ -1,16 +1,35 @@
...
+    DOC "clBLAS root directory.")
+    
+FIND_PATH(_CLBLAS_INCLUDE_DIRS
+    NAMES clBLAS.h
+    HINTS ${CLBLAS_ROOT_DIR}/include
+    DOC "clBLAS Include directory")
+
+FIND_LIBRARY(_CLBLAS_LIBRARY
+    NAMES CLBLAS
+    HINTS ${CLBLAS_ROOT_DIR}/lib64/import)
```
abergeron: I would rework that part to use 

```
HINTS ${CLBLAS_ROOT_DIR}/lib64 ${CLBLAS_ROOT_DIR}/lib
PATH_SUFFIXES import
```

The above should work on linux and max, so if it works for you on win 64, then I would merge it.

thebucc: Hi @abergeron, 
I have added your suggested change, plus another hint with "lib32".
Also, I changed line 16 to have $ENV{CLBLAS_ROOT} instead of ${CLBLAS_ROOT}, because otherwise it still wasn't finding the libraries for me: clearly yesterday I'd set the paths explicitly. I don't know CMake at all, so I'd rather run this change with you before submitting a pull request. Please have a look at the modified FindclBLAS.cmake in my branch https://github.com/thebucc/libgpuarray and let me know if it's ok.

abergeron: The changes to FindclBLAS.cmake look ok.

abergeron:
You've added two commits.  One to fix some problems with python 3 (which I'm ok with) and one to add strndup.

Instead of the new strndup implementation, I would use memdup to make the string copy at the one spot where it is used.

thebucc:
You are perfectly right about strndup , memdup does the job just as well and is a much smaller change. I decided to write an implementation for Windows because this could happen again, given that strndup exists in linux and mac (and I guess Cygwin too). As long as people are aware of this and avoid using strndup, we don't need it.

abergeron:
I'm not against adding it, if it's done in the same style as the other functions.

Currently it is always added if _MSC_VER is defined and this will break if Microsoft decides to add it in a future version.  Look at how mkstemp is handled:

https://github.com/Theano/libgpuarray/blob/master/src/CMakeLists.txt#L47
https://github.com/Theano/libgpuarray/blob/master/src/CMakeLists.txt#L57
https://github.com/Theano/libgpuarray/blob/master/src/private_config.h.in#L8
https://github.com/Theano/libgpuarray/blob/master/src/private_config.h.in#L40

This is what you should do for strndup.

Or you can also just use memdup if you prefer.

thebucc:
I banged my head long enough against CMake last night, memdup will be fine, thank you very much. If someone needs strndup in the future, it's still in my history, ready to be reimplemented properly.

thebucc:
Git is doing its very best to piss me off. How do I associate single changes to a pull request? Essentially, I have 3 separate issues here (#92, #94 and #95), and I want to create a pull request for each one, pointing at only the relevant change from my branch. I definitely don't want all thrown in together. How do I do that?

lamblin:
@thebucc You would have to have 3 different local branches to do that, each with only the relevant changes, then open one pull request from each of these branches. You can use `git cherry-pick` to select which commits from your master branch to include in those minimal branches.

abergeron:
I'm fine with having all the changes in a single PR.  It may delay the merge by a bit since we'll have to make sure all the changes are ok before merging though.

thebucc:
Too much effort creating three branches, I started but I give up. I'll leave everything in a single pull request. Not ideal, but git is just not up to the task.

PR #96, opened 2015-10-27
Fix for issue #92 only: improved detection of clBLAS

thebucc:
Only changes to FindclBLAS.cmake


thebucc:
Too much effort creating three branches, I'll leave everything in a single pull request. Not ideal, but git is just not up to the task.

PR #97, opened 2015-11-04
Use object instead of type.

abergeron:


abergeron:
This should address the problem that cropped up with recent versions of cython that use PyTypeObject instead of PyObject for `type`.

nouiz:
This still compile with the old gcc, so I merge.

PR #98, opened 2015-11-09
Cache and reuse allocations for cuda.

abergeron:


src/gpuarray_buffer_cuda.c
```diff
@@ -304,61 +295,139 @@ static void cuda_deinit(void *c) {
...
+  (*res)->next = next;
+  if (*prev)
+    (*prev)->next = *res;
+  else
+    ctx->freeblocks = *res;
+
+  return GA_NO_ERROR;
+}
+
+static int extract(gpudata *curr, gpudata *prev, size_t size) {
```
nouiz: add a comment that tell what this method do. It took me some times to understand that.

abergeron: ok

abergeron:
I've pushed a fix for the freelist issue.  Now it works with or without allow_gc.

I don't think there are any issues left with the freelist code since I did some pretty intense tracing over everything and all the sizes match all the time.  Barring style issues, I think it's ready for a merge.

PR #99, opened 2015-11-11
Make cuda_read() and cuda_write() asynchronous.

abergeron:
I am not 100% sure that this change is completely safe, so don't merge until we are sure it is.

Otherwise, this speeds up some workloads significantly.

### DIFF
--- a/src/gpuarray_buffer_cuda.c
+++ b/src/gpuarray_buffer_cuda.c
@@ -599,17 +599,15 @@ static int cuda_read(void *dst, gpudata *src, size_t srcoff, size_t sz) {
 
     cuda_enter(ctx);
 
-    ctx->err = cuEventSynchronize(src->ev);
+    cuda_wait(src, CUDA_WAIT_READ);
+
+    ctx->err = cuMemcpyDtoHAsync(dst, src->ptr + srcoff, sz, ctx->s);
     if (ctx->err != CUDA_SUCCESS) {
       cuda_exit(ctx);
       return GA_IMPL_ERROR;
     }
 
-    ctx->err = cuMemcpyDtoH(dst, src->ptr + srcoff, sz);
-    if (ctx->err != CUDA_SUCCESS) {
-      cuda_exit(ctx);
-      return GA_IMPL_ERROR;
-    }
+    cuda_record(src, CUDA_WAIT_READ);
     cuda_exit(ctx);
     return GA_NO_ERROR;
 }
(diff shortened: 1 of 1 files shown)


src/gpuarray_buffer_cuda.c
```diff
@@ -484,17 +482,15 @@ static int cuda_write(gpudata *dst, size_t dstoff, const void *src,
...
 
-    ctx->err = cuEventSynchronize(dst->ev);
-    if (ctx->err != CUDA_SUCCESS) {
-      cuda_exit(ctx);
-      return GA_IMPL_ERROR;
-    }
+    cuda_wait(dst, CUDA_WAIT_WRITE);
 
-    ctx->err = cuMemcpyHtoD(dst->ptr + dstoff, src, sz);
+    ctx->err = cuMemcpyHtoDAsync(dst->ptr + dstoff, src, sz, ctx->s);
```
nouiz: Je ne pense pas qu'on peut faire cela async, car le code CPU ne fait pas de wait. Le prochain noeud va commencé à exécuter automatiquement. Peut-tu timmer sans cette modif?

abergeron: (This is public so english please)

Also this part give 100% of the observed speedup in my alexnet example.

nouiz: Can you explain why the HtoD make this difference? It allow computation between Theano function call?

abergeron: Basically it does the copy to a staging buffer and returns.  Then the actual copy to the GPU will be done later so we can go on scheduling some other things and not wait for the copy to finish.

However the documentation states that the copy to the staging buffer happens in a worker thread, so maybe there is some time after it returns where it isn't actually safe to touch the CPU memory which is why I don't want to merge this until we are sure.

src/gpuarray_buffer_cuda.c
```diff
@@ -456,17 +456,15 @@ static int cuda_read(void *dst, gpudata *src, size_t srcoff, size_t sz) {
...
 
-    ctx->err = cuEventSynchronize(src->ev);
-    if (ctx->err != CUDA_SUCCESS) {
-      cuda_exit(ctx);
-      return GA_IMPL_ERROR;
-    }
+    cuda_wait(src, CUDA_WAIT_READ);
 
-    ctx->err = cuMemcpyDtoH(dst, src->ptr + srcoff, sz);
+    ctx->err = cuMemcpyDtoHAsync(dst, src->ptr + srcoff, sz, ctx->s);
```
nouiz: Why this is safe from Theano point of view? Do the host have a complete copy after that? If not, then other ops on the CPU could start to compute on it before the transfer is complete.

src/gpuarray_buffer_cuda.c
nouiz: thanks for the detail. You are right, we need to make sure how it work. Otherwise we will need to copy the data to pinned memory ourself.

src/gpuarray_buffer_cuda.c
abergeron: Again I'm not sure and that is why I want to confirm before merging.  You were the one who asked me for the code.

PR #101, opened 2015-11-13
Enable async read/write ops on a separate stream.

abergeron:
Tests have assured me that we don't need to wait on the host side
after a transfer in either direction to play with host memory.  The
only required synchronizations are with the device-side operations
which we perform correctly here.

This supersedes #99

### DIFF
--- a/src/gpuarray_buffer_cuda.c
+++ b/src/gpuarray_buffer_cuda.c
@@ -638,15 +638,23 @@ static int cuda_read(void *dst, gpudata *src, size_t srcoff, size_t sz) {
 
     cuda_enter(ctx);
 
-    cuda_waits(src, CUDA_WAIT_READ, ctx->mem_s);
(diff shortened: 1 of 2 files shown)


abergeron:
Since NVIDIA can't guarantee that cuMemcpy{D,H}to{D,H}Async will not need host memory after it returns in *all* cases, I've made my own wrapper which guarantees that.

It might be suboptimal for the case where we play with host-mapped memory, but as far as I know this only happens with errbuf.  If we ever consider using host-mapped memory more, then maybe some adjustments will have to be done.

abergeron:
I changed the strategy to use memcpy directly when involving a host-mapped pointer.  I also insert the appropriate synchronization to make the the read memory is valid.

abergeron:
Added a fix for some potential problems with concurrent read and writes on more than one stream.  This also makes the code easier to follow.

Issue #100, opened 2015-11-11
Test errors with numpy 1.10


abergeron:
```
======================================================================
ERROR: pygpu.tests.test_elemwise.test_ielemwise2_ops_array(<built-in function iadd>, 'int8', 'float32', (50,))
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Tmp/lisa/os_v5/anaconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/u/bastienf/.local/lib64/python2.7/site-packages/pygpu-0.2.1-py2.7-linux-x86_64.egg/pygpu/tests/support.py", line 39, in f
    func(*args, **kwargs)
  File "/u/bastienf/.local/lib64/python2.7/site-packages/pygpu-0.2.1-py2.7-linux-x86_64.egg/pygpu/tests/test_elemwise.py", line 81, in ielemwise2_ops_array
    out_c = op(ac, bc)
TypeError: Cannot cast ufunc add output from dtype('float32') to dtype('int8') with casting rule 'same_kind'
```

nouiz:
I'm working in fixing problem with numpy 1.10, it make it harder to run libgpuarray tests to check PR.

PR #110, opened 2015-11-20
Update for NumPy 1.10

nouiz:
fix gh-100


pygpu/tests/test_gpu_ndarray.py
```diff
@@ -26,7 +26,7 @@ def permutations(elements):
     if len(elements) <= 1:
         yield elements
     else:
-        for perm in permutations(elements[1:]):
```
abergeron: elements has to be a list (or tuple) already.

nouiz:
I updated the PR

PR #111, opened 2015-11-20
Fix PTX version.

### LINKED ISSUE #107: Verify that the PTX version in extcpy is actually working on all supported cuda version

abergeron:
This works for 6.5, 7.0 and 7.5.  I still have to test for 6.0, but I hope it will work as-is.

Fixes #107


nouiz:
Ok to merge when the doc is updated.

abergeron:
Updated the doc (with also an advice about running ldconfig on linux).

PR #112, opened 2015-11-20
Fix pickling returning broken objects

abergeron:
We just disable pickling because we can't preserve and restore the GPU state.


nouiz:
Should you add test that this really raise an error?

abergeron:
Added a test for GpuArray, GpuKernel and GpuContext.  flags does support pickle, but I don't think it's very useful and I don't want to make any promises with tests for it.

PR #117, opened 2015-11-27
[MRG] Release the gil

nouiz:
the other nogil attribute, just tell we can call those function without having the gil, it wasn't freeing them.

This isn't finished, I just wanted to have this public to show this. We need to do this at other point.


nouiz:
I added all missing nogil that I think of.

abergeron:
You're right.

PR #120, opened 2016-01-13
Create a primary context for better compat with the runtime API.

abergeron:
This functionality doesn't exist in 6.5 so we keep the old behavior for that version in order to not drop support right now.


src/gpuarray_buffer_cuda.c
```diff
@@ -291,21 +321,6 @@ static void *cuda_init(int ord, int flags, int *ret) {
     cuda_context *res;
     static int init_done = 0;
 
-    if (ord == -2) {
```
nouiz: add back with ifdef.

abergeron:
Finally the -2 thing was never documented anywhere, so we can just remove it silently I think.  It's a dirty hack anyway.

PR #121, opened 2016-01-24
Fix for clBLAS library name capitalization

gmcgoldr:
I tried installing with a fresh clone of clBLAS. The generated library name was libclBLAS.so, which is not found with the given cmake routine.


gmcgoldr:
I tried installing with a fresh clone of clBLAS. The generated library name was libclBLAS.so, which is not found with the given cmake routine.

abergeron:
Thanks for the fix.

PR #128, opened 2016-02-25
ENH: allow integer shape of empty and zero, close #125

### LINKED ISSUE #125: Simple calls for 1d array construction
Numpy allows both type of calls:

```python
>>> numpy.zeros(3)
array([0, 0, 0])
>>> numpy.zeros([3])
array([0, 0, 0])
```

it would be nice libgpuarray did this aswell, currently the "non-list" case throws an error: "object of type 'int' has no len()".

adler-j:


pygpu/gpuarray.pyx
```diff
@@ -729,13 +729,19 @@ def empty(shape, dtype=GA_DOUBLE, order='C', GpuContext context=None,
     """
     cdef size_t *cdims
 
-    cdims = <size_t *>calloc(len(shape), sizeof(size_t))
+    try:
```
abergeron: Add a `cdef unsigned int nd` to have the right type for the C calls.

abergeron:
Ok, with the fix I indicated.

adler-j:
Fixed the comment

PR #130, opened 2016-02-26
BUG: fix repeat init throwing error, close #129

### LINKED ISSUE #129: Repeat init throws an error
If you call `pygpu.init` repeatedly you get an error:

```
pygpu.gpuarray.GpuArrayException: cannot set while device is active in this process: -1
```

this also applies if you at any point in the same process called `cudaSetDevice` from some other library. This is problematic since it interferes with using libgpuarray together with other cuda libraries. The simple fix is to simply ignore `cudaErrorSetOnActiveProcess`, as is done by other libraries [e.g. ASTRA](https://github.com/astra-toolbox/astra-toolbox/blob/3743fdc534b39958c105f4124ad1130d3e8b042a/cuda/3d/astra3d.cu#L420-L422).

adler-j:
Very minimal pull request, it seems this may have worked for CUDA_VERSION < 7000 before.


abergeron:
This was fixed by #140

PR #135, opened 2016-02-26
Windows fixes

adler-j:
This is very hacky atm, but could perhaps use some input. Some things left to do:

- [ ] Tests still dont run (missing openCL)
- [ ] Verify that this still builds on other platforms
- [ ] I added default includes to the setup.py file. This is nice for windows (no need to twiddle with PATH, install just "works"), but should be customize-able.
- [ ] Update windows install doc
- [ ] This should be rebased

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build
+lib
 .*.sw[po]
 *~
 *.pyc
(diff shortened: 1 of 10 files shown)


abergeron:
For the issue of stdint.h, I would be ready to assume that everybody has 2012 or better and drop the custom stdint since 2010 is no longer publicly supported.

However we can still merge as-is.  Your call.

adler-j:
Valid point, I'll see to it.

abergeron:
Since there has been no activity for a while, I'll just merge this as-is.  We can drop the custom stdint.h later.

PR #136, opened 2016-02-29
C collapse

abergeron:
Add a C API to do dimension collapsing for elemwise (and maybe reduction too).

### DIFF
--- a/src/gpuarray/util.h
+++ b/src/gpuarray/util.h
@@ -88,10 +88,10 @@ GPUARRAY_PUBLIC int gpuarray_type_flags(int init, ...);
  *
  * For scalar arguments, strs[k] can be NULL.
  *
- * \param n The number of arguemnts
+ * \param n The number of arguments
  * \param nd The number of dimensions of all arguments
  * \param dim The compute shape
- * \param strs The strides for all arguemnts
+ * \param strs The strides for all arguments
  *
  */
 GPUARRAY_PUBLIC void gpuarray_elemwise_collapse(unsigned int n,
--- a/src/gpuarray/util.h
+++ b/src/gpuarray/util.h
@@ -75,6 +75,28 @@ GPUARRAY_PUBLIC size_t gpuarray_get_elsize(int typecode);
  */
 GPUARRAY_PUBLIC int gpuarray_type_flags(int init, ...);
 
+/**
+ * Perform dimension collapsing on the specified arguments.
+ *
+ * This function will check for dimension that are next to each other
+ * and contiguous for all inputs and merge them together. This allows
+ * to reduce the complexity of the indexing code in kernels and
+ * therefore enables faster runtime for kernels.
+ *
+ * On return the nd, dims and strs will be overwritten with the
+ * collapsed versions.
+ *
+ * For scalar arguments, strs[k] can be NULL.
+ *
+ * \param n The number of arguemnts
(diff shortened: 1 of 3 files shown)


src/gpuarray/util.h
```diff
@@ -75,6 +75,28 @@ GPUARRAY_PUBLIC size_t gpuarray_get_elsize(int typecode);
...
+ * and contiguous for all inputs and merge them together. This allows
+ * to reduce the complexity of the indexing code in kernels and
+ * therefore enables faster runtime for kernels.
+ *
+ * On return the nd, dims and strs will be overwritten with the
+ * collapsed versions.
+ *
+ * For scalar arguments, strs[k] can be NULL.
+ *
+ * \param n The number of arguemnts
```
adler-j: arguments

abergeron: Fixed, thanks.

src/gpuarray_util.c
```diff
@@ -130,3 +130,45 @@ int gpuarray_type_flags(int init, ...) {
...
+    int collapse = 1;
+    for (k = 0; k < n; k++) {
+      collapse &= (strs[k] == NULL ||
+                   strs[k][i - 1] == dims[i] * strs[k][i]);
+    }
+    if (collapse) {
+      dims[i-1] *= dims[i];
+      shiftdown((ssize_t *)dims, i, nd);
+      for (k = 0; k < n; k++) {
+        if (strs[k] != NULL) {
```
nouiz: This is already done inside shiftdown

nouiz: forget that comment.

tests/check_util.c
```diff
@@ -37,11 +37,73 @@ START_TEST(test_type_flags)
...
+START_TEST(test_elemwise_collapse)
+{
+  size_t dims[2];
+  ssize_t *strs[2];
+  ssize_t _strs0[3];
+  ssize_t _strs1[3];
+  unsigned int nd;
+
+  strs[0] = _strs0;
+  strs[1] = _strs1;
```
nouiz: Can you add a test with all inputs in fortran order?

src/gpuarray_util.c
```diff
@@ -130,3 +130,45 @@ int gpuarray_type_flags(int init, ...) {
   va_end(ap);
   return flags;
 }
+
+static inline void shiftdown(ssize_t *base, unsigned int i, unsigned int nd) {
+  if (base != NULL)
+    memmove(&base[i], &base[i+1], (nd - i - 1)*sizeof(size_t));
+}
+
+void gpuarray_elemwise_collapse(unsigned int n, unsigned int *_nd,
```
nouiz: In the python version, there is a parameter broadcast. When True, it set the strides to 0 for dimensions that need to be broadcasted. Can you confirm this is already done elsewhere and don't need to be replicated here?

Otherwise, this PR seem good. Can you make the PR on Theano that use it? This would allow to test it more before merging.

abergeron: The python version does argument validity checking, this only does dimension collapsing and you should have set the relevant broadcastable dimensions strides to 0 beforehand.

I'm working on the theano part, but it's trickier than I thought.

tests/check_util.c
abergeron: Done.

PR #142, opened 2016-03-18
Batchgemm

cooijmanstim:
Not ready for merge, just to show my work so far.

I've implemented multiple dispatch for `sgemmBatch` and the entry point `rgemmBatch`.

None of this is tested, testing from C seems like pain (as evidenced by the dearth of tests in this repository) so my next step is to implement the Theano Op and add tests in Theano. Once I'm happy with `sgemmBatch` I'll implement `dgemmBatch` and `hgemmBatch` as well.

### DIFF
--- a/pygpu/blas.pyx
+++ b/pygpu/blas.pyx
@@ -17,10 +17,6 @@ cdef extern from "gpuarray/blas.h":
                        double beta, _GpuArray *C, int nocopy)
     int GpuArray_rger(double alpha, _GpuArray *X, _GpuArray *Y, _GpuArray *A,
                       int nocopy)
-    int GpuArray_rgemmBatch(cb_transpose transA, cb_transpose transB,
-                            double alpha, _GpuArray **A, _GpuArray **B,
-                            double beta, _GpuArray **C, size_t batchCount,
-                            int nocopy);
 
 cdef api int pygpu_blas_rgemv(cb_transpose transA, double alpha, GpuArray A,
                               GpuArray X, double beta, GpuArray Y,
--- a/pygpu/blas.pyx
+++ b/pygpu/blas.pyx
@@ -17,6 +17,10 @@ cdef extern from "gpuarray/blas.h":
                        double beta, _GpuArray *C, int nocopy)
     int GpuArray_rger(double alpha, _GpuArray *X, _GpuArray *Y, _GpuArray *A,
                       int nocopy)
+    int GpuArray_rgemmBatch(cb_transpose transA, cb_transpose transB,
+                            double alpha, _GpuArray **A, _GpuArray **B,
+                            double beta, _GpuArray **C, size_t batchCount,
+                            int nocopy);
 
 cdef api int pygpu_blas_rgemv(cb_transpose transA, double alpha, GpuArray A,
                               GpuArray X, double beta, GpuArray Y,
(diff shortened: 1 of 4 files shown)


cooijmanstim:
I redid rgemmBatch and renamed it to rgemmBatch_3d as suggested by @abergeron, implemented dgemmBatch and tested everything from Theano (using the tests added in Theano/Theano#4371).

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -295,44 +287,80 @@ static int sgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
...
-  cuMemFree(Ta);
-  if (err != CUBLAS_STATUS_SUCCESS) {
-    cuda_exit(ctx);
-    if (err == CUBLAS_STATUS_ARCH_MISMATCH)
-      return GA_DEVSUP_ERROR;
-    return GA_BLAS_ERROR;
-  }
+  // use parallel cublasSgemm calls rather than cublasSgemmBatched for large products
+  // (compute products in double because they can be large and we don't need to be exact)
+  const double threshold = 650;
```
abergeron: Why a double here?

pygpu/blas.pyx
```diff
@@ -17,6 +17,10 @@ cdef extern from "gpuarray/blas.h":
                        double beta, _GpuArray *C, int nocopy)
     int GpuArray_rger(double alpha, _GpuArray *X, _GpuArray *Y, _GpuArray *A,
                       int nocopy)
+    int GpuArray_rgemmBatch(cb_transpose transA, cb_transpose transB,
+                            double alpha, _GpuArray **A, _GpuArray **B,
+                            double beta, _GpuArray **C, size_t batchCount,
+                            int nocopy);
```
abergeron: This prototype is both wrong and unused, please remove it.

src/gpuarray_array_blas.c
```diff
@@ -383,3 +384,191 @@ int GpuArray_rger(double alpha, GpuArray *X, GpuArray *Y, GpuArray *A,
...
+  if (!GpuArray_ISONESEGMENT(A)) {
+    if (nocopy)
+      return GA_COPY_ERROR;
+    else {
+      err = GpuArray_copy(&copyA, A, GA_F_ORDER);
+      if (err != GA_NO_ERROR)
+	goto cleanup;
+      Ap = &copyA;
+    }
+  }
```
abergeron: This check is ok, if a bit restrictive.  It should be ok to have a non-contiguous A as long as the last two dims are.  If you don't want to do this for now that's ok but please add a comment saying that for future reference.

abergeron: The same applies for B and C

src/gpuarray_blas_cuda_cublas.c
cooijmanstim: Did you see the comment? Should I use a 64-bit integer instead?

abergeron: 64-bit integer (or size_t) will be big enough.  Otherwise the array wouldn't fit on the GPU.

cooijmanstim: Right.

src/gpuarray_array_blas.c
cooijmanstim: I added a note.

PR #144, opened 2016-03-30
Native f16

abergeron:
This is the base to eventually have transparent support for storage-only or full float16.

### DIFF
--- a/src/gpuarray/buffer.h
+++ b/src/gpuarray/buffer.h
@@ -568,6 +568,12 @@ typedef struct _gpuarray_buffer_ops {
  */
 #define GA_CTX_PROP_FREE_GMEM 10
 
+/**
+ * Get the status of native float16 support on the device.
+ *
+ * Type: `int`
+ */
+#define GA_CTX_PROP_NATIVE_FLOAT16   11
 
 /* Start at 512 for GA_BUFFER_PROP_ */
 /**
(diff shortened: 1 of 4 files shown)


nouiz:
```
$ make test
Running tests...
    Start 1: test_types
1/5 Test #1: test_types .......................   Passed    0.03 sec
    Start 2: test_util
2/5 Test #2: test_util ........................   Passed    0.02 sec
    Start 3: test_array
3/5 Test #3: test_array .......................***Failed    0.07 sec
    Start 4: test_error
4/5 Test #4: test_error .......................   Passed    0.02 sec
    Start 5: test_buffer
5/5 Test #5: test_buffer ......................   Passed    2.53 sec

80% tests passed, 1 tests failed out of 5

Total Test time (real) =   2.68 sec

The following tests FAILED:
          3 - test_array (Failed)
Errors while running CTest
```

This also fail with current master, so I merge now.

danielhauagge:
I was having issues with the same test and solved it by setting the env variable GPUARRAY_TEST_DEVICE to cuda. I couldn't find this documented anywhere and it only occurs in check_array.c.

nouiz:
Thanks.

We can set that variable or DEVICE. The problem was that when I use
DEVICE=cuda, the tests picked the first GPU. If it isn't available, it
fail. If I specify which GPU to use with DEVICE=cudaN, then it works.

PR #149, opened 2016-04-12
Batch blas ops

abergeron:
This implements very specialized kernels for gemvBatch and gerBatch (within a generic interface) to support the blocksparse op from Theano.

We plan to update the implementation later to support the full interface.

### DIFF
--- a/src/gpuarray_blas_cuda_cublas.c
+++ b/src/gpuarray_blas_cuda_cublas.c
@@ -34,9 +34,9 @@ typedef struct _blas_handle {
 
 static const char *code_sgemvBH_N_a1_b1_small =                         \
   "__global__ void _sgemv(const float *A[], size_t lda, "               \
-  "                       const float *x[], size_t incX, "              \
+  "                       const float *x[], size_t incx, "              \
   "                       float *y[], size_t incy, "                    \
-  "                       ga_size b, ga_size m, ga_size n) {"           \
+  "                       size_t b, size_t m, size_t n) {"              \
   "  for (size_t p = blockIdx.y * blockDim.y + threadIdx.y; p < b;"     \
   "       p += gridDim.y * blockDim.y) {"                               \
   "    for (size_t i = blockIdx.x * blockDim.x + threadIdx.x; i < m;"   \
@@ -57,9 +57,9 @@ static const char *code_sgemvBH_N_a1_b1_small =                         \
 
 static const char *code_sgemvBH_T_a1_b1_small =                 \
   "__global__ void _sgemv(const float *A[], size_t lda, "       \
-  "                       const float *x[], size_t incX, "      \
+  "                       const float *x[], size_t incx, "      \
   "                       float *y[], size_t incy, "            \
-  "                       ga_size b, ga_size m, ga_size n) {"   \
+  "                       size_t b, size_t m, size) n) {"       \
   "  size_t i = blockIdx.x * blockDim.x + threadIdx.x;"         \
   "  size_t p = blockIdx.y * blockDim.y + threadIdx.y;"         \
   "  if (i >= m || p >= b) return;"                             \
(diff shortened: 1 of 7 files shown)


src/gpuarray_blas_cuda_cublas.c
```diff
@@ -528,6 +763,269 @@ static int dgemv(cb_order order, cb_transpose transA, size_t M, size_t N,
...
+                      size_t batchCount, int flags) {
+  return GA_DEVSUP_ERROR;
+}
+
+static int sgemvBatch(cb_order order, cb_transpose transA,
+                      size_t M, size_t N, float alpha,
+                      gpudata **A, size_t *offA, size_t lda,
+                      gpudata **x, size_t *offX, size_t incX,
+                      float beta, gpudata **y, size_t *offY, size_t incY,
+                      size_t batchCount, int flags) {
```
nouiz: Why sgemvBatch have a flag attribute that isn't used which sgemmBatch don't have one? What possible futur utility do you see? Idem for all new ?gemvBatch and ?gerBatch method

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -528,6 +763,269 @@ static int dgemv(cb_order order, cb_transpose transA, size_t M, size_t N,
...
+  if (err != GA_NO_ERROR) {
+    cuda_exit(ctx);
+    return err;
+  }
+
+
+  for (i = 0; i < batchCount; i++) {
+    cuda_record(A[i], CUDA_WAIT_READ);
+    cuda_record(x[i], CUDA_WAIT_READ);
+    cuda_record(y[i], CUDA_WAIT_READ|CUDA_WAIT_WRITE);
```
nouiz: Out of curiosity, doing all those event, could this slow down computation? I won't delay this PR for that. Just wanted to know.

src/gpuarray_buffer_cuda.c
```diff
@@ -42,7 +42,7 @@ void *cuda_make_ctx(CUcontext ctx, int flags) {
   cuda_context *res;
   void *p;
 
-  res = malloc(sizeof(*res));
+  res = calloc(1, sizeof(*res));
```
nouiz: why this change? It should be fine.

src/gpuarray_blas_cuda_cublas.c
abergeron: When I did sgemmBatch it was a mirror to the one in cublas.

But for sgemvBatch, it doesn't act the same since I use atomics and don't cover all cases.  My idea was to leave that in the interface for now so that we can later introduce different values for the flags that would modify the behavior.  This would avoid breaking the interface again in the future.

src/gpuarray_blas_cuda_cublas.c
abergeron: It does take a bit of time to do this, but I have to to keep track of which buffer is used where for the memory allocator and the (future) concurrent kernel feature.

src/gpuarray_buffer_cuda.c
abergeron: This gets rid of undefined memory usage in the cache for kernels with regards to the bin_id.  Since it might not be written to completely, this makes sure that the unwritten parts are zeroed.

abergeron:
Rebase done.

abergeron:
Rebase done.

nouiz:
After fixing the installation problem, I get this batch of error:

ERROR: pygpu.tests.test_blas.test_ger(4, 5, 'float32', 'f', 1, 1, False, True)
  File "/Tmp/lisa/os_v5/anaconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/u/bastienf/.local/lib64/python2.7/site-packages/pygpu-0.2.1-py2.7-linux-x86_64.egg/pygpu/tests/test_blas.py", line 155, in ger
    gr = gblas.ger(1.0, gX, gY, gA, overwrite_a=overwrite)
  File "pygpu/blas.pyx", line 127, in pygpu.blas.ger (pygpu/blas.c:2907)
  File "pygpu/blas.pyx", line 44, in pygpu.blas.pygpu_blas_rger (pygpu/blas.c:1655)
GpuArrayException: ('Device does not support operation', 8)

ERROR: pygpu.tests.test_blas.test_ger(4, 5, 'float32', 'f', 1, 1, False, False)
  File "/Tmp/lisa/os_v5/anaconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/u/bastienf/.local/lib64/python2.7/site-packages/pygpu-0.2.1-py2.7-linux-x86_64.egg/pygpu/tests/test_blas.py", line 155, in ger
    gr = gblas.ger(1.0, gX, gY, gA, overwrite_a=overwrite)
  File "pygpu/blas.pyx", line 127, in pygpu.blas.ger (pygpu/blas.c:2907)
  File "pygpu/blas.pyx", line 44, in pygpu.blas.pygpu_blas_rger (pygpu/blas.c:1655)
GpuArrayException: ('Device does not support operation', 8)

nouiz:
https://github.com/abergeron/libgpuarray/pull/2

abergeron:
With your diff, I don't get any errors on the tests for blas.  Are you passing DEVICE=cuda0 to the tests?

nouiz:
I recompiled everything from scratch with an clean environment and now it work. So merging.

PR #151, opened 2016-04-15
Context manager for cuda contexts

abergeron:
This enables us to use other cuda-specific libraries by setting a context on the cuda context stack for some time.  This is enough to work with pycuda.


nouiz:
Seem good, but this need at least a test.

abergeron:
I don't really see how to test it.

nouiz:
Make an op that use it with pycuda? That is the goal I think. The computation inside could be trivial, just to make sure the mechanism work.

If too long, just an op with a perform that call it to make sure it don't crash would be better then nothing.

abergeron:
Sure we can do that in Theano, but this PR is for libgpuarray.  In fact that was the plan for the fft ops.

But this PR is groundwork in libgpuarray for that.  So I can't make a test in here.

nouiz:
Right. If you prefer to merge now with only a code review and nothing else, we can do it and let the Theano PR also test this one. Otherwise, we can wait for that PR to be advanced enough to test it a little before mergning.

If you prefer to merge now, you can do it.

abergeron:
I did test manually and it works to initialize a pycuda context, so I'll merge it for now to facilitate use.

PR #154, opened 2016-04-18
Add properties to expose the individual limits on each of the dimensions for grid/block sizes

abergeron:


src/gpuarray_buffer_cuda.c
```diff
@@ -1895,6 +1895,102 @@ static int cuda_property(void *c, gpudata *buf, gpukernel *k, int prop_id,
...
+    if (ctx->err != CUDA_SUCCESS) {
+      cuda_exit(ctx);
+      return GA_IMPL_ERROR;
+    }
+    ctx->err = cuDeviceGetAttribute(&i, CU_DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_Z, id);
+    if (ctx->err != CUDA_SUCCESS) {
+      cuda_exit(ctx);
+      return GA_IMPL_ERROR;
+    }
+    cuda_exit(ctx);
```
nouiz: Just a comment. cuda_exit(ctx) could be before the if to make the code more compact It would be equivalent. No need to change it.

src/gpuarray_buffer_opencl.c
```diff
@@ -1293,6 +1293,87 @@ static int cl_property(void *c, gpudata *buf, gpukernel *k, int prop_id,
...
+  case GA_CTX_PROP_MAXLSIZE1:
+    ctx->err = clGetContextInfo(ctx->ctx, CL_CONTEXT_DEVICES, sizeof(id),
+                                &id, NULL);
+    if (ctx->err != CL_SUCCESS)
+      return GA_IMPL_ERROR;
+    ctx->err = clGetDeviceInfo(id, CL_DEVICE_MAX_WORK_ITEM_SIZES, 0, NULL,
+                               &sz);
+    if (ctx->err != CL_SUCCESS)
+      return GA_IMPL_ERROR;
+    psz = malloc(sz);
```
nouiz: You can use alloca more frequently to lower the number of call to malloc/free.

Also, you seem to duplicate the code pretty heavily here. Maybe next time, refactor it with a service function.

PR #157, opened 2016-04-22
Fix Memory leak

nouiz:
fix https://github.com/Theano/Theano/issues/4318

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -80,6 +80,7 @@ computer, you can install them in your home directory like this:
 ::
 
   cd <dir>
+  rm -rf build Build
   mkdir Build
   cd Build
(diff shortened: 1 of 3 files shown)


doc/installation.rst
```diff
@@ -80,11 +80,13 @@ computer, you can install them in your home directory like this:
 ::
 
   cd <dir>
+  rm -rf build Build
```
abergeron: This is not a mandatory step.

abergeron:
I'm not sure about the change in install instructions, but the fix for the leak is ok.

doc/installation.rst
nouiz: Without this, if we compiled it in the past, it can generate corrupted stuff. We get hit ourself by this and this is was reported by user.

abergeron: I never had a problem reusing the Build directory, but I did have some problems with build.  So ok, I'll merge it like this.

PR #158, opened 2016-04-22
Change the interface of GpuArray_fromdata

nouiz:
I checked we don't use *fromdata in Theano and I think I updated all use of it in libgpuarray.

Should I update one of the version?

### DIFF
--- a/pygpu/tools.py
+++ b/pygpu/tools.py
@@ -127,6 +127,7 @@ def check_args(args, collapse=False, broadcast=False):
     if dims is None:
         raise TypeError("No arrays in kernel arguments, "
                         "something is wrong")
+    tdims = dims
 
     if broadcast or collapse:
         # make the strides and dims editable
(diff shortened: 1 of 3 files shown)


pygpu/tools.py
```diff
@@ -127,15 +127,16 @@ def check_args(args, collapse=False, broadcast=False):
...
 
     if broadcast or collapse:
         # make the strides and dims editable
         dims = list(dims)
         strs = [list(str) if str is not None else str for str in strs]
 
     if broadcast:
+        # Set strides to 0s when needed.
         # Get the full shape in dims (no ones unless all arrays have it).
+        tdims = dims
```
abergeron: This change is bad because now tdims will sometimes not be defined and also not be a tuple.

abergeron:
I'm om with the first commit, but not the second one.

pygpu/tools.py
nouiz: tdims is only used in this if. So it will always be defined in the part needed. I'll move it back to have it stay a tuple.

nouiz:
I did the fix.

PR #159, opened 2016-04-22
Work around the 32-bit signed /  64-bit unsigned bug in CUDA.

abergeron:
This has now resurged because of the call32 option.

### DIFF
--- a/tests/check_elemwise.c
+++ b/tests/check_elemwise.c
@@ -327,6 +327,76 @@ START_TEST(test_basic_collapse)
 }
 END_TEST
 
+START_TEST(test_basic_neg_strides)
+{
+  GpuArray a;
+  GpuArray b;
+  GpuArray c;
+
+  GpuElemwise *ge;
+
+  static const uint32_t data1[6] = {1, 2, 3, 4, 5, 6};
+  static const uint32_t data2[6] = {7, 8, 9, 10, 11, 12};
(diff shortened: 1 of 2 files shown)


abergeron:
I've added a test for that case.

tests/check_elemwise.c
```diff
@@ -327,6 +327,76 @@ START_TEST(test_basic_collapse)
...
+
+  ga_assert_ok(GpuArray_empty(&a, ops, ctx, GA_UINT, 1, dims, GA_C_ORDER));
+  ga_assert_ok(GpuArray_write(&a, data1, sizeof(data1)));
+
+  ga_assert_ok(GpuArray_empty(&b, ops, ctx, GA_UINT, 1, dims, GA_C_ORDER));
+  ga_assert_ok(GpuArray_write(&b, data2, sizeof(data2)));
+
+  starts[0] = 5;
+  stops[0] = -1;
+  steps[0] = -1;
```
nouiz: In Python, using [5:-1:-1] will return an view of 0 elements. Why this isn't the python slice operator? What it is?

abergeron: This does not use the python convention because it's not python ...

Specifically it does not do negative indexing like python so this actually selects 6 elements.

The pygpu wrapper does the proper translation so that you get the usual python semantics

PR #165, opened 2016-05-09
Align allocations to 64 bytes instead of 16.

abergeron:
Apparently that is what cuMemAlloc does and I read the docs wrong.


nouiz:
cnmem use 512, so 64 shouldn't be a problem!

PR #166, opened 2016-05-09
Use 32-bit addressing in take1 when possible

abergeron:
This be faster and doesn't change the public interface.

### DIFF
--- a/src/gpuarray_array.c
+++ b/src/gpuarray_array.c
@@ -382,7 +382,7 @@ static int gen_take1_kernel(GpuKernel *k, const gpuarray_buffer_ops *ops,
       strb_appendf(&sb, "      p += pos * (%s)s%u;\n", ssz, i);
     }
   }
-  strb_appendf(&sb, "      r[i0*((%s)n1) + i1] = (((GLOBAL_MEM %s *)(GLOBAL_MEM char *)v) + p);\n",
+  strb_appendf(&sb, "      r[i0*((%s)n1) + i1] = ((GLOBAL_MEM %s *)(((GLOBAL_MEM char *)v) + p));\n",
                sz, gpuarray_get_type(v->typecode)->cluda_name);
   strb_appends(&sb, "    }\n"
                "  }\n"
--- a/src/gpuarray_array.c
+++ b/src/gpuarray_array.c
@@ -328,7 +328,7 @@ static int gen_take1_kernel(GpuKernel *k, const gpuarray_buffer_ops *ops,
 
   if (addr32) {
     sz = "ga_uint";
-    ssz "ga_int";
+    ssz = "ga_int";
   } else {
     sz = "ga_size";
     ssz = "ga_ssize";
@@ -369,7 +369,7 @@ static int gen_take1_kernel(GpuKernel *k, const gpuarray_buffer_ops *ops,
                "    }\n"
                "    pos0 += ii0 * (%s)s0;\n"
                "    for (i1 = idx1; i1 < n1; i1 += numThreads1) {\n"
-               "      %s p = pos0;\n", ssz, sz, sz, ssz, sz);
+               "      %s p = pos0;\n", ssz, sz, sz, sz);
   if (v->nd > 1) {
     strb_appendf(&sb, "      %s pos, ii = i1;\n", sz);
     for (i2 = v->nd; i2 > 1; i2--) {
@@ -382,8 +382,8 @@ static int gen_take1_kernel(GpuKernel *k, const gpuarray_buffer_ops *ops,
       strb_appendf(&sb, "      p += pos * (%s)s%u;\n", ssz, i);
     }
   }
-  strb_appendf(&sb, "      r[i0*((%s)n1) + (%s)i1] = (((GLOBAL_MEM %s *)(GLOBAL_MEM char *)v) + p);\n",
-               sz, sz, gpuarray_get_type(v->typecode)->cluda_name);
+  strb_appendf(&sb, "      r[i0*((%s)n1) + i1] = (((GLOBAL_MEM %s *)(GLOBAL_MEM char *)v) + p);\n",
+               sz, gpuarray_get_type(v->typecode)->cluda_name);
   strb_appends(&sb, "    }\n"
                "  }\n"
                "}\n");
@@ -411,7 +411,6 @@ int GpuArray_take1(GpuArray *a, const GpuArray *v, const GpuArray *i,
   size_t argp;
   GpuKernel k;
   unsigned int j;
-  unsigned int _n[2], _o;
   int err, kerr = 0;
   int addr32 = 0;
 
@@ -443,7 +442,7 @@ int GpuArray_take1(GpuArray *a, const GpuArray *v, const GpuArray *i,
     n[1] *= v->dimensions[j];
   }
 
-  if (n[0] * n[1] < INT_MAX) {
+  if (n[0] * n[1] < SADDR32_MAX) {
     addr32 = 1;
   }
 
--- a/src/gpuarray_array.c
+++ b/src/gpuarray_array.c
@@ -382,7 +382,7 @@ static int gen_take1_kernel(GpuKernel *k, const gpuarray_buffer_ops *ops,
       strb_appendf(&sb, "      p += pos * (%s)s%u;\n", ssz, i);
     }
   }
-  strb_appendf(&sb, "      r[i0*((%s)n1) + i1] = ((GLOBAL_MEM %s *)(((GLOBAL_MEM char *)v) + p));\n",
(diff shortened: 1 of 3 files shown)


src/gpuarray_array.c
```diff
@@ -400,11 +408,12 @@ int GpuArray_take1(GpuArray *a, const GpuArray *v, const GpuArray *i,
 #if DEBUG
   char *errstr = NULL;
 #endif
-  void **args = NULL;
   size_t argp;
   GpuKernel k;
   unsigned int j;
+  unsigned int _n[2], _o;
```
nouiz: Where this is used? I don't see it being used.

src/gpuarray_array.c
```diff
@@ -338,44 +347,43 @@ static int gen_take1_kernel(GpuKernel *k, const gpuarray_buffer_ops *ops,
...
+                     "      ii /= (%s)d%u;\n", sz, i, sz, i);
       else
         strb_appends(&sb, "      pos = ii;\n");
-      strb_appendf(&sb, "      p += pos * s%u;\n", i);
+      strb_appendf(&sb, "      p += pos * (%s)s%u;\n", ssz, i);
     }
   }
-  strb_appendf(&sb, "      r[i0*n1 + i1] = ((GLOBAL_MEM %s *)p)[0];\n",
-               gpuarray_get_type(v->typecode)->cluda_name);
+  strb_appendf(&sb, "      r[i0*((%s)n1) + (%s)i1] = (((GLOBAL_MEM %s *)(GLOBAL_MEM char *)v) + p);\n",
```
nouiz: The 2 first cast are useless.

src/gpuarray_array.c
```diff
@@ -338,44 +347,43 @@ static int gen_take1_kernel(GpuKernel *k, const gpuarray_buffer_ops *ops,
...
+               "    if ((ii0 < 0) || (ii0 >= d0)) {\n"
                "      *err = -1;\n"
                "      continue;\n"
                "    }\n"
-               "    pos0 += ii0 * s0;\n"
+               "    pos0 += ii0 * (%s)s0;\n"
                "    for (i1 = idx1; i1 < n1; i1 += numThreads1) {\n"
-               "      GLOBAL_MEM ga_byte *p = (GLOBAL_MEM ga_byte *)v;\n"
-               "      p += pos0;\n");
+               "      %s p = pos0;\n", ssz, sz, sz, ssz, sz);
```
nouiz: There is one extra parameter here. I think it should be:
ssz, sz, sz, sz

src/gpuarray_array.c
```diff
@@ -338,44 +347,43 @@ static int gen_take1_kernel(GpuKernel *k, const gpuarray_buffer_ops *ops,
...
   if (v->nd > 1) {
-    strb_appends(&sb, "      ga_size pos, ii = i1;\n");
+    strb_appendf(&sb, "      %s pos, ii = i1;\n", sz);
     for (i2 = v->nd; i2 > 1; i2--) {
       i = i2 - 1;
       if (i > 1)
-        strb_appendf(&sb, "      pos = ii %% d%u;\n"
```
nouiz: What the d in this line and the one bellow mean? Is there a special meaning for strb_appendf?

src/gpuarray_array.c
nouiz: Why this extra cast to char*? Why cast n1 an i1 ? They are already of type sz.

nouiz:
I finished my review

src/gpuarray_array.c
abergeron: I need to cast `n1` because it is ga_size, but `i1` is not required.

Also the cast to char * is so that we can add a byte offset to the pointer.

src/gpuarray_array.c
abergeron: Leftover from previous code.

src/gpuarray_array.c
abergeron: Yes.

src/gpuarray_array.c
abergeron: No, it will just be a literal 'd' followed by the formatted number.  It's the name of one of the inputs.

abergeron:
I should have fixed all the problems.

lamblin:
Running `make tests` I have:
```
3/6 Test #3: test_array .......................***Failed    0.52 sec
```
Not sure how to get the full information.

I'm running the benchmarks to check the performance.

nouiz:
This is probably just that the GPU is used. those tests don't handle that
well. Use the env var DEVICE=cudaN with the right number.

lamblin:
Yes, I did that, I actually ran `DEVICE=cuda3 make test`, and checked with `nvidia-smi` that it was running on the right gpu.

nouiz:
It seem a bug, @abergeron will check.

lamblin:
Strange error when running the benchmark:
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
/home/benchmark/code/rnn_exps/theano/ptb/train_lm.py in <module>()
     23     config = getattr(config_lm, args.proto)()
     24     logger.info("Model options:\n{}".format(pprint.pformat(config)))
---> 25     train(**config)

/home/benchmark/code/rnn_exps/theano/ptb/lm.pyc in train(dim_word, dim, encoder, max_epochs, finish_after, dispFreq, decay_c, lrate, n_words, maxlen, batch_size, valid_batch_size, max_grad_norm, nlayers, data_path, use_dropout)
    596 
    597             # compute cost, grads and copy grads to shared variables
--> 598             cost = f_grad_shared(x)
    599 
    600             # do the update on parameters

/home/benchmark/repos/Theano/theano/compile/function_module.pyc in __call__(self, *args, **kwargs)
    906                     node=self.fn.nodes[self.fn.position_of_error],
    907                     thunk=thunk,
--> 908                     storage_map=getattr(self.fn, 'storage_map', None))
    909             else:
    910                 # old-style linkers raise their own exceptions

/home/benchmark/repos/Theano/theano/gof/link.pyc in raise_with_op(node, thunk, exc_info, storage_map)
    312         # extra long error message in that case.
    313         pass
--> 314     reraise(exc_type, exc_value, exc_trace)
    315 
    316 

/home/benchmark/repos/Theano/theano/compile/function_module.pyc in __call__(self, *args, **kwargs)
    893         try:
    894             outputs =\
--> 895                 self.fn() if output_subset is None else\
    896                 self.fn(output_subset=output_subset)
    897         except Exception:

RuntimeError: Success
Apply node that caused the error: GpuAdvancedSubtensor1(Wemb, GpuReshape{1}.0)
Toposort index: 39
Inputs types: [GpuArrayType<None>(float32, (False, False)), GpuArrayType<None>(int64, (False,))]
Inputs shapes: [(10000, 200), (400,)]
Inputs strides: [(800, 4), (8,)]
Inputs values: ['not shown', 'not shown']
Outputs clients: [[GpuReshape{3}(GpuAdvancedSubtensor1.0, MakeVector{dtype='int64'}.0)]]
```

abergeron:
Problem should be fixed.

lamblin:
The code seems to run. I still have to relaunch with profiling and compare.

lamblin:
Updated timings with that change below.
TL;DR: It helps, but it's not all of it.

model | op | old | new | this PR
------|----|----|-----|---------------------------
small | GpuAdvancedSubtensor1 | 2.2 | 16 | 9.8
large | GpuAdvancedSubtensor1 | 1.1 | 10.5 | 8.2
small | GpuAdvancedIncSubtensor1 | 77 | 209 | 131
large | GpuAdvancedIncSubtensor1 | 75 | 204 | 129
small | GpuAdvancedIncSubtensor1_dev20 | 0.861 | 0.68 | 0.475
large | GpuAdvancedIncSubtensor1_dev20 | 0.518 | 0.438 | 0.370

nouiz:
should we merge this? I think what @abergeron told about making GpuJoin reuse elemwise to get all the speed benefit would be useful here too. If so, this would make this PR useless.

But if we merge now, we will have now some good speed up. So if people absolutly need float16, they won't have too much slow down.

I vote to merge now.

lamblin:
I did not review the code and I'm not sure I understand how the other solution would interact with this one. I'm OK with merging if you think it is a good idea.

lamblin:
Btw, the overall performance was similar to the old back-end, due to Gemm being faster.

nouiz:
I reviewed it complety and I think it is good to merge.

This could use the GpuElemwise kernel to make a copy for each element, but mark them as being able to run in parallel.

But it would be probably better to implement the dimensions collapsing here. There is function that can be reused and this would make only 1 kernel call.

nouiz:
I need to leave, When I can I'll run gpuarray and Theano tests to make sure it work well and I'll merge.

lamblin:
I can run the tests.

nouiz:
I'm running them.

lamblin:
Me too. The gpuarray tests passed, I'm running the Theano ones now.

nouiz:
The Theano one passed, I'm finishing the gpuarray ones. so merging.

PR #168, opened 2016-05-10
Special case axis 0 to use the faster move instead of extcopy in concatenate.

abergeron:
There still remains to speedup extcopy for the other cases, but this would require a more significant time investment.

### DIFF
--- a/src/gpuarray_array.c
+++ b/src/gpuarray_array.c
@@ -941,8 +941,8 @@ int GpuArray_split(GpuArray **rs, const GpuArray *a, size_t n, size_t *p,
 int GpuArray_concatenate(GpuArray *r, const GpuArray **as, size_t n,
                          unsigned int axis, int restype) {
   size_t *dims;
-  size_t i;
-  size_t res_off;
+  size_t i, j;
+  size_t res_off, sz;
   unsigned int p;
   int err = GA_NO_ERROR;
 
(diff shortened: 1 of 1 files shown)


src/gpuarray_array.c
```diff
@@ -988,19 +988,27 @@ int GpuArray_concatenate(GpuArray *r, const GpuArray **as, size_t n,
...
-    err = r->ops->buffer_extcopy(as[i]->data, as[i]->offset, r->data,
-                                 res_off, as[i]->typecode, r->typecode,
-                                 as[i]->nd, as[i]->dimensions,
-                                 as[i]->strides, r->nd, as[i]->dimensions,
-                                 r->strides);
+    if (axis == 0 &&
+        GpuArray_IS_C_CONTIGUOUS(as[i]) &&
+        as[i]->typecode == r->typecode) {
+      sz = gpuarray_get_elsize(r->typecode);
+      for (j = 0; j < r->nd; j++) sz *= as[i]->dimensions[j];
```
nouiz: I don't see any shapes or nd check anywhere related to concatenate. Why this is missing?

I would more use `as[i]->nd` then `r->nd` in that case. To not read bad data.

abergeron: This is checked before.

nouiz: Where? line 975 check the size of each dim, but nd isn't checked for equality. So if nd isn't good, we could detect it with reading bad memory, but it could also pass.

abergeron: Line 970 checks for nd equality.

abergeron:
This could still go in but it make about no difference in the timing for concatenate.  However a hackish version that replaces the custom extcopy with one based on GpuElemwise is about 20 times faster, so we should probably go with that instead.

src/gpuarray_array.c
nouiz: ok

nouiz:
We need to confirm, but I would go to not merge this PR if it don't make a
speed difference. I would go to make the other change in the short term.

abergeron:
ok

PR #174, opened 2016-05-12
First part of the rework to support a faster extcopy.

abergeron:
This creates a struct gpucontext which holds part of the information for the gpu context.

There are no ABI changes for compiled programs since the old struct was opaque anyway and it is still handled everywhere as a pointer.  But now it allows us to add things that are shared for all contexts (such as a GpuElemwise instance).

There is no such change yet as this is mostly a sweeping rename to make sure nothing broke at this step.

PR #175, opened 2016-05-13
New buffer api

abergeron:
This includes #174, which should be merged first.

This rework is more significant and does cause changes to have to be made downstream, but after this, the API should be much more stable.

### DIFF
--- a/pygpu/gpuarray.pxd
+++ b/pygpu/gpuarray.pxd
@@ -142,6 +142,7 @@ cdef extern from "gpuarray/array.h":
 
     bint GpuArray_CHKFLAGS(_GpuArray *a, int fl)
     bint GpuArray_ISONESEGMENT(_GpuArray *a)
+    bint GpuArray_IS_C_CONTIGUOUS(_GpuArray *a)
 
     ctypedef enum ga_order:
         GA_ANY_ORDER, GA_C_ORDER, GA_F_ORDER
--- a/pygpu/gpuarray.pxd
+++ b/pygpu/gpuarray.pxd
@@ -70,11 +70,13 @@ cdef extern from "gpuarray/buffer.h":
     ctypedef struct gpukernel:
         pass
 
-    ctypedef struct gpuarray_buffer_ops:
-        gpucontext *buffer_init(int devno, int flags, int *ret)
-        void buffer_deinit(gpucontext *ctx)
-        char *ctx_error(gpucontext *ctx)
-        int property(gpucontext *c, gpudata *b, gpukernel *k, int prop_id, void *res)
+    gpucontext *gpucontext_init(const char *name, int devno, int flags, int *ret)
+    void gpucontext_deref(gpucontext *ctx)
+    char *gpucontext_error(gpucontext *ctx, int err)
+    int gpucontext_property(gpucontext *ctx, int prop_id, void *res)
+    int gpukernel_property(gpukernel *k, int prop_id, void *res)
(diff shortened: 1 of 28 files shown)


pygpu/gpuarray.pyx
```diff
@@ -955,40 +935,33 @@ cdef class GpuContext:
...
             else:
-                raise get_exc(err), self.ops.ctx_error(NULL) + ": " + str(devno)
+                raise get_exc(err), gpucontext_error(NULL, err) + ": " + str(devno)
 
     def __enter__(self):
         if cuda_enter == NULL:
             raise RuntimeError("cuda_enter not available")
         if cuda_exit == NULL:
             raise RuntimeError("cuda_exit not available")
-        if self.ops != gpuarray_get_ops("cuda"):
```
nouiz: You still need a check like this one here.

pygpu/gpuarray.pyx
```diff
@@ -1517,8 +1488,11 @@ cdef class GpuArray:
         """
         return pygpu_copy(self, to_ga_order(order))
 
-    def transfer(self, GpuContext new_ctx, share=False):
-        return pygpu_transfer(self, new_ctx, share)
+    def transfer(self, GpuContext new_ctx):
+        cdef GpuArray r
+        r = pygpu_empty(self.ga.nd, self.ga.dimensions, self.ga.typecode,
+                        GA_C_ORDER, new_ctx, None)
```
nouiz: If the source if f contiguous?

pygpu/gpuarray.pyx
abergeron: ok

pygpu/gpuarray.pyx
abergeron: I've added a check and a case for that.

pygpu/gpuarray.pxd
```diff
@@ -142,6 +142,7 @@ cdef extern from "gpuarray/array.h":
 
     bint GpuArray_CHKFLAGS(_GpuArray *a, int fl)
     bint GpuArray_ISONESEGMENT(_GpuArray *a)
+    bint GpuArray_IS_C_CONTIGUOUS(_GpuArray *a)
```
nouiz: Can you implement IS_F_CONTIGUOUS at the same time?

abergeron: It is already implemented, but we don't need it for the python module.

nouiz:
Except for the last commit, all the commis are in gh-178. I would make a new PR with just the last commit and close this PR. Then we can merge gh-178 and https://github.com/Theano/Theano/pull/4542. Then after we can merge the blas change with another bump in the version number with a new corresponding Theano PR.

abergeron:
I would rather rebase #178 with everything if you want to do this way,

pygpu/gpuarray.pxd
nouiz: Why we don't need it in the python module? I understand that currently it won't be used, but if someone make a new Theano op and need this, I don't want him to need to make change to libgpuarray.

nouiz:
ok. the only commit here also break the interface, so doing all this in one PR would only break the interface once.

pygpu/gpuarray.pxd
abergeron: In Theano we can use the C function from libgpuarray directly.

abergeron:
Integrated into #178.

PR #167, opened 2016-05-09
Update the call32 mode of GpuElemwise to have the same kernel interface.

abergeron:
This should update the call32 interface correctly, I'll slip in the change to preallocate the necessary buffers for collapsing tomorrow.

PR #178, opened 2016-05-16
Move extcopy out the backends

abergeron:
Now we implement extcopy with a GpuElemwise instance and it runs about 20x faster.  The code is also much less complex.

This depends on #174 #167 and #175, which should be merged before and this one will be rebased after.


nouiz:
This need a rebase.

src/gpuarray_elemwise.c
```diff
@@ -16,15 +18,20 @@ struct _GpuElemwise {
   GpuKernel *k_basic_32;
   size_t *dims;
   ssize_t **strides;
-  unsigned int n;
   unsigned int nd;
+  unsigned int n;
+  unsigned int narray;
```
nouiz: Add comments to tell what n and narray are.

src/gpuarray_elemwise.c
```diff
@@ -72,6 +79,41 @@ static void free_args(unsigned int n, gpuelemwise_arg *args) {
...
+  if (res == NULL) return 1;
+  memset(res + (e*o), 0, e*(n-o));
+  *p = (void *)res;
+  return 0;
+}
+
+static int ge_grow(GpuElemwise *ge, unsigned int nd) {
+  unsigned int i;
+
+  assert(nd > ge->nd);
```
nouiz: assert in c are very bad. This give completly useless user error message. Why not just return 2 and make the calling code change the returned user error message?

src/gpuarray_elemwise.c
```diff
@@ -72,6 +79,41 @@ static void free_args(unsigned int n, gpuelemwise_arg *args) {
...
+
+  if (reallocaz((void **)&ge->k_basic, sizeof(GpuKernel), ge->nd, nd) ||
+      reallocaz((void **)&ge->k_basic_32, sizeof(GpuKernel), ge->nd, nd) ||
+      reallocaz((void **)&ge->dims, sizeof(size_t), ge->nd, nd))
+    return 1;
+  for (i = 0; i < ge->narray; i++) {
+    if (reallocaz((void **)&ge->strides[i], sizeof(ssize_t), ge->nd, nd))
```
nouiz: You must also grow ge->strides.

src/gpuarray_elemwise.c
```diff
@@ -72,6 +79,41 @@ static void free_args(unsigned int n, gpuelemwise_arg *args) {
...
+      return 1;
+  }
+  ge->nd = nd;
+  return 0;
+}
+
```
nouiz: bellow in the fct  gen_elemwise_basic_kernel, document the nd and n parameter.

src/gpuarray_elemwise.c
```diff
@@ -522,18 +577,24 @@ GpuElemwise *GpuElemwise_new(gpucontext *ctx,
...
+  for (i = 0; i < res->n; i++)
+    if (is_array(res->args[i])) res->narray++;
+
+  res->nd = 16;
+  while (res->nd < nd) res->nd *= 2;
+  res->dims = calloc(res->nd, sizeof(size_t));
   if (res->dims == NULL)
     goto fail_dims;
-  res->strides = strides_array(n, nd);
+  res->strides = strides_array(res->narray, res->nd);
```
nouiz: don't forget to remove the +1 on line 234

src/gpuarray/elemwise.h
```diff
@@ -81,6 +81,7 @@ GPUARRAY_PUBLIC int GpuElemwise_call(GpuElemwise *ge, void **args, int flags);
 
 #define GE_BROADCAST   0x0100
```
nouiz: This commit comment should be in the doc of the GpuElemwise_new()

nouiz:
Don't forget to bump the version number.

src/gpuarray_array.c
```diff
@@ -19,37 +19,67 @@
...
+  a.otype = dst->typecode;
+
+  if (ctx->extcopy_cache != NULL)
+    k = cache_get(ctx->extcopy_cache, &a);
+  if (k == NULL) {
+    gpuelemwise_arg gargs[2];
+    gargs[0].name = "a";
+    gargs[0].typecode = src->typecode;
+    gargs[0].flags = GE_READ;
+    gargs[1].name = "b";
```
nouiz: I think you should set gargs[1].{typecode,flags}

nouiz: In fact, it is the 2 that should be 1 in the next 2 lines.

src/gpuarray_array.c
```diff
@@ -544,9 +574,12 @@ int GpuArray_setarray(GpuArray *a, const GpuArray *v) {
...
 
-  err = GA_UNSUPPORTED_ERROR;/*a->ops->buffer_extcopy(v->data, v->offset, a->data, a->offset,
-			       v->typecode, a->typecode, a->nd, a->dimensions,
-			       strs, a->nd, a->dimensions, a->strides);*/
+
+  memcpy(&tv, v, sizeof(GpuArray));
+  tv.nd = a->nd;
+  tv.dimensions = a->dimensions;
+  tv.strides = strs;
+  err = ga_extcopy(a, &tv);
```
nouiz: Can you confirm that this broadcast? Is this tested?

src/gpuarray_array.c
```diff
@@ -813,10 +846,7 @@ int GpuArray_move(GpuArray *dst, const GpuArray *src) {
   if (!GpuArray_ISONESEGMENT(dst) || !GpuArray_ISONESEGMENT(src) ||
       GpuArray_ISFORTRAN(dst) != GpuArray_ISFORTRAN(src) ||
       dst->typecode != src->typecode) {
-    return GA_UNSUPPORTED_ERROR; /*dst->ops->buffer_extcopy(src->data, src->offset, dst->data,
-                                    dst->offset, src->typecode, dst->typecode,
-                                    src->nd, src->dimensions, src->strides,
-                                    dst->nd, dst->dimensions, dst->strides);*/
+    return ga_extcopy(dst, src);
```
nouiz: is this tested?

src/gpuarray_array.c
```diff
@@ -978,16 +1007,16 @@ int GpuArray_concatenate(GpuArray *r, const GpuArray **as, size_t n,
...
   res_off = r->offset;
+  res_dims = r->dimensions;
   for (i = 0; i < n; i++) {
-    err = GA_UNSUPPORTED_ERROR; /*r->ops->buffer_extcopy(as[i]->data, as[i]->offset, r->data,
-                                 res_off, as[i]->typecode, r->typecode,
-                                 as[i]->nd, as[i]->dimensions,
-                                 as[i]->strides, r->nd, as[i]->dimensions,
-                                 r->strides);*/
+    r->dimensions = as[i]->dimensions;
+    err = GpuArray_move(r, as[i]);
```
nouiz: is this tested?

src/gpuarray_elemwise.c
abergeron: ok

src/gpuarray_elemwise.c
abergeron: Assert in C are for things that should always be true but if they are not are indicator of a more serious problem.

I don't want a pretty user error message in this case since the only reason that would be false is a bug.  Also, this way the check is only done in the debug build.

src/gpuarray_elemwise.c
abergeron: No, because its size is the number of arrays which doesn't change.

src/gpuarray_elemwise.c
abergeron: ok

src/gpuarray_array.c
abergeron: Now it does.

src/gpuarray_array.c
abergeron: I'm not sure.

src/gpuarray_array.c
abergeron: Not sure.

abergeron:
I'll do the rebase once the prerequistes are in to avoid redoing the work all the time.

src/gpuarray_elemwise.c
nouiz: right.

src/gpuarray_elemwise.c
nouiz: ok

src/gpuarray_array.c
abergeron: Yes

src/gpuarray_elemwise.c
abergeron: ok

Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+Release/Makefile: Release
+	(cd Release && cmake .. -DCMAKE_BUILD_TYPE=Release)
+
+relc: Release/Makefile
+	(cd Release && make && make install)
+
+testpy:
+	python setup.py build_ext --inplace
+
```
lamblin: In a future version, would it be possible to have a way of specifyint `-DCMAKE_INSTALL_PREFIX=~/.local` for instance?

src/gpuarray_array.c
```diff
@@ -1008,15 +1009,18 @@ int GpuArray_concatenate(GpuArray *r, const GpuArray **as, size_t n,
 
   res_off = r->offset;
   res_dims = r->dimensions;
+  res_flags = r->flags;
+  r->flags &= ~(GA_C_CONTIGUOUS|GA_F_CONTIGUOUS);
```
nouiz: why remove the contiguous flag?

src/gpuarray_array.c
```diff
@@ -1008,15 +1009,18 @@ int GpuArray_concatenate(GpuArray *r, const GpuArray **as, size_t n,
...
   for (i = 0; i < n; i++) {
     r->dimensions = as[i]->dimensions;
-    err = GpuArray_move(r, as[i]);
+    err = ga_extcopy(r, as[i]);
```
nouiz: why change this? Can you confirm me this is just a shortcut as GpuArray_Move would have called this?

src/gpuarray_array.c
```diff
@@ -574,11 +574,11 @@ int GpuArray_setarray(GpuArray *a, const GpuArray *v) {
     }
   }
 
-
   memcpy(&tv, v, sizeof(GpuArray));
   tv.nd = a->nd;
   tv.dimensions = a->dimensions;
   tv.strides = strs;
+  tv.flags &= ~(GA_C_CONTIGUOUS|GA_F_CONTIGUOUS);
```
nouiz: why?

src/gpuarray_array.c
abergeron: Other it will think that it is contiguous and produce wrong results.

src/gpuarray_array.c
abergeron: It is a shortcut and should be a bit faster since we avoid the checks in GpuArray_move().  I used move as an attempt to use memcopy whenever possible, but this broke a lot of cases so I rolled back to always extcopy.

We can attempt to re-introduce the memcopy later if it ever becomes a problem.

src/gpuarray_array.c
abergeron: Since we change the dimensions the array is no longer contiguous and the code would do the wrong thing if we left those flags.

lamblin:
This definitely solves the slowdown in the last convnet benchmark. \o/

src/gpuarray_array.c
nouiz: What if it is contiguous? Will the dimensions collapsing still use the contiguous kernel? Or it will use the 1d kernel?

abergeron: It will use 1d if not explicitly tagged as contiguous.

nouiz: is it useful to change that in collapse? At the end, if we end up with 1
dim, check if the stride is elemsize. If so, consider contiguous.

This would prevent to do this optimization at everyplace that call extcopy.

abergeron: This is a corner case that doesn't happen anywhere in practice because all operations set the flags correctly on the arrays.  Here we muck with the array properties directly and therefore we have to adjust the flags to match.  In the interest of code size and speed I err on the safe side and just always say that the array is not contiguous because the non-contiguous code is safe in the contiguous case and the contiguous case is rare in this instance.

I oppose adding code to handle what is technically a violation of the interface.

nouiz: It is much simpler to fix after the collapsing. As it is faster then then
old back-end, we don't need to speed it up more, but at least add a comment
here that we could check here if the strides are contiguous or not instead
of doing what we do here.

abergeron: 1) Once we decide that we need to do collapsing it's too late and the code is already on the path for non-contiguous.
2) Also the collapsing code has currently no way to indicate that what we work on is actually contiguous.
3) Finally the purpose of those flags is to avoid having to do checks like what you propose.

For these three reason I still oppose adding a dirty hack for an obscure corner case.

nouiz: ok, but add a todo comment here that this can be optimized if we detect and set the contiguous flags correctly.

nouiz: don't forget the comment.

abergeron: done

nouiz:
The last commit break the compilation of pygpu:

```
bastienf@leto18:~/repos/libgpuarray$   python setup.py install --user
running install
running bdist_egg
running egg_info
writing requirements to pygpu.egg-info/requires.txt
writing pygpu.egg-info/PKG-INFO
writing top-level names to pygpu.egg-info/top_level.txt
writing dependency_links to pygpu.egg-info/dependency_links.txt
writing pbr to pygpu.egg-info/pbr.json
reading manifest file 'pygpu.egg-info/SOURCES.txt'
writing manifest file 'pygpu.egg-info/SOURCES.txt'
installing library code to build/bdist.linux-x86_64/egg
running install_lib
running build_py
running build_ext
building 'pygpu.gpuarray' extension
gcc -pthread -fno-strict-aliasing -g -O2 -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -DGPUARRAY_SHARED -I/Tmp/lisa/os_v5/anaconda/lib/python2.7/site-packages/numpy/core/include -I/Tmp/lisa/os_v5/anaconda/include/python2.7 -c pygpu/gpuarray.c -o build/temp.linux-x86_64-2.7/pygpu/gpuarray.o
In file included from /Tmp/lisa/os_v5/anaconda/lib/python2.7/site-packages/numpy/core/include/numpy/ndarraytypes.h:1781:0,
                 from /Tmp/lisa/os_v5/anaconda/lib/python2.7/site-packages/numpy/core/include/numpy/ndarrayobject.h:18,
                 from /Tmp/lisa/os_v5/anaconda/lib/python2.7/site-packages/numpy/core/include/numpy/arrayobject.h:4,
                 from pygpu/gpuarray.c:265:
/Tmp/lisa/os_v5/anaconda/lib/python2.7/site-packages/numpy/core/include/numpy/npy_1_7_deprecated_api.h:15:2: warning: #warning "Using deprecated NumPy API, disable it by " "#defining NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION" [-Wcpp]
 #warning "Using deprecated NumPy API, disable it by " \
  ^
pygpu/gpuarray.c: In function ‘__pyx_pf_5pygpu_8gpuarray_10GpuContext_4__cinit__’:
pygpu/gpuarray.c:12233:17: warning: assignment discards ‘const’ qualifier from pointer target type
       __pyx_t_1 = gpucontext_error(NULL, __pyx_v_err);
                 ^
pygpu/gpuarray.c: In function ‘__pyx_pf_5pygpu_8gpuarray_9GpuKernel_4__cinit__’:
pygpu/gpuarray.c:24365:38: error: ‘GA_USE_PTX’ undeclared (first use in this function)
     __pyx_v_flags = (__pyx_v_flags | GA_USE_PTX);
                                      ^
pygpu/gpuarray.c:24365:38: note: each undeclared identifier is reported only once for each function it appears in
error: command 'gcc' failed with exit status 1
```

abergeron:
Fixed.

src/gpuarray/buffer.h
```diff
@@ -746,10 +746,6 @@ typedef enum _ga_usefl {
   /* If you add a new flag, don't forget to update both
      gpuarray_buffer_{cuda,opencl}.c with the implementation of your flag */
   /**
-   * The kernel is made of PTX code.
-   */
-  GA_USE_PTX =      0x1000,
```
nouiz: Do we want to keep that options in case you or someone else in the futur make new kernel in ptx? The probability is low, but if the code still work we could keep it. But don't loose time to make it work if it get broken by other change.

abergeron: If you want to do that you can use GA_USE_BINARY.  It will work with PTX.

src/gpuarray_array.c
```diff
@@ -578,6 +578,7 @@ int GpuArray_setarray(GpuArray *a, const GpuArray *v) {
   tv.nd = a->nd;
   tv.dimensions = a->dimensions;
   tv.strides = strs;
+  /* THis could be optiomized by setting the right flags */
```
nouiz: THis -> This

src/gpuarray/buffer.h
nouiz: Add a note in the docstring of GA_USE_BINARY about this.

nouiz:
The code is cleaner. This is good.
It is great that this cache more stuff now. I don't know the % of time in call_compiler vs the rest of the compilation, so I can't really evaluate if it will be a problem in the multi-gpu case or not. We can go with this for now and use 2 cache for multi-gpu if needed later.

I'm running tests now.

abergeron:
Rebased with the new commits from #175

src/gpuarray_array.c
```diff
@@ -558,9 +574,13 @@ int GpuArray_setarray(GpuArray *a, const GpuArray *v) {
...
   }
 
-  err = a->ops->buffer_extcopy(v->data, v->offset, a->data, a->offset,
-			       v->typecode, a->typecode, a->nd, a->dimensions,
-			       strs, a->nd, a->dimensions, a->strides);
+  memcpy(&tv, v, sizeof(GpuArray));
+  tv.nd = a->nd;
+  tv.dimensions = a->dimensions;
+  tv.strides = strs;
+  /* THis could be optiomized by setting the right flags */
```
nouiz: THis -> This

src/gpuarray/buffer.h
nouiz: don't forget that comment.

abergeron: Why?  The PTX option was only added because the extcopy kernel was written in PTX.  I don't want people to actually use it if they can avoid it.

pygpu/gpuarray.pyx
```diff
@@ -955,40 +935,35 @@ cdef class GpuContext:
...
-                raise get_exc(err), self.ops.ctx_error(NULL) + ": " + str(devno)
+                raise get_exc(err), gpucontext_error(NULL, err).decode('utf-8') + ": " + str(devno)
 
     def __enter__(self):
         if cuda_enter == NULL:
             raise RuntimeError("cuda_enter not available")
         if cuda_exit == NULL:
             raise RuntimeError("cuda_exit not available")
-        if self.ops != gpuarray_get_ops("cuda"):
+        if self.kind != b"cuda":
```
nouiz: Why use bytes for the kind? I saw many changes at many places related to that and it seem to make the code more ugly. Can we avoid that?

src/gpuarray/buffer.h
nouiz: In case people look for this, having the information as a comment would be useful. This is super fast to add.

Makefile
abergeron: Now there is with Makefile.conf.

src/gpuarray/buffer.h
abergeron: Fine, I've added it.  But I still feel like this is an interface violation of the documentation.

nouiz:
@abergeron update https://github.com/Theano/Theano/issues/4448 with the one that are fixed (faster or as fast as the old back-end)

PR #179, opened 2016-05-27
Expose GpuElemwise from python and remove ElemwiseKernel

abergeron:
Since we now have the C version for elemwise, we can get rid of the experimental python version.

This should also be appreciatively faster.


src/gpuarray_elemwise.c
```diff
@@ -561,57 +562,55 @@ GpuElemwise *GpuElemwise_new(gpucontext *ctx,
...
+  res->nd = 8;
+  res->n = n;
 
   res->expr = strdup(expr);
   if (res->expr == NULL)
-    goto fail_expr;
-  if (preamble == NULL) {
-    res->preamble = NULL;
-  } else {
+    goto fail;
```
nouiz: To be safe, on line 561 you should call calloc, as otherwise res isn't zeroed and if there is an error, the new free code can have problems.

pygpu/gpuarray.pyx
```diff
@@ -1942,9 +1942,9 @@ cdef class GpuKernel:
                     _types[i] = GA_BUFFER
                 else:
                     _types[i] = dtype_to_typecode(types[i])
-                self.callbuf[i] = malloc(gpuarray_get_elsize(_types[i]))
-                if self.callbuf[i] == NULL:
-                    raise MemoryError
+                    self.callbuf[i] = malloc(gpuarray_get_elsize(_types[i]))
```
nouiz: On line 1891, you don't free the those malloc. Just callbuf itself.

pygpu/tests/test_elemwise.py
```diff
@@ -101,87 +100,6 @@ def ielemwise2_ops_array(op, dtype1, dtype2, shape):
     assert numpy.allclose(out_c, numpy.asarray(out_g), atol=1e-6)
 
 
-def test_elemwise_layouts():
```
nouiz: Why remove those tests? I think this tests more cases then the C tests. So I would have a tendency to keep them.

pygpu/tests/test_elemwise.py
```diff
@@ -101,87 +100,6 @@ def ielemwise2_ops_array(op, dtype1, dtype2, shape):
...
-                for sliced in [1, 2]:
-                    for order in ['c', 'f']:
-                        yield elemwise_layouts, shape, offseted_outer, \
-                            offseted_inner, sliced, order
-                        yield elemwise_layouts_mixed, shape, offseted_outer, \
-                            offseted_inner, sliced, order
-
-
-def test_elemwise_0():
-    elemwise_layouts((0,), False, False, 1, 'c')
```
nouiz: I don't think this is in the c test too.

pygpu/tests/test_elemwise.py
```diff
@@ -328,117 +246,3 @@ def broadcast(shapea, shapeb):
     rg = ag + bg
 
     check_meta_content(rg, rc)
-
-
-def test_elemwise_collapse():
```
nouiz: good idea to remove.

pygpu/tests/test_elemwise.py
```diff
@@ -7,9 +7,9 @@
 from six import PY2
 
 from .support import (guard_devsup, rand, check_flags, check_meta, check_all,
-                      context, gen_gpuarray, dtypes_no_complex,
-                      check_meta_content)
+                      context, gen_gpuarray, check_meta_content)
 
+dtypes_test = ['float32', 'int32', 'uint64']
```
nouiz: I woud also test int8. This is corner case that already caused problems.

src/gpuarray_elemwise.c
abergeron: Yes

pygpu/tests/test_elemwise.py
abergeron: I would rather extend the C tests than keep those.

pygpu/tests/test_elemwise.py
abergeron: Ok.

pygpu/gpuarray.pyx
abergeron: I know.  It was like that before, but I'll fix it.

pygpu/tests/test_elemwise.py
nouiz: I'm fine with that.

pygpu/tests/test_elemwise.py
abergeron: No it's not.  I'm not sure I want to add it, because I'm not sure I want to support it.  Currently it doesn't work.

pygpu/tests/test_elemwise.py
abergeron: Now that I look at it in detail, it's almost all tested, but with smaller arrays.  I'll add a test for an offset, but it's all that's missing.  It might look like more things are tested but they are irrelevant and where only there because I figure I might as well use all the options of gen_gpuarray().

Now that I've seen that it makes for extra-long tests, I tend to be more minimalist.

pygpu/tests/test_elemwise.py
nouiz: If we don't support it there, then we will need to support it at every places that call elemwise. I prefer to support it at only 1 place...

tests/check_elemwise.c
```diff
@@ -131,6 +131,72 @@ START_TEST(test_basic_simple)
...
+
+  ge = GpuElemwise_new(ctx, "", "c = a + b", 3, args, 2, 0);
+
+  ck_assert_ptr_ne(ge, NULL);
+
+  rargs[0] = &a;
+  rargs[1] = &b;
+  rargs[2] = &c;
+
+  ga_assert_ok(GpuElemwise_call(ge, rargs, GE_NOCOLLAPSE));
```
nouiz: call it with collapsing too.

Otherwise the test seem good.

abergeron: The collapsing code is tested elsewhere and doesn't interact with the offset.  I prefer to keep things separate if possible.

pygpu/tests/test_elemwise.py
abergeron: Yes, I suppose it's better to have that in one place.

abergeron: I added support.

nouiz:
You need to bump the feature version number and check that number in the corresponding Theano PR.

nouiz:
Don't forget to make a comment when you update a PR. This PR need a rebase. After that, we can merge.

abergeron:
Rebase is done, all should be well.

PR #182, opened 2016-05-31
float16 support for  GpuElemwise

abergeron:
This enable support for float16 in elemwise operations by performing the computation in float32.

This behaviour will only be active if the GE_CONVERT_F16 flags is passed when creating the GpuElemwise object.

It does not make GPUs which don't have even the most basic support for float16 work since those don't have either the data types or the conversion primitives that we need.  On the cuda side it should mostly always work, but on OpenCL your device needs to support the cl_khr_fp16 extension (which notably NVIDIA and most CPU vendors chose not to enable).

### DIFF
--- a/tests/check_elemwise.c
+++ b/tests/check_elemwise.c
@@ -12,7 +12,7 @@ void setup(void);
 void teardown(void);
 
 #define ga_assert_ok(e) ck_assert_int_eq(e, GA_NO_ERROR)
-/* float 16 table */
+/* float 16 table (0 through 10) */
 static const uint16_t F16[10] = {0x0000, 0x3c00, 0x4000, 0x4200, 0x4400, 0x4500, 0x4600, 0x4700, 0x4800, 0x4880};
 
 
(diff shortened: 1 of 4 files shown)


tests/check_elemwise.c
```diff
@@ -12,6 +12,9 @@ void setup(void);
 void teardown(void);
 
 #define ga_assert_ok(e) ck_assert_int_eq(e, GA_NO_ERROR)
+/* float 16 table */
+static const uint16_t F16[10] = {0x0000, 0x3c00, 0x4000, 0x4200, 0x4400, 0x4500, 0x4600, 0x4700, 0x4800, 0x4880};
```
nouiz: Can you add in the comment the values that correspond to those hex value?

nouiz:
ok to merge after the small comment update.

nouiz:
Do this need an update in Theano to use this?

tests/check_elemwise.c
abergeron: done.

abergeron:
Yes we need some changes in Theano to use it, but it will not break anything that currently works.

PR #185, opened 2016-06-04
Initial support for MaxAndArgmax and GpuArray_maxandargmax().

obilaniu:
What remains incomplete is a scheduler for the kernel, as noted in the
source code.

### DIFF
--- a/src/gpuarray_reduction.c
+++ b/src/gpuarray_reduction.c
@@ -26,6 +26,8 @@ struct gen_kernel_ctx{
 	unsigned        numIdx;
 	unsigned        reduxLen;
 	const unsigned* reduxList;
+	unsigned        hwAxisLen;
+	const unsigned* hwAxisList;
 	unsigned        numFreeIdx;
 	unsigned*       axisList;
 	const char*     dstMaxType;
@@ -36,14 +38,25 @@ typedef struct gen_kernel_ctx gen_kernel_ctx;
 
 
 /* Function prototypes */
+static int   axisInSet              (unsigned         v,
+                                     const unsigned*  set,
+                                     size_t           setLen,
+                                     size_t*          where);
 static int   checkargsMaxAndArgmax  (GpuArray*        dstMax,
                                      GpuArray*        dstArgmax,
                                      const GpuArray*  src,
                                      unsigned         reduxLen,
                                      const unsigned*  reduxList);
-static char* genkernelMaxAndArgmax  (unsigned         numIdx,
+static void  selectHwAxes           (unsigned*        hwAxesLen,
+                                     unsigned*        hwAxisList,
+                                     const GpuArray*  src,
+                                     unsigned         reduxLen,
+                                     const unsigned*  reduxList);
+static char* gensourceMaxAndArgmax  (unsigned         numIdx,
                                      unsigned         reduxLen,
                                      const unsigned*  reduxList,
+                                     const unsigned   hwAxisLen,
+                                     const unsigned*  hwAxisList,
                                      const char*      dstMaxType,
                                      const char*      dstArgmaxType);
 static void  appendKernel           (strb*            s,
@@ -79,6 +92,9 @@ static void  computeAxisList        (unsigned*        axisList,
                                      unsigned         numAxis,
                                      const unsigned*  reduxList,
                                      unsigned         reduxLen);
+static int   compileMaxAndArgmax    (GpuKernel*       kernel,
+                                     const char*      src,
+                                     gpucontext*      ctx);
 static void  scheduleMaxAndArgmax   (const GpuKernel* kernel,
                                      const GpuArray*  src,
                                      unsigned         reduxLen,
@@ -91,7 +107,9 @@ static int   invokeMaxAndArgmax     (GpuKernel*       kernel,
                                      GpuArray*        dstArgmax,
                                      const GpuArray*  src,
                                      unsigned         reduxLen,
-                                     const unsigned*  reduxList);
+                                     const unsigned*  reduxList,
+                                     unsigned         hwAxisLen,
+                                     const unsigned*  hwAxisList);
 
 
 /* Function implementation */
@@ -100,69 +118,74 @@ GPUARRAY_PUBLIC int GpuArray_maxandargmax(GpuArray*       dstMax,
                                           const GpuArray* src,
                                           unsigned        reduxLen,
                                           const unsigned* reduxList){
-	/**
-	 * Sanity check on arguments
-	 */
+	int          ret           = 0;
+	unsigned     hwAxisLen     = 0;
+	unsigned     hwAxisList[3] = {0,0,0};
+	const char*  dstMaxType    = NULL;
+	const char*  dstArgmaxType = NULL;
+	char*        s             = NULL;
+	GpuKernel    kernel;
 	
+	/* Sanity-check arguments */
 	if(!checkargsMaxAndArgmax(dstMax, dstArgmax, src, reduxLen, reduxList)){
 		return GA_INVALID_ERROR;
 	}
 	
+	/* Select hardware axis mapping */
+	selectHwAxes(&hwAxisLen, hwAxisList, src, reduxLen, reduxList);
 	
-	/**
-	 * Generate kernel source code.
-	 */
-	
-	int          ret;
-	const char*  dstMaxType      = gpuarray_get_type(src->typecode) -> cluda_name;
-	const char*  dstArgmaxType   = gpuarray_get_type(GA_SIZE)       -> cluda_name;
-	char*        s               = genkernelMaxAndArgmax(src->nd,
-	                                                     reduxLen,
-	                                                     reduxList,
-	                                                     dstMaxType,
-	                                                     dstArgmaxType);
+	/* Generate kernel source code */
+	dstMaxType    = gpuarray_get_type(src->typecode)->cluda_name;
+	dstArgmaxType = gpuarray_get_type(GA_SSIZE)     ->cluda_name;
+	s = gensourceMaxAndArgmax(src->nd,
+	                          reduxLen,   reduxList,
+	                          hwAxisLen,  hwAxisList,
+	                          dstMaxType, dstArgmaxType);
 	if(!s){return GA_MEMORY_ERROR;}
 	
-	
-	/**
-	 * Compile it.
-	 */
-	
-	const int    ARG_TYPECODE[11] = {
-		GA_BUFFER, /* src */
-		GA_SIZE,   /* srcOff */
-		GA_BUFFER, /* srcSteps */
-		GA_BUFFER, /* srcSize */
-		GA_BUFFER, /* chnkSize */
-		GA_BUFFER, /* dstMax */
-		GA_SIZE,   /* dstMaxOff */
-		GA_BUFFER, /* dstMaxSteps */
-		GA_BUFFER, /* dstArgmax */
-		GA_SIZE,   /* dstArgmaxOff */
-		GA_BUFFER  /* dstArgmaxSteps */
-	};
-	
-	GpuKernel kernel;
-	const size_t l = strlen(s);
-	ret = GpuKernel_init(&kernel, gpudata_context(src->data),
-	                     1, (const char**)&s, &l, "maxandargmax",
-	                     11, ARG_TYPECODE, GA_USE_CLUDA, (char**)0);
+	/* Compile kernel source code */
+	ret = compileMaxAndArgmax(&kernel, s, gpudata_context(src->data));
 	free(s);
 	if(ret != GA_NO_ERROR){
 		return ret;
 	}
 	
-	
-	/**
-	 * Invoke it.
-	 */
-	
-	return invokeMaxAndArgmax(&kernel, dstMax, dstArgmax, src, reduxLen, reduxList);
+	/* Invoke compiled kernel */
+	return invokeMaxAndArgmax(&kernel, dstMax, dstArgmax, src,
+	                          reduxLen, reduxList, hwAxisLen, hwAxisList);
 }
 
 /**
+ * @brief Check whether axis numbered v is already in the given set of axes.
+ * 
+ * @param [in]  v
+ * @param [in]  set
+ * @param [in]  setLen
+ * @param [out] where
+ * @return Non-zero if the set is non-empty and v is in it; Zero otherwise.
+ */
+
+static int   axisInSet              (unsigned          v,
+                                     const unsigned*   set,
+                                     size_t            setLen,
+                                     size_t*           where){
+	size_t i;
+	
+	for(i=0;i<setLen;i++){
+		if(set[i] == v){
+			if(where){*where = i;}
+			return 1;
+		}
+	}
+	
+	return 0;
+}
+
+
+/**
  * @brief Check the sanity of the arguments, in agreement with the
- *        documentation for GpuArray_maxandargmax().
+ *        documentation for GpuArray_maxandargmax(). The list of arguments is
+ *        precisely in agreement with GpuArray_maxandargmax().
  * 
  * @param [in] dstMax
  * @param [in] dstArgmax
@@ -177,7 +200,7 @@ static int   checkargsMaxAndArgmax  (GpuArray*         dstMax,
                                      const GpuArray*   src,
                                      unsigned          reduxLen,
                                      const unsigned*   reduxList){
-	int i, j;
+	int i;
 	
 	/* Insane src or reduxLen? */
 	if(!dstMax || !dstArgmax || !src || src->nd == 0 || reduxLen == 0 ||
@@ -186,20 +209,51 @@ static int   checkargsMaxAndArgmax  (GpuArray*         dstMax,
 	}
 	
 	for(i=0;i<(int)reduxLen;i++){
-		/* Insane list entry? */
-		if(reduxList[i] >= src->nd){
+		/* Insane or duplicate list entry? */
+		if(reduxList[i] >= src->nd                ||
+		   axisInSet(reduxList[i], reduxList, i, 0)){
 			return 0;
-		}
-		
-		for(j=i-1;j>=0;j--){
-			/* Duplicate list entry? */
-			if(reduxList[i] == reduxList[j]){
-				return 0;
-			}
 		}
 	}
 	
 	return 1;
+}
+
+/**
+ * @brief Select which axes (up to 3) will be assigned to hardware
+ *        dimensions.
+ * 
+ * @param [out] hwAxisLen
+ * @param [out] hwAxisList
+ * @param [in]  src
+ * @param [in]  reduxLen
+ * @param [in]  reduxList
+ */
+
+static void  selectHwAxes           (unsigned*        hwAxisLen,
+                                     unsigned*        hwAxisList,
+                                     const GpuArray*  src,
+                                     unsigned         reduxLen,
+                                     const unsigned*  reduxList){
+	unsigned i, j;
+	
+	*hwAxisLen = src->nd-reduxLen < 3 ? src->nd-reduxLen : 3;
+	
+	for(i=0;i<*hwAxisLen;i++){
+		size_t   maxV = 0;
+		unsigned maxI = 0;
+		
+		for(j=0;j<src->nd;j++){
+			if(!axisInSet(j, hwAxisList, i,        0) &&
+			   !axisInSet(j, reduxList,  reduxLen, 0) &&
+			   src->dimensions[j] > maxV){
+				maxV = src->dimensions[j];
+				maxI = j;
+			}
+		}
+		
+		hwAxisList[i] = maxI;
+	}
 }
 
 /**
@@ -208,15 +262,19 @@ static int   checkargsMaxAndArgmax  (GpuArray*         dstMax,
  * @param [in]  numIdx
  * @param [in]  reduxLen
  * @param [in]  reduxList
+ * @param [in]  hwAxisLen
+ * @param [in]  hwAxisList
  * @param [in]  dstMaxType
  * @param [in]  dstArgmaxType
  * @return A free()'able string containing source code implementing the
  *         kernel, or else NULL.
  */
 
-static char* genkernelMaxAndArgmax (unsigned          numIdx,
+static char* gensourceMaxAndArgmax (unsigned          numIdx,
                                     const unsigned    reduxLen,
                                     const unsigned*   reduxList,
+                                    const unsigned    hwAxisLen,
+                                    const unsigned*   hwAxisList,
                                     const char*       dstMaxType,
                                     const char*       dstArgmaxType){
 	/* Save the parameters of the reduction in a generator context. */
@@ -224,6 +282,8 @@ static char* genkernelMaxAndArgmax (unsigned          numIdx,
 	ctx.numIdx        = numIdx;
 	ctx.reduxLen      = reduxLen;
 	ctx.reduxList     = reduxList;
+	ctx.hwAxisLen     = hwAxisLen;
+	ctx.hwAxisList    = hwAxisList;
 	ctx.numFreeIdx    = ctx.numIdx - ctx.reduxLen;
 	ctx.axisList      = malloc(numIdx*sizeof(unsigned));
 	ctx.dstMaxType    = dstMaxType;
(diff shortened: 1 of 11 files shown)


src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+#ifdef _MSC_VER
+#define _CRT_SECURE_NO_WARNINGS
+#endif
+
+#include <assert.h>
+#include <stdarg.h>
+#include <stddef.h>
+#if !defined(_MSC_VER) || _MSC_VER < 1600
+#include <stdint.h>
+#endif
```
abergeron: `#include gpuarray/config.h` instead.  It does that dance for stdint and has a fallback for older MSVC.

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+/* Datatypes */
+struct GEN_KERNEL_CTX{
+	int         numIdx;
+	const int*  isReduced;
+	int         numRedIdx;
+	int         numFreeIdx;
+	const char* dstMaxType;
+	const char* dstArgmaxType;
+};
+typedef struct GEN_KERNEL_CTX GEN_KERNEL_CTX;
```
abergeron: Please have the struct name be different (like with an underscore before the name).  Also, names in all caps suggest a macro which this isn't.

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+	/* Compile it */
+	const int    ARG_TYPECODE[8] = {
+		GA_POINTER, /* src */
+		GA_POINTER, /* srcSteps */
+		GA_POINTER, /* srcSize */
+		GA_POINTER, /* numBlk */
+		GA_POINTER, /* dstMax */
+		GA_POINTER, /* dstMaxSteps */
+		GA_POINTER, /* dstArgmax */
+		GA_POINTER  /* dstArgmaxSteps */
```
abergeron: The proper type is `GA_BUFFER`, not `GA_POINTER` for those.

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+	ctx.numRedIdx     = getRdxIdx(ctx.numIdx, ctx.isReduced);
+	ctx.numFreeIdx    = ctx.numIdx - ctx.numRedIdx;
+	ctx.dstMaxType    = dstMaxType;
+	ctx.dstArgmaxType = dstArgmaxType;
+	
+	/**
+	 * Allocate string buffer.
+	 */
+	
+	strb* s = strb_alloc(5*1024);
```
abergeron: Why don't you make the strb a local struct?  You can still reserve some space with strb_ensure().

obilaniu: @abergeron How does one in-place init a `strb`? I saw `strb_alloc()` but where's a `strb_init(strb* s)`?

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+GPUARRAY_PUBLIC int GpuArray_maxandargmax(GpuArray*       dstMax,
+                                          GpuArray*       dstArgmax,
+                                          const GpuArray* src,
+                                          const int*      isReduced){
+	/**
+	 * Generate kernel source code
+	 */
+	
+	const char*  dstMaxType      = gpuarray_get_type(src->typecode) -> cluda_name;
+	const char*  dstArgmaxType   = gpuarray_get_type(GA_LONG)       -> cluda_name;
```
abergeron: The proper type here would be ga_size (and not a signed one).

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+	strb_appendf(s, "typedef %s     T;/* The type of the array being processed. */\n", ctx->dstMaxType);
+	strb_appendf(s, "typedef %s     X;/* Index type: signed 32/64-bit. */\n",          ctx->dstArgmaxType);
+	strb_appends(s, "\n");
+	strb_appends(s, "\n");
+	strb_appends(s, "\n");
+}
+static void  appendPrototype        (strb*           s,
+                                     GEN_KERNEL_CTX* ctx){
+	strb_appends(s, "__global__ void maxandargmax(const T*        src,\n");
+	strb_appends(s, "                             const X*        srcSteps,\n");
```
abergeron: The input types should always be 64 bits to simplify the calling protocol.  Cast the values to 32 bits inside if you need them.

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+	strb_appends(s, "/* Typedefs */\n");
+	strb_appendf(s, "typedef %s     T;/* The type of the array being processed. */\n", ctx->dstMaxType);
+	strb_appendf(s, "typedef %s     X;/* Index type: signed 32/64-bit. */\n",          ctx->dstArgmaxType);
+	strb_appends(s, "\n");
+	strb_appends(s, "\n");
+	strb_appends(s, "\n");
+}
+static void  appendPrototype        (strb*           s,
+                                     GEN_KERNEL_CTX* ctx){
+	strb_appends(s, "__global__ void maxandargmax(const T*        src,\n");
```
abergeron: Instead of __global__ you should use KERNEL.  This will also make the name reachable since it won't be C++ mangled.

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+	strb_appends(s, "                             X*              dstArgmax,\n");
+	strb_appends(s, "                             const X*        dstArgmaxSteps)");
+}
+static void  appendIndexDeclarations(strb*           s,
+                                     GEN_KERNEL_CTX* ctx){
+	strb_appends(s, "\t/* GPU kernel coordinates. Always 3D. */\n");
+	
+	strb_appends(s, "\tX bi0 = blockIdx.x,  bi1 = blockIdx.y,  bi2 = blockIdx.z;\n");
+	strb_appends(s, "\tX bd0 = blockDim.x,  bd1 = blockDim.y,  bd2 = blockDim.z;\n");
+	strb_appends(s, "\tX ti0 = threadIdx.x, ti1 = threadIdx.y, ti2 = threadIdx.z;\n");
```
abergeron: Make this not CUDA-specific by using the cluda aliases.

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+}
+static void  appendLoopMacroDefs    (strb*           s,
+                                     GEN_KERNEL_CTX* ctx){
+	int i;
+	
+	/**
+	 * FOROVER Macro
+	 */
+	
+	strb_appends(s, "#define FOROVER(idx)    for(i##idx = i##idx##Start; i##idx < i##idx##End; i##idx++)\n");
```
abergeron: I'm not convinced that using macros here make the code any more readable.  In fact I think it obscures things.

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+static void  scheduleMaxAndArgmax   (size_t*         blockSize,
+                                     size_t*         gridSize,
+                                     const GpuArray* src,
+                                     const int*      isReduced){
+	//int maxThreadPerBlock = 1024;
+	//int numFreeIdx = src->nd - getRdxIdx(src->nd, isReduced);
+	
+	/* Naive solution. Optimization is a tough problem. */
+	blockSize[0] = blockSize[1] = blockSize[2] = 1;
+	gridSize [0] = gridSize [1] = gridSize [2] = 1;
```
abergeron: This is not ok for final code.  You run with a single thread in a single block.  This will be extremely slow no matter what the algorithm.

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+	size_t blockSize[3];
+	size_t gridSize[3];
+	
+	scheduleMaxAndArgmax(blockSize, gridSize, src, isReduced);
+	GpuKernel_call(kernel,
+	               getRdxIdx(src->nd, isReduced),
+	               blockSize,
+	               gridSize,
+	               0,
+	               NULL);
```
abergeron: You don't pass any arguments to the kernel anywhere.  I don't see how this can work.

src/gpuarray_reduction.c
abergeron: `strb s = STRB_STATIC_INIT;`  This initializes an empty string.  Then you can call strb_ensure() on it.

There are plenty of examples of that in the rest of the code.

src/gpuarray_reduction.c
obilaniu: @abergeron Defining the macro, once, the way I've done it allows me to greatly simplify the `strb_appendf()` that would otherwise result. Generating

    FOROVER(0){ESCAPE(0)
    FOROVER(1){ESCAPE(1)
    FOROVER(2){ESCAPE(2)
    FOROVER(3){ESCAPE(3)

is far easier, using `strb_appendf()`, than
    for(i0 = i0Start; i0<i0End; i0++)
    for(i1 = i1Start; i1<i1End; i1++)
    for(i2 = i2Start; i2<i2End; i2++)
    for(i3 = i3Start; i3<i3End; i3++)

abergeron: These are appended by a for loop.  So there's not much difference anyway.

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,548 @@
...
+ * FIXME: Implement a working scheduler and invoker.
+ * 
+ * To schedule effectively the work across several dimensions of possibly
+ * ugly numbers, taking into account the limitations on thread & block
+ * scheduling, will require some library capable of providing a "factoring"
+ * of the tensor dimensions into "nice" prime numbers. Their product may
+ * be allowed to be larger than the original number, provided it remains
+ * within bounds.
+ */
+
```
obilaniu: @abergeron w.r.t. your comment below, the invoker does not run because I've not yet managed to define an algorithm capable of factorizing the tensor dimensions that is 1) robust to large prime factors, 2) Increases the number to factor minimally, and only if required, 3) Selects a subset of factors that get as close as possible to the max number of threads permitted at a time by the runtime (e.g. 1024) while 4) Not violating the limits in every dimension (e.g. 512,512,64 for 3D). No approach I can think of couldn't be perverted by a pathologically-sized tensor.

src/gpuarray_reduction.c
obilaniu: @abergeron `X`, which is typedef'ed to `ctx->dstArgmaxType`, is currently always 64 bits due to the later's initialization as `gpuarray_get_type(GA_LONG)       -> cluda_name` (Will get changed to `GA_SIZE`). If in the future this will change, we can set that to 32 bits easily.

src/gpuarray_reduction.c
obilaniu: @abergeron I thought the strides could be negative, and hence made them signed. Maybe `ssize_t` would be appropriate?

I made everything index-related of type `X` and let it be signed because it's hell to fight C's integer promotion rules: Unsigned outranks signed. 64-bit signed values should store any (unsigned) dimension we could be reasonably interested in for the next few centuries.

src/gpuarray_reduction.c
abergeron: What I mean is that the inputs should always be 64-bits no matter what.  But the index computations are what should be casted to 32-bits if possible.

src/gpuarray_reduction.c
abergeron: A best-effort approch is OK.  It doesn't have to be perfect on the first try, we can always tune the scheduling later.

Also, I've seen that keeping the number of threads in the 32-256 range is best for CUDA devices.  The number of blocks should ideally be around 16 times the number of MP.  Of course for smaller arrays all these sizes have to be tuned down.

src/gpuarray_reduction.c
obilaniu: @abergeron A prefix of `_` followed by a capital letter, or a prefix of `__`, are reserved by ISO C (and other offshoots of C like C++, OpenCL and CUDA) for the C implementation.

I propose to use `gen_kernel_ctx`.

src/gpuarray_reduction.c
obilaniu: @abergeron A bit off-topic, but if `STRB_STATIC_INIT` were defined as `(strb){NULL, 0, 0}` instead of `{NULL, 0, 0}`, then it could be used both at initialization time (`strb foo = STRB_STATIC_INIT;`) and later on (`foo = STRB_STATIC_INIT;`).

src/gpuarray_reduction.c
obilaniu: @abergeron What are these aliases and what functions do I call to obtain them?

src/gpuarray_reduction.c
abergeron: I prefer the lowercase version, so yeah.

src/gpuarray_reduction.c
abergeron: Strides can be negative, yes.  Sizes can't.

You are right about the C promotion rules, but ultimately `(unsigned)10 + (signed)-2 = (unsigned)8` so we are fine as long as we don't go under 0 for the total.  On that note it made me notice that you forgot to handle offsets.

src/util/integerfactoring.h
```diff
@@ -0,0 +1,241 @@
+/* Include Guards */
+#ifndef INTEGERFACTORING_H
+#define INTEGERFACTORING_H
+
+
+/* Includes */
+#include <stdio.h>
+#include <stdint.h>
+#include <stdint-gcc.h>
```
abergeron: We can't rely on gcc-exclusive headers.

src/util/integerfactoring.c
```diff
@@ -0,0 +1,610 @@
...
+int main(int argc,char* argv[]){
+	runTest(18446744073709551615ULL,                            0,  4096);
+	runTest(18446744073709551615ULL,                            0,     0);
+	runTest(18446744073709551615ULL,                           -1,     0);
+	runTest( 2196095973992233039ULL,                            0,     0);
+	runTest( 2196095973992233039ULL,  2196095973992233039ULL*1.01,    64);
+	
+	return 0;
+}
+#endif
```
abergeron: Please don't leave `#if 0` code in the file.  Also since this is test code you might want to turn it into a real test (like check_util_integerfactoring).

src/util/integerfactoring.h
```diff
@@ -0,0 +1,241 @@
...
+/* C++ Extern "C" Guard */
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+
+/* Data Structure Prototypes & Typedefs */
+struct GA_FACTOR_LIST;
+typedef struct GA_FACTOR_LIST GA_FACTOR_LIST;
```
abergeron: Make the struct name lowercase (it's not a macro) and don't have the typedef and the struct share the name.

src/util/integerfactoring.h
```diff
@@ -0,0 +1,241 @@
...
+/* Functions */
+
+/**
+ * @brief Count trailing zeros of a 64-bit integer.
+ * 
+ * @param [in] n  The integer whose trailing zero count is to be computed.
+ * @return     If n != 0, returns trailing zero count; Else returns 64.
+ */
+
+int      gaICtz(uint64_t n);
```
abergeron: Are these functions useful outside of integerfactoring.c?  If not, make them static functions in the file and remove them from the header.

src/gpuarray_reduction.c
```diff
@@ -481,46 +504,128 @@ static void  appendLoopMacroUndefs  (strb*           s,
...
+	size_t warpSize,
+	       maxL, maxL0, maxL1, maxL2,  /* Maximum total and per-dimension thread/block sizes */
+	       maxG, maxG0, maxG1, maxG2;  /* Maximum total and per-dimension block /grid  sizes */
+	gpukernel_property(kernel->k, GA_KERNEL_PROP_PREFLSIZE, &warpSize);
+	gpukernel_property(kernel->k, GA_KERNEL_PROP_MAXLSIZE,  &maxL);
+	gpudata_property  (src->data, GA_CTX_PROP_MAXLSIZE0,    &maxL0);
+	gpudata_property  (src->data, GA_CTX_PROP_MAXLSIZE1,    &maxL1);
+	gpudata_property  (src->data, GA_CTX_PROP_MAXLSIZE2,    &maxL2);
+	gpudata_property  (src->data, GA_CTX_PROP_MAXGSIZE,     &maxG);
+	gpudata_property  (src->data, GA_CTX_PROP_MAXGSIZE0,    &maxG0);
```
abergeron: Just so you know, MAXLSIZE and MAXLSIZE0 are the same value.

src/util/integerfactoring.h
obilaniu: They _might_ be, to whom I don't know. Those functions are intended to be libgpuarray-public but user-private, in much the same way that the other `utils/` are (or that I imagine them to be).

abergeron: The functions to count trailing/leading zeros and the modular arithmetic are pretty much only for factoring so they should probably be statics.  There may be others, but I'm not 100% sure.

The interface to the rest of the library (i.e. functions that will be called from other files) should stay in the header.

src/util/integerfactoring.c
obilaniu: @abergeron Alright here's my problem w.r.t. tests.

It appears that the tests are implemented by compiling the library, then linking the check executables against it. Unfortunately, I intentionally don't declare the integer factoring bits with `GPUARRAY_PUBLIC`, since they're _not_ meant to be public. I assume the black magic in `config.h` and elsewhere causes only symbols marked `GPUARRAY_PUBLIC` to be "external" as reported by `nm`. Indeed, my `gaI*()` functions are all reported as `t` instead of `T`, meaning that they're local.

So I can't compile `check_util_integerfactoring` - unless I build against `gpuarray-static`, which is my current workaround. But how do I write a conditional in CMake that only builds my check if the static build is enabled?

abergeron: Yes, that is a problem.  I haven't figured out a way to go around this, but maybe forcing all the tests to build against the static version would do.

If you figure it out that would be great, but don't spend too much time on this.

src/util/integerfactoring.h
obilaniu: @abergeron All four critiques above have been fixed. Running and passing testcase code is in new commit. The code in `gpuarray_reduction.c` won't build anymore (datatype name change) but changes on selfsame file are also inbound.

src/util/integerfactoring.h
obilaniu: @abergeron Incidentally, why must the typedef and struct not share the same name? Doing this is valid in C; After `typedef struct X X;`, using `X` is shorthand for `struct X`, but if you choose to do so you can still use `struct X` as a visual cue of its `struct`-ness.

abergeron: It's allowed in C but breaks on C++ compilers (and the headers sometimes go though one).  Also this is the style that the rest of the code uses.

obilaniu: @abergeron On _which_ C++ compilers? [This Godbolt link](https://godbolt.org/g/6Fsgra) shows the pattern compiling just fine with the oldest and strictest C++ diagnostics imaginable, on ICC, GCC and Clang. [This post](https://blogs.msdn.microsoft.com/oldnewthing/20080326-00/?p=22993) by a Microsoft employee claims that this used to be required in K&R C, but C89 permits it and C++98 has always permitted it [(Example within Standard at §7.1.3(3) )](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2005/n1905.pdf). We're not targeting K&R C and even MSVC supports C89 and C++.

abergeron: This appears to be the case, yes.  So I guess it's fine to share the name, then.

src/util/integerfactoring.h
abergeron: ok

abergeron:
Ok so now it seems complete.  Could you add some tests (possibly inspired by the ones for elemwise).  It doesn't need to be comprehensive and cover every case, but should at least try to cover some bases and make sure nothing is broken.

obilaniu:
@abergeron Will start at once.

src/gpuarray_reduction.c
```diff
@@ -0,0 +1,804 @@
...
+	 * Call kernel, release arguments and return error code
+	 */
+	
+	ret = GpuKernel_call(k, 3, blockSize, gridSize, 0, args);
+	releaseGpudata:
+	gpudata_release(srcStepsGD);
+	gpudata_release(srcSizeGD);
+	gpudata_release(chunkSizeGD);
+	gpudata_release(dstMaxStepsGD);
+	gpudata_release(dstArgmaxStepsGD);
```
obilaniu: @abergeron To note what has been said in email, this is apparently not safe, since `gpudata_release()` chokes on `NULL`. IMHO it would be worthwhile to add a quick `if(ptr){...}` to `gpudata_release()`, to imitate `free(NULL)`'s safe semantics and to simplify code that uses several `gpudata_alloc()` calls.

obilaniu:
@abergeron Don't merge yet! This doesn't yet have a solution to the test segfaults!

abergeron:
Yes, I know.

obilaniu:
@abergeron Incidentally, can I make a small recommendation?

When debugging tests, I find it incredibly annoying that the test forks itself out and, after a timeout, kills its copy (whether or not it's halted by a debugger). The first problem can be solved with `set follow-fork-mode child`, but the second one seems to have no solution other than to

    #ifndef NDEBUG
    tcase_set_timeout(tc, 8.0);
    #endif

What do you think of applying this to the other location where a timeout is set (`check_array.c:252`)?

abergeron:
You can also export `CK_FORK=no` to disable the forking completely.  This is the ideal to debug.

src/gpuarray_reduction.c
```diff
@@ -726,10 +810,11 @@ static int   invokeMaxAndArgmax     (GpuKernel*       k,
 	 * 
 	 * This implies choosing the block, grid and chunk size appropriately.
 	 */
-	
+	/*
 	scheduleMaxAndArgmax(k, src, reduxLen, reduxList,
 	                     blockSize, gridSize, chunkSize);
-	
+	*/
+	(void)scheduleMaxAndArgmax;
```
abergeron: Why this?

obilaniu: @abergeron As I wrote in my email, I was doing static scheduling to make sure the rest works. You can also see a hardcoded `50` above. This will disappear soon, I'm working on it.

abergeron: Ok

abergeron:
jenkins test this please

obilaniu:
@abergeron Going by the logs, it has neither compiled nor executed the two new check executables I've added.

tests/CMakeLists.txt
```diff
@@ -48,6 +48,14 @@ add_executable(check_util main.c check_util.c)
 target_link_libraries(check_util ${CHECK_LIBRARIES} gpuarray)
 add_test(test_util "${CMAKE_CURRENT_BINARY_DIR}/check_util")
 
+add_executable(check_util_integerfactoring main.c check_util_integerfactoring.c)
+target_link_libraries(check_util_integerfactoring ${CHECK_LIBRARIES} gpuarray-static)
+add_test(test_util_integerfactoring ${CMAKE_CURRENT_BINARY_DIR}/check_util_integerfactoring)
+
```
abergeron: Add quotes around the second parameter like for the other tests.

tests/CMakeLists.txt
```diff
@@ -48,6 +48,14 @@ add_executable(check_util main.c check_util.c)
...
+add_executable(check_reduction main.c device.c check_reduction.c)
+target_link_libraries(check_reduction ${CHECK_LIBRARIES} gpuarray)
+add_test(test_reduction ${CMAKE_CURRENT_BINARY_DIR}/check_reduction)
```
abergeron: Here too.

obilaniu: Done.

abergeron:
Please remove the merge commit and do a rebase if necessary.

Other than that the code looks ok.  I'll try to look into why jenkins doesn't run the additional tests.

abergeron:
jenkins test this please

obilaniu:
@abergeron I *think* I rebased? Within the bounds of my understanding of Git.

abergeron:
Ok that good.  And I figured out why the new tests didn't ran, it was because the test script was setup to always test master.  I fixed that in another PR so we can run the tests on this.

jenkins test this please

abergeron:
There are a couple of warnings about declarations after statements in the new files.  These will break on MSVC.

obilaniu:
@abergeron Oh crap. I had fixed this in the merge-commit but somehow this got undone and there are conflict markers all over the place in my working directory at those lines, yet somehow I managed to push it in.

obilaniu:
@abergeron Ok, I need help, since this Git rebase stuff is apparently way over my head.

Before the rebase the fixes to the declarations-after-code were there, they disappeared without notice during that rebase. I had tar-xzipped my tree immediately after the quotes commit and before trying to do this rebase, and I've gone back to that.  `git branch -a` currently reads
```
* master
  remotes/origin/HEAD -> origin/master
  remotes/origin/master
  remotes/remote/master
```
and I `git push --force`'d the state of my repo on Github so that it is what it used to be. I'm now back at step 0, with everything as it was immediately before _"Please remove the merge commit and do a rebase if necessary."_.

What, _exactly_, do I enter in the command-line?

abergeron:
It depends.  It should be `git rebase remote/master` while in your branch.  However, if you've made some changes in the merge commit (in addition to merging branches then that is going to be a pain to manage.  Also instead of gziping your tree you can create a branch based on the current state so that you can go back to it in case it fails again.

We can talk tomorrow so disentangle this.

obilaniu:
@abergeron Unfortunately I *did* make changes in the merge commit. In particular, I edited the test CMakeLists to match the changes that had been done, and the other thing I did in the merge commit was to make those declarations ISO-C90 compatible. It's exactly those things that went \*poof!\* above.

Ok, we'll talk tomorrow.

abergeron:
jenkins test this please

PR #186, opened 2016-06-05
Add missing cuStreamDestroy a context's memory stream

tsirif:
This will add `cuStreamDestroy(ctx->mem_s)` in `static void cuda_free_ctx(cuda_context *ctx)` as discussed [here](https://groups.google.com/forum/#!topic/theano-dev/pRuqbclL_Cw).


nouiz:
thanks

PR #188, opened 2016-06-07
Add take1() to GpuArray.

abergeron:
I just realized that this is missing.


pygpu/gpuarray.pyx
```diff
@@ -1695,6 +1704,20 @@ cdef class GpuArray:
 
         array_setarray(tmp, gv)
 
+    def take1(self, GpuArray idx):
```
nouiz: Add tests for that method. It can be basic test just to make sure it work and we don't remove/break it by mistake.

nouiz:
It is good that you add it. Did you plan to use it in Theano?

abergeron:
Yes probably for the GpuAdvancedSubtensor python code.

abergeron:
Now with test™.

PR #190, opened 2016-06-08
Add a way to get a better description of blas errors.

abergeron:


src/gpuarray_blas_cuda_cublas.c
```diff
@@ -275,6 +278,39 @@ static void teardown(gpucontext *c) {
...
+    case CUBLAS_STATUS_EXECUTION_FAILED:
+      return "(cublas) Execution failed.";
+    case CUBLAS_STATUS_INTERNAL_ERROR:
+      return "(cublas) Internal error.";
+    case CUBLAS_STATUS_NOT_SUPPORTED:
+      return "(cublas) Unsupported functionality.";
+    case CUBLAS_STATUS_LICENSE_ERROR:
+      return "(cublas) License error.";
+    default:
+      return "(cublas) Unknown error.";
```
nouiz: Can you add the error code in the string?

src/gpuarray_blas_opencl_clblas.c
```diff
@@ -56,6 +56,10 @@ static void teardown(gpucontext *ctx) {
     clblasTeardown();
 }
 
+static const char *error(gpucontext *ctx) {
+  return "(clblas) error in blas call, no details for now.";
```
nouiz: I think this is a bad sentence. Even if there is no error, we tell there is an error. What about:

(clblas) We didn't implement the detailed error message. Maybe there is an error, maybe there isn't.

abergeron: This shouldn't be called when there are no errors.  I don't want to be too vague at this point and this is just to have something there.

I can spend another half-hour to get something about equivalent to the cublas thing.

src/gpuarray_blas_cuda_cublas.c
abergeron: No, since these are static strings.  I can't have anything variable.

nouiz: I don't understand, why can't you build a string and return the string you build? Is this because you fear you can't allocate memory on the CPU for the error string?

abergeron: Yes exactly, I can't allocate memory otherwise it would leak.

nouiz:
It seem good to me, but as the user had problem in the error message, I'll wait to merge this PR that he get a useful error message.

tsirif:
Is this a good idea about this? 
Read these [comments](https://github.com/tsirif/libgpuarray/blob/feature/nccl/src/gpuarray/buffer_collectives.h#L64) 
And this [impl](https://github.com/tsirif/libgpuarray/blob/feature/nccl/src/gpuarray_buffer_collectives.c#L19)

tsirif:
In addition with the error switch function :)

cmyr:
@nouiz I've run this branch and updated the original issue. No need to merge for that.

nouiz:
I saw the other comment. thanks. But the error message is strange and don't
help. I think there is a problem with the error message.

Can you share the code to reproduce it?

cmyr:
"sort of".

As I mentioned in the original issue, I'm working on a keras wrapper for theano's multigpu functionality, and the user-facing code that produces this error is a few keras functions.

the keras code produces a theano function. I can use d3viz to produce a graph of that; is there anything else I can do that would be useful?

abergeron:
@tsirif I would invert the switch and make gpucontext_error dispatch to gpucomm_error instead.

@cmyr I just realised that I didn't make any of the blas call actually save their errors in the new spot, I'll update this PR with that.

cmyr:
@abergeron ping me when the PR is updated and I'll get you an error message asap!

abergeron:
@cmyr I just updated the PR.

cmyr:
@abergeron updated but the error doesn't seem any different: https://github.com/Theano/Theano/issues/4597#issuecomment-225240049

I'm going for lunch will see if i made any obvious mistakes when I get back cheers

abergeron:
Yes, it changed, we now see "Execution failed" as the cublas error.

tsirif:
@abergeron Ok I will do this otherwise. But in that case there's no need to expose a `gpucomm_error`, because all errors will be handled from `gpucontext_error`. I will keep `comm_get_error` in the ops though.

cmyr:
@abergeron If it's at all useful: I'm actually in montreal, and (assuming you are) I'd be happy to meet up somewhere and look at this together?

PR #191, opened 2016-06-08
Add convert_f16 param to pygpu GpuElemwise.

abergeron:


nouiz:
Ideally, there should be minimal test of that here.

nouiz:
Ok to merge with some minimal tests.

abergeron:
Now with more test™.

PR #192, opened 2016-06-08
Add a shortcut to cuda_waits() in case we are on the same stream.

abergeron:
This helps lower the overhead of cuda_wait() for the single-stream case.

### DIFF
--- a/pygpu/gpuarray.pyx
+++ b/pygpu/gpuarray.pyx
@@ -498,9 +498,9 @@ cdef GpuContext pygpu_init(dev, int flags):
         raise ValueError, "Unknown device format:" + dev
     return GpuContext(kind, devnum, flags)
 
-def init(dev, sched='default', disable_alloc_cache=False):
+def init(dev, sched='default', disable_alloc_cache=False, single_stream=False):
     """
-    init(dev, opt='default', disable_alloc_cache=False)
+    init(dev, opt='default', disable_alloc_cache=False, single_stream=False)
 
     Creates a context from a device specifier.
 
@@ -510,6 +510,8 @@ def init(dev, sched='default', disable_alloc_cache=False):
     :type sched: {'default', 'single', 'multi'}
     :param disable_alloc_cache: disable allocation cache (if any)
     :type disable_alloc_cache: bool
+    :param single_stream: enable single stream mode
+    :type single_stream: bool
     :rtype: GpuContext
 
     Device specifiers are composed of the type string and the device
@@ -535,8 +537,8 @@ def init(dev, sched='default', disable_alloc_cache=False):
     expected_version = -9997
     if gpuarray_api_major != expected_version or gpuarray_api_minor < 0:
         raise RuntimeError(
-            "Pygpu was expecting libgpuarray version %d, but %d is available "
-            " Recompile it to avoid problems.",
+            "Pygpu was expecting libgpuarray version %d, but %d is available. "
+            "Recompile it to avoid problems.",
             expected_version, gpuarray_api_major)
     if sched == 'single':
         flags |= GA_CTX_SINGLE_THREAD
@@ -546,6 +548,8 @@ def init(dev, sched='default', disable_alloc_cache=False):
         raise TypeError('unexpected value for parameter sched: %s' % (sched,))
     if disable_alloc_cache:
         flags |= GA_CTX_DISABLE_ALLOCATION_CACHE
+    if single_stream:
+        flags |= GA_CTX_SINGLE_STREAM
     return pygpu_init(dev, flags)
 
 def zeros(shape, dtype=GA_DOUBLE, order='C', GpuContext context=None,
--- a/pygpu/gpuarray.pyx
+++ b/pygpu/gpuarray.pyx
@@ -500,7 +500,7 @@ cdef GpuContext pygpu_init(dev, int flags):
 
 def init(dev, sched='default', disable_alloc_cache=False, single_stream=False):
     """
-    init(dev, opt='default', disable_alloc_cache=False, single_stream=False)
+    init(dev, sched='default', disable_alloc_cache=False, single_stream=False)
 
     Creates a context from a device specifier.
 
--- a/pygpu/gpuarray.pyx
+++ b/pygpu/gpuarray.pyx
@@ -37,13 +37,13 @@ def cl_wrap_ctx(size_t ptr):
     Wrap an existing OpenCL context (the cl_context struct) into a
     GpuContext class.
     """
-    cdef gpucontext *(*cl_make_ctx)(void *)
+    cdef gpucontext *(*cl_make_ctx)(void *, int)
     cdef GpuContext res
-    cl_make_ctx = <gpucontext *(*)(void *)>gpuarray_get_extension("cl_make_ctx")
+    cl_make_ctx = <gpucontext *(*)(void *, int)>gpuarray_get_extension("cl_make_ctx")
     if cl_make_ctx == NULL:
         raise RuntimeError, "cl_make_ctx extension is absent"
     res = GpuContext.__new__(GpuContext)
-    res.ctx = cl_make_ctx(<void *>ptr)
+    res.ctx = cl_make_ctx(<void *>ptr, 0)
     if res.ctx == NULL:
         raise RuntimeError, "cl_make_ctx call failed"
     return res
--- a/src/gpuarray/buffer.h
+++ b/src/gpuarray/buffer.h
@@ -90,7 +90,7 @@ GPUARRAY_PUBLIC gpucontext *gpucontext_init(const char *name, int dev,
 #define GA_CTX_SINGLE_THREAD 0x02
 
 /**
- * Perform all operations in the order they appear.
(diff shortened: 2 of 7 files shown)


src/private_cuda.h
```diff
@@ -88,6 +88,7 @@ struct _gpudata {
      struct _partial_gpudata */
   CUevent rev;
   CUevent wev;
+  CUstream ls; /* last stream used */
```
nouiz: is this the last stream to read, write or both?

src/gpuarray_buffer_cuda.c
```diff
@@ -637,6 +643,7 @@ static int cuda_records(gpudata *a, int flags, CUstream s) {
   if (flags & CUDA_WAIT_WRITE)
     a->ctx->err = cuEventRecord(a->wev, s);
   cuda_exit(a->ctx);
+  a->ls = s;
```
nouiz: this is changed even if no flags was set. Should we return an error if no flag is set?

I'm not sure the shortcut logic is always valid. Is it fine to record more events then wait? I think so, but I'm not 100% sure.

I'm not sure the logic is good. I'll try to call you later this afternoon to discuss this.

nouiz:
Did you measure any speed difference? Should we also not record even when there is only 1 context?
We could just disable most of event if there is only 1 context. But this won't speed up multi-gpu.
Why we don't need to ls? one for read and one for write?

I think it is a good time to write how this work to have that documented. I always forget enough of the detail to slow down the review everytime we change this.

src/private_cuda.h
abergeron: This the last stream used to do anything.

src/gpuarray_buffer_cuda.c
abergeron: Normally wait/record calls should go in pairs, so the logic would be fine.  Of course if that is not respected, then there could be problems with the shortcut logic.

It's true that if there are no flags set, then we should probably not record the stream.

nouiz: I agree that we should call cuda_wait/cuda_record in pair. But what about the cuStreamWaitEvent and cuEventRecord. The current shortcut logic break those pair. I think it is fine, but I would like a confirmation.

abergeron: Yes that is fine, these don't need to go in pairs.

nouiz:
Can you give the timing information that you have? What is the timing with the old back-end, with the current master and with this PR with multi-stream and without-multi-stream.

src/gpuarray_buffer_cuda.c
```diff
@@ -71,9 +71,13 @@ cuda_context *cuda_make_ctx(CUcontext ctx, int flags) {
   if (err != CUDA_SUCCESS) {
     goto fail_stream;
   }
-  err = cuStreamCreate(&res->mem_s, CU_STREAM_NON_BLOCKING);
-  if (err != CUDA_SUCCESS) {
-    goto fail_mem_stream;
+  if (ISSET(res->flags, GA_CTX_SINGLE_STREAM)) {
+    res->mem_s = res->s;
+  } else {
+    err = cuStreamCreate(&res->mem_s, 0);
```
nouiz: You removed the NON_BLOCKING flag. Is that wanted? I don't understand why this should change.

src/gpuarray_buffer_opencl.c
```diff
@@ -90,9 +90,10 @@ cl_ctx *cl_make_ctx(cl_context ctx) {
   res->refcnt = 1;
   res->exts = NULL;
   res->blas_handle = NULL;
-  res->q = clCreateCommandQueue(ctx, id,
-				qprop&CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
-				&err);
+  res->q = clCreateCommandQueue(
```
nouiz: here you don't set "res->flags = flags;" as in cuda.

src/gpuarray/buffer.h
```diff
@@ -90,6 +90,19 @@ GPUARRAY_PUBLIC gpucontext *gpucontext_init(const char *name, int dev,
...
+ *
+ * This will remove any attempt at exploiting parallelism in the
+ * underlying device by performing unrelated operations concurrently
+ * and/or out of order.
+ *
+ * This can help performance by removing the small cost paid for each
+ * operation to keep everything coherent in the face of parallelism.
+ * It can also hinder performance by doing everything one after the other.
+ */
+#define GA_CTX_SINGLE_STREAM 0x4
```
nouiz: I'm not sure if this work with multiple context for CUDA and OpenCL. Can you confirm this and add this in the documentation.

This will also request a way to configure this in Theano.

pygpu/gpuarray.pyx
```diff
@@ -498,9 +498,9 @@ cdef GpuContext pygpu_init(dev, int flags):
         raise ValueError, "Unknown device format:" + dev
     return GpuContext(kind, devnum, flags)
 
-def init(dev, sched='default', disable_alloc_cache=False):
+def init(dev, sched='default', disable_alloc_cache=False, single_stream=False):
     """
-    init(dev, opt='default', disable_alloc_cache=False)
+    init(dev, opt='default', disable_alloc_cache=False, single_stream=False)
```
nouiz: Change opt for sched.

pygpu/gpuarray.pyx
```diff
@@ -510,6 +510,8 @@ def init(dev, sched='default', disable_alloc_cache=False):
     :type sched: {'default', 'single', 'multi'}
     :param disable_alloc_cache: disable allocation cache (if any)
     :type disable_alloc_cache: bool
+    :param single_stream: enable single stream mode
```
nouiz: I didn't understood at first the docstring for sched. Can you tell a little bit more about it? From the flag it cause later, I think it is for single vs multi threads. You could link to nvidia doc if it help.

src/gpuarray_buffer_cuda.c
abergeron: Yes, this will lead to less problems for interacting with libraries that use the NULL stream since all the internal streams will synchronize with it implicitly.

src/gpuarray_buffer_opencl.c
abergeron: There is no flags field on for the opencl context.

src/gpuarray/buffer.h
abergeron: Yes, there is no problem with multiple contexts.

There is no way to prevent multiple contexts from working in the whole library and I will never add something that could.  So I don't why there should be a note on every single function and flag documenting that fact.

Yes, we might want to add an option to Theano, but this was mainly test code to see how much of an effect it would have.  Since it did have a major impact, I decided to keep it in.

pygpu/gpuarray.pyx
abergeron: What?  I don't get what you want.

abergeron: Oh, I got it. The name in the prototype is wrong.

pygpu/gpuarray.pyx
abergeron: Yes that is exactly it.  The flags determines the synchronization mechanism to use. One of them is heavy on the CPU, but fast.  The other will yield, but introduces additional delays when synchronizing.

src/gpuarray_buffer_cuda.c
nouiz: Can you add a comment about that? This is the thing that woudl prbably be forgotten when you write the doc about this.

src/gpuarray_buffer_opencl.c
nouiz: ok. I don't want to spend time on this, but I find that strange. If this is something that should be added, make an issue for this. Otherwise, I'll be surprised again when I see this next time:)

src/gpuarray/buffer.h
nouiz: Can you at least tell that this is one stream per context? Like in the first line:

 * Use only 1 steram per context, so this force to perform all operations in the order they appear for one context.

abergeron: I've updated the doc.

src/gpuarray_buffer_cuda.c
abergeron: I've added a comment.

src/gpuarray_buffer_opencl.c
abergeron: I don't need to keep the flags around for anything in the OpenCL backend.  If there is ever a need for that, then I'll add the field and fill it.

nouiz:
Before merging this PR, I would like the PR on Theano to be done and the profiling result. Is there still slowdown in the bench-mark with the new back-end?

abergeron:
In single-stream mode no.  There is a speedup actually.  But a profiling run shows that GpuCAReduce, GpuAdvancedSubtensor1 and GpuAdvancedSubtensor1_dev20 still have some slowdown and the amount of GpuElemwise node is not the same (which does lead to GpuElemwise taking more total time on the new backend).

I don't think we should wait on a PR to Theano simply because this does not introduce any backwards incompatibilities.

PR #193, opened 2016-06-08
Wrap multi-gpu communicator and collectives functionality of nccl

tsirif:
A first commit to declare the API. Pull request to keep track of this branch's commits.


src/gpuarray/buffer.h
```diff
@@ -24,25 +24,31 @@ extern "C" {
...
 typedef struct _gpucontext gpucontext;
 
+struct _gpukernel;
 
 /**
- * Opaque struct for buffer data.
+ * Opaque struct for kernel data.
  */
-typedef struct _gpudata gpudata;
+typedef struct _gpukernel gpukernel;
```
abergeron: Why did you have to move those?

src/gpuarray/buffer.h
```diff
@@ -757,6 +763,48 @@ typedef enum _ga_usefl {
...
+/**
+ * \brief Creates a new communicator (multi process version)
+ * \param clique_id [const char*] TODO
+ * \param ndev [int] TODO
+ * \param rank [int] TODO
+ * \param res [int*] TODO
+ * \return gpucomm* TODO
+ */
+GPUARRAY_PUBLIC gpucomm* gpucomm_init(const char* clique_id, int ndev, int rank,
+                                      int* res);
```
abergeron: Maybe this should also take a `gpucontext *` to identify the device to work with.

src/gpuarray/buffer.h
```diff
@@ -757,6 +763,48 @@ typedef enum _ga_usefl {
...
+ * \param err [int] TODO
+ * \return const char* TODO
+ */
+GPUARRAY_PUBLIC const char* gpucomm_error(gpucomm* comm, int err);
+
+// TODO cuda extension:
+// 1. ncclGetUniqueId
+// 2. ncclCommCount (below mby ops in gpucomm struct?!)
+// 3. ncclCommUserRank
+// 4. ncclCommCuDevice
```
abergeron: I don't think those should be cuda-specific extensions.  We should arrive at some sort of API we want to support, like what you declared for GpuArray and have those APIs at the gpucomm level too.

src/gpuarray/array.h
```diff
@@ -607,6 +607,78 @@ GPUARRAY_PUBLIC void GpuArray_fprintf(FILE *fd, const GpuArray *a);
...
+ * \param count [int] TODO
+ * \param typecode [int] TODO
+ * \param opcode [int] TODO
+ * \param root [int] TODO
+ * \param comm [gpucomm*] TODO
+ * \return int TODO
+ */
+GPUARRAY_PUBLIC int GpuArray_reduce(const GpuArray* src, GpuArray* dest,
+                                    int count, int typecode, int opcode,
+                                    int root, gpucomm* comm);
```
abergeron: Ideally I don't want to mix levels like this (GpuArray and gpucomm).  If you absolutely need to then maybe we should implement those functions at the gpucomm level only and wait before we have a higher-level wrapper.

abergeron:
I've made a few comments on the API structure, but otherwise it seems to be going in a good direction.

src/gpuarray/buffer.h
tsirif: It does not change anything in functionality. I reordered them in order to be in line :P

src/gpuarray/buffer.h
tsirif: :+1:

src/gpuarray/buffer.h
abergeron: Ok

src/gpuarray/buffer.h
tsirif: :+1: 
`ncclGetUniqueId`, which is a helper function independent with gpus, shouldn't this suit more a as cuda extension? I mean it's not a gpu-related functionality.

src/gpuarray/array.h
tsirif: I think I got this. The design with which you developed this. It would be good to write documentation about this. Question: Why did you choose C over C++? :)

abergeron: More documentation is always good.  However the time to write it is always missing.  

Also, I chose C to be more compatible with other languages what might want to bind to the library.

tsirif: I am trying to think of a array-level wrapper, that would not include the privately defined `gpucomm`. But I think we will have to cope with some unnecessary difficulties:

A `GpuArray` holds a single `gpudata` pointer and "metadata". A `gpudata` instance holds a single `gpucontext`. So from any `GpuArray` you can access the `gpucontext`. But a gpu can be associated with multiple `gpucomm`s, because a user could want to have a group of gpus to execute a certain operation together and a different group to execute another on the same `GpuArray`. For example, one could define a grid of gpus with their respective `GpuArray`s and would like first to concurrently execute an AllReduce sum operation on the columns and then to concurrently execute an AllReduce product operation on the rows so that we can get - let's say - the Πi(Σj arrays[i, j]) on every gpu. And I do not believe that the user can spare the multiple `gpucomm`s per gpu in any implementation of this or that we should restrict them to use one `gpucomm` per gpu. So, because I believe that we cannot associate a `gpucontext` with a dynamic array of `gpucomm`s without providing a "clumsy"** interface for the user to act on a desired `gpucomm` in array-level, I think that we should expose `gpucomm` to the array-level and let the user manage where he thinks he should use it. As an argument in favor of this, nccl or mpi library would not expose the communication worlds if there could be a more user-friendly way without restricting the possibilities.  

** By "clumsy" I mean that an interface of `GpuArray_all_reduce`, which would include an integer to have user select one of the already registered `gpucomm` within a `gpucontext`, would not be well-defined.

abergeron: I agree that the reduce operations would need to take a gpucomm parameter to know with what to reduce.  What I'm proposing is that we only build the low-level interface that takes gpucomm and gpudata for now, and think about a high-level one later.

I agree with you that passing an index or something like that is a bad idea.

src/gpuarray/buffer.h
abergeron: I'm not sure we actually need to expose it at all.  We only need to call it ourselves to pass it to ncclCommInitRank and that's it.  It's not used anywhere else.

tsirif: But pygpu should bind to libgpuarray and only to that. How one will get the unique id through pygpu interface without involving a python bind to nccl? I think of the platoon case, where someone should get a unique id somehow and write in python.

abergeron: I mean that libgpuarray can get the unique ID and use it in the gpucomm_init call. I don't see any reason to do it separately.

tsirif: The unique clique id needs to be exposed in order to support multi-node/gpu as every call to `ncclGetUniqueId` returns a different id every time that needs to be shared among participants who will compose a CommWorld.

Quoting [this](https://devblogs.nvidia.com/parallelforall/fast-multi-gpu-collectives-nccl/): 
"This function assumes that the GPU belonging to the specified rank has already been selected using cudaSetDevice(). nGPUs is the number of GPUs in the clique. cliqueId allows the ranks of the clique to find each other. The same cliqueId must be used by all ranks. To achieve this, call ncclGetUniqueId() in one rank and broadcast the resulting uniqueId to the other ranks of the clique using the communication framework of your choice (for example, MPI_Bcast)."

And [this](https://github.com/NVIDIA/nccl/blob/master/test/mpi/mpi_test.cu#L62) test example of theirs.

tsirif: Still is a helper function that is somewhat independent with main functionality and it's not necessary to provide a multi-gpu collectives framework (if there will exist a one in opencl), so we cannot guarantee that a framework (other than nccl) will provide a similar one. This is why I propose to implement this as a cuda extension.

abergeron: I see that you are right about the id.  I somehow had some other idea in mind about how it worked.

The problem I have with a cuda extension is that we still need a way to pass it through as the clique_id parameter.  This would make the extension a required part for it to work, which makes it not exactly an extension, but rather a different form of binding.

I would prefer some wrapper that returns a `void *` or `char *` pointer.  However the problem remains that we need to tell the user how big it is so that they can copy it over to the other processes and use it.  We could always mimic the approach that nccl has taken with a dummy struct of a given size or we could return a pointer and size.

tsirif: I don't believe that a call to `gpucomm_gen_clique_id` is an obligation for things to work out, because the user can provide a unique clique_id of his own creation. But, yes, most of the times it would be useful to use this helper function. To sum up we have the follow restrictions:

1. We need to provide the byte-array size with the unique byte-array id: Yes, we could provide a dummy struct of a given size along with a macro which defines the "internal" byte-array size. I prefer this approach in general, but because of the case of a possible opencl collectives implementation, it's hard to know what this constant byte-array length will be. The second option seems better in the cost of having a larger interface. We could also provide just the `const char*` along with macros for each libgpuarray predefined backend. It is user's responsibility to know which implementation is actually using anyways. I will first go for the dummy struct which will at first be just a combatible type with `ncclUniqueId`.

2. The necessity of this function: Ok I will place this function in `gpucomm`, but as it is not a crucial part of such an interface, I recommend to provide ourselves an implementation of this functionality to be used in cases where one is not provided natively by the backend (opencl may choose not to provide such a helper function, but leave it up to the user).

tsirif:
@abergeron Can you commend on these? 
[1](https://github.com/tsirif/libgpuarray/blob/feature/nccl/src/gpuarray_buffer_collectives.c#L46) 
[2](https://github.com/tsirif/libgpuarray/blob/feature/nccl/src/gpuarray/collectives.h#L72)

abergeron:
1) Yes those checks should be made however sz is the full size of the buffer including any offset.
2) It would be best to be as flexible as possible with regards to layouts, however it's ok to restrain the operations to contiguous buffers if necessary.

src/gpuarray/buffer.h
```diff
@@ -490,11 +490,14 @@ GPUARRAY_PUBLIC int gpukernel_binary(gpukernel *k, size_t *sz, void **obj);
 GPUARRAY_PUBLIC int gpukernel_property(gpukernel *k, int prop_id, void *res);
 
 GPUARRAY_PUBLIC gpucontext *gpukernel_context(gpukernel *k);
+
 /**
  * \defgroup props Properties
  * @{
  */
 /* Start at 1 for GA_CTX_PROP_ */
+#define GA_CTX_PROP_START    1
```
abergeron: I don't feel this is relevant.

src/gpuarray/buffer.h
```diff
@@ -634,8 +637,16 @@ GPUARRAY_PUBLIC gpucontext *gpukernel_context(gpukernel *k);
  */
 #define GA_CTX_PROP_MAXLSIZE2 17
 
+/**
+ * Get the vector of collective ops for the context.
+ *
+ * Type: `const gpuarray_comm_ops *`
+ */
+#define GA_CTX_PROP_COMM_OPS  18
```
abergeron: I would prefer to make that a field in the gpucontext struct.

src/gpuarray/buffer_collectives.h
```diff
@@ -31,47 +31,43 @@ enum _gpucomm_reduce_ops {
...
+  GA_MIN = 3,
 };
 
 #define GA_COMM_ID_BYTES 128  // sizeof(gpucommCliqueId)
 /**
  * Dummy struct to define byte-array's length through a type
  */
 typedef struct _gpucommCliqueId {
-  char internal[GA_COMM_ID_BYTES];
+  char (*internal)[][GA_COMM_ID_BYTES];
```
abergeron: Why make the internal a pointer to a char table?  This also makes the struct 4 to 8 times bigger.

src/gpuarray/buffer_collectives.h
```diff
@@ -31,47 +31,43 @@ enum _gpucomm_reduce_ops {
...
+                                gpucommCliqueId cliqueId, int ndev, int rank);
 
 /**
  * \brief TODO
  * \param comm [gpucomm*] TODO
- * \return void TODO
+ * \return int TODO
  */
-GPUARRAY_PUBLIC void gpucomm_free(gpucomm* comm);
+GPUARRAY_PUBLIC int gpucomm_free(gpucomm* comm);
```
abergeron: The free operation should not fail.

src/gpuarray/buffer.h
tsirif: yeap sorry, I saw what you wrote about magic numbers and I added 3 more macros to define positions of CTX, BUFFER, KERNEL in `Properties` defgroup

src/gpuarray_buffer_collectives.c
```diff
@@ -4,43 +4,58 @@
...
+  }
+  return ctx->comm_ops->comm_new(comm, ctx, cliqueId, ndev, rank);
 }
 
-void gpucomm_free(gpucomm* comm) {
-  gpucomm_context(comm)->comm_ops->comm_free(comm);
+int gpucomm_free(gpucomm* comm) {
+  gpucontext* ctx = gpucomm_context(comm);
+  if (ctx->comm_ops == NULL)
+    return GA_COMM_ERROR;
```
abergeron: That shouldn't be an error.

src/gpuarray/buffer.h
tsirif: I don't understand what you mean exactly. I thought you do the same for blas.

src/gpuarray_buffer_collectives.c
```diff
@@ -4,43 +4,58 @@
...
+  return ctx->comm_ops->get_count(comm, count);
 }
 
-int gpucomm_get_device(const gpucomm* comm, int* device) {
-  return gpucomm_context(comm)->comm_ops->get_device(comm, device);
+int gpucomm_get_device(gpucomm* comm, int* device) {
+  gpucontext* ctx = gpucomm_context(comm);
+  if (ctx->comm_ops == NULL)
+    return GA_COMM_ERROR;
+  return ctx->comm_ops->get_device(comm, device);
```
abergeron: This should return a gpucontext ideally.

src/gpuarray/buffer_collectives.h
tsirif: Yes I revert this back. This is wrong.

src/gpuarray_buffer_collectives.c
tsirif: In case user calls this, while comm_ops are not supported (because he ignored the errors returned), should we let this cause a fault?

src/gpuarray/buffer.h
abergeron: The stuff for blas is a historical accident.  It'll go away at some point.

src/gpuarray/buffer.h
abergeron: The other ones are good.

src/gpuarray/buffer.h
tsirif: Do you have some time to chat? Because I find myself mimicking the way that you've wrapped things for blas in order to be consistent with your design style, but if you say that you' ve changed your mind, I think we should discuss this. Can you tell me what happened: what did you like and what you did not? Things like that.

src/gpuarray_buffer_collectives.c
abergeron: No, it should just skip the call below in that case.

src/gpuarray_buffer_collectives.c
tsirif: so be it :)

src/gpuarray/buffer.h
abergeron: The only part that changed is the property part, the rest is ok, so if you follow that you're good.

src/gpuarray_buffer_collectives.c
abergeron: Actually, since you have gpucomm_context, it should probably just not be present.

src/gpuarray/buffer.h
tsirif: Do you have these new changes about property in some branch, so I can check them?

abergeron: It should be in the master.  I just left the GA_CTX_PROP_BLAS_OP thing there so as not to break existing code for now.

src/gpuarray_collectives_cuda_nccl.c
```diff
@@ -53,8 +54,25 @@ static void comm_free(gpucomm* comm) {
 
 static const char* comm_error(gpucontext* c) {
   cuda_context* ctx = (cuda_context*) c;
-  // find a way to concatenate a constact "(nccl) " infront
-  return ncclGetErrorString(ctx->nccl_err);
```
abergeron: If you can get an error string from nccl, then we can forget about the "(nccl) " in front.  cublas doesn't offer a cublasGetErrorString() so that's why it went this way.

src/gpuarray_collectives_cuda_nccl.c
```diff
@@ -95,6 +95,31 @@ static int get_rank(const gpucomm* comm, int* rank) {
...
+  switch(opcode) {
+  case GA_SUM : return ncclSum;
+  case GA_PROD: return ncclProd;
+  case GA_MAX : return ncclMax;
+  case GA_MIN : return ncclMin;
+  }
+  return nccl_NUM_OPS;
+}
+
+inline ncclDataType_t convert_data_type(int typecode) {
```
abergeron: Declare those static so that they don't pollute the symbol table.

src/gpuarray_collectives_cuda_nccl.c
```diff
@@ -0,0 +1,353 @@
...
+do { \
+  int err = (cmd); \
+  if (err != GA_NO_ERROR) \
+    return err; \
+} while (0)
+
+
+extern const gpuarray_buffer_ops cuda_ops;
+
+void comm_clear(gpucomm* comm)
```
abergeron: All of these should be static.

src/private.h
```diff
@@ -48,7 +54,7 @@ typedef struct _gpuarray_blas_ops gpuarray_blas_ops;
 struct _gpucontext {
   GPUCONTEXT_HEAD;
   void *ctx_ptr;
-  void *private[7];
+  void *private[8];
```
abergeron: Do you really need to grow this?  If you don't then it's best to leave it as it is.  There is a static assertion in the source that will trigger if it's too small.

src/gpuarray_collectives_cuda_nccl.c
tsirif: So I should declare/define as static whatever I choose not to expose. So pretty much every helper function in this file?

src/private_cuda.h
```diff
@@ -50,6 +59,9 @@ typedef struct _cuda_context {
   gpudata *freeblocks;
   cache *kernel_cache;
   unsigned int enter;
```
abergeron: Move the err member below this one so that they can be packed together and avoid padding.

src/private.h
tsirif: Does void* [7] suffice to hold a `ncclResult_t` variable (it's an enum)? I will check this through the static idd.

src/private_cuda.h
```diff
@@ -128,4 +140,14 @@ struct _gpukernel {
 #endif
 };
 
+struct _gpucomm {
+  cuda_context* ctx;  // Start after the context
+// #ifdef WITH_CUDA_NCCL
+  ncclComm_t c;
+// #endif  // WITH_CUDA_NCCL
```
abergeron: For now this doesn't compile without nccl.  It's ok for development, but think about some way to not have variable-witdh structures that will also work without nccl.

src/gpuarray_collectives_cuda_nccl.c
abergeron: Yes.

src/private.h
abergeron: An enum is an int so 4 bytes.  A pointer is 4 or 8 bytes (depending on 32 or 64 bits systems).  However int can be packed together if placed one after the other so that they take a single 8-bytes slot when possible.

src/private_cuda.h
tsirif: Below `unsigned int enter`? (it's already)

abergeron: Oh yeah, the nccl error is already there.

src/private_cuda.h
tsirif: Definition of `struct _gpucomm` should surely contain the ncclComm_t, but in the case that user does not fulfill the optional dep of nccl, the code should be able to be built nevertheless (that's why I will uncomment the defs around `ncclComm_t c` when code is overall cool). Right now I have these on comments, so I can work with autocompletion :stuck_out_tongue: . But I do not think we can somehow avoid the def around, because nccl is an optional dep.

abergeron: I agree that the struct should contain it.  I'm just wondering if it has to be defined here, or if it could be defined in the nccl file and just have a `struct _gpucomm;` declaration.

tsirif: I will search if the rest definition (gpucomm->ctx, where ctx is a `cuda_context`) is needed elsewhere and I will respond. Probably not, so you would be right. What about the `ncclResult_t` inside of `cuda_context`??

abergeron: You could move it to the gpucomm struct and make gpucomm_error take a gpucomm instead.  But that would break the central thing with gpucontext_error so I don't know.

I would like to avoid forcing users to create giant switches based on the error code to get an accurate error description in any case.  Maybe having a place to store a string in the gpucontext struct that would be the error message for the last error?

tsirif: Allocating static array inside of a `gpucontext` and calling always `ncclGetErrorString(...)`?  Do we want to avoid the macros inside `cuda_context` for sure?

abergeron: I mean having a char * field in cuda_context and setting the pointer with ncclGetErrorString().  This can also be extended for other things (like the blas errors).

I would like to avoid the macros inside cuda_context so that there won't be some cases where the struct will fit in the generic size and others where it doesn't.  I also don't feel that adding a status code for every cuda library in the context is a viable long-term approach.

src/gpuarray/buffer.h
tsirif: I still cannot find what do you mean. Do you want to replace the "properties" macros with a enum defined inside the `gpucontext`?

abergeron: After a review of all the impacted code, it's probably better to leave it like this.  I can always change it later when I have a clearer idea of what I want.

src/gpuarray_array_collectives.c
```diff
@@ -0,0 +1,108 @@
...
+                                  int times_dest, const GpuArray* dest, int* count)
+{
+  int count_src, count_dest;
+  count_src = find_total_elems(src);
+  count_dest = find_total_elems(dest);
+  if (times_src * count_src != times_dest * count_dest)
+    return GA_VALUE_ERROR;
+  if (src->typecode != dest->typecode)
+    return GA_VALUE_ERROR;
+  if (!(src->flags & GA_ALIGNED) || !(dest->flags & GA_BEHAVED))
```
abergeron: The GA_BEHAVED check might not do what you want since it will pass if something is aligned OR writeable and I suppose you want aligned AND writable.

The correct code would be `(dest->flags & GA_BEHAVED == GA_BEHAVED)`.  To make sure you get this correctly you can use the `GpuArray_CHKFLAGS()` function

src/gpuarray/collectives.h
```diff
@@ -19,6 +19,17 @@ extern "C" {
 /**
  * \brief TODO
  * \param src [const GpuArray*] TODO
+ * \param opcode [int] TODO
+ * \param root [int] TODO
+ * \param comm [gpucomm*] TODO
+ * \return int TODO
+ */
+GPUARRAY_PUBLIC int GpuArray_reduce_from(const GpuArray* src, int opcode, int root,
+                                         gpucomm* comm);
```
abergeron: This function should be declared in gpuarray/array.h and implemented in gpuarray_array.c

src/gpuarray_array_collectives.c
```diff
@@ -0,0 +1,108 @@
...
+                                            int opcode, gpucomm* comm)
+{
+  int count = 0, ndev = 0;
+  GA_CHECK(gpucomm_get_count(comm, &ndev));
+  GA_CHECK(check_gpuarrays(1, src, ndev, dest, &count));
+  return gpucomm_reduce_scatter(src->data, src->offset, dest->data, dest->offset,
+                                count, src->typecode, opcode, comm);
+}
+
+GPUARRAY_PUBLIC int GpuArray_broadcast(GpuArray* array, int root, gpucomm* comm)
```
abergeron: No GPUARRAY_PUBLIC tag in source files.

src/gpuarray_array_collectives.c
```diff
@@ -0,0 +1,108 @@
+#include "gpuarray/array.h"
+#include "gpuarray/buffer_collectives.h"
+#include "gpuarray/collectives.h"
+#include "gpuarray/error.h"
+
+#include "private.h"
+
+static inline int find_total_elems(const GpuArray* array)
+{
```
abergeron: Make this return a `size_t`.  `int` is too small.

tsirif: Eventually nccl functions will get an `int count`. Should I convert later? Should we put a constraint for not too large arrays?

https://github.com/NVIDIA/nccl/blob/master/src/nccl.h#L130

src/gpuarray_array_collectives.c
tsirif: I forgot to remove this :P

src/gpuarray/collectives.h
tsirif: It's simply the same as GpuArray_reduce, but for non root ranks of communicator world. Should it go to array.h/c, array.c will include `gpuarray/buffer_collectives.h`. Is this good?

abergeron: Sorry I was confused.  It's fine where it is.

src/gpuarray_array_collectives.c
abergeron: I know, but I would prefer to hide that specific detail inside the API in case they ever decide to keep up with reality and switch to size_t.

Otherwise this limits the size that we can handle to 4G and we would have to expose a new API to handle 64-bits later.

abergeron: You can always add a check inside the nccl wrappers to make sure that we don't actually bust 4G.

tsirif: I will do both :)

CMakeModules/FindNCCL.cmake
```diff
@@ -0,0 +1,24 @@
...
+# The following variables are optionally searched for defaults
+#  NCCL_ROOT_DIR:    Base directory where all NCCL components are found
+#
+# The following are set after configuration is done:
+#  NCCL_FOUND
+#  NCCL_INCLUDE_DIR
+#  NCCL_LIBRARY
+
+find_path(NCCL_INCLUDE_DIR NAMES nccl.h
+    PATHS ${NCCL_ROOT_DIR}/include
```
abergeron: Are there any "standard" path where it is most likely to be found (like /usr/local/cuda, or something like that).

tsirif: User can define the extra environmental `NCCL_ROOT_DIR` to show cmake where to search for in addition. Also, system path and `PATH` variable are searched anyways. So supposing we are on a unix/linux system and user has installed NCCL (or whatever in a certain \<dir\> which has bin/include/lib ), it suffices to have \<dir\>/bin included in `PATH`. And usually in order for a shell to find executables, user (or the system for him) has already properly defined the `PATH` to include at least '\<dir\>/cuda', '/usr/local/bin' and '/usr/bin'.

See [find_path documentation](https://cmake.org/cmake/help/v3.4/command/find_path.html):  
> 4. Search the standard system environment variables. This can be skipped if NO_SYSTEM_ENVIRONMENT_PATH is an argument.
>    * Directories in INCLUDE, <prefix>/include/<arch> if CMAKE_LIBRARY_ARCHITECTURE is set, and <prefix>/include for each <prefix>/[s]bin in PATH, and <entry>/include for other entries in PATH, and the directories in PATH itself."

So, I think this is fine. By the way I found this [here](https://github.com/NVIDIA/caffe/blob/caffe-0.15/cmake/Modules/FindNCCL.cmake).

src/gpuarray_collectives_cuda_nccl.c
```diff
@@ -28,7 +28,7 @@ static const char* nccl_success_error = ncclGetErrorString(ncclSuccess);
       (ctx)->error_msg = ncclGetErrorString(nccl_err); \
       return GA_COMM_ERROR;                            \
     }                                                  \
-    (ctx)->error_msg = nccl_success_error;             \
```
abergeron: I would say that this is unnecessary.  We can leave whatever was there before if there is no error.

tests/check_buffer_collectives.h
```diff
@@ -0,0 +1,209 @@
+#include <math.h>
```
abergeron: Why is this file a .h?

abergeron: Never mind, it's fixed later.

tsirif: Was the question "why I used math.h"? I didn't understand

src/gpuarray_collectives_cuda_nccl.c
```diff
@@ -253,7 +253,7 @@ static int reduce(gpudata* src, size_t offsrc, gpudata* dest, size_t offdest,
   GA_EXIT_ON_ERROR(ctx, cuda_wait(src, CUDA_WAIT_READ));
   // sync: wait till a read/write has finished (out of concurrent kernels)
   if (rank == root)
-    GA_EXIT_ON_ERROR(ctx, cuda_wait(dest, CUDA_WAIT_WRITE));
+    GA_EXIT_ON_ERROR(ctx, cuda_wait(dest, CUDA_WAIT_ALL));
```
abergeron: Are you reading from the destination?

tests/check_buffer_collectives.h
```diff
@@ -74,35 +101,37 @@ typedef unsigned long ulong;
...
+    ck_assert_int_eq(err, GA_NO_ERROR);                                             \
+                                                                                    \
+    err = gpucomm_reduce(Adev, 0, RESdev, 0, count, gatype, gaoptype, ROOT_RANK,    \
+                         comm);                                                     \
+    ck_assert_int_eq(err, GA_NO_ERROR);                                             \
+    gpudata_sync(RESdev);                                                           \
+    gpudata_sync(Adev);                                                             \
+                                                                                    \
+    err = MPI_Reduce(A, EXP, count, mpitype, mpioptype, ROOT_RANK, MPI_COMM_WORLD); \
+    ck_assert_msg(err == MPI_SUCCESS, "openmpi error: cannot produced expected");   \
```
abergeron: I am questioning the use of MPI here.  Can we just do a single-thread reduction/broadcast/whatever?

It will not be parallel and performant, but that doesn't matter because it's just to have a reference value.

tests/check_buffer_collectives.h
abergeron: No it was about check_buffer_collectives.h, but you can't comment on filenames.

tests/check_buffer_collectives.h
tsirif: You are right. In reduce operation, we can have a reference value and a single thread. But this would not be convenient in the case where AllGather or ReduceScatter need to be tested, because an expected result is not easy to be created (or I got lazy). But anyways for generality, we can test against MPI to see whether we get the expected results anyways. Note that this code can test with whatever input easily without many changes. These changes would be required if we were having an expected reference value. Also, we can test for randomized inputs easily.  

Also, we cannot escape the requirement completely for multi-thread or process because for example `gpucomm_new` blocks and needs to be called in parallel.

tsirif: So far, I have run the latest tests on my pc with 1 process/gpu and they pass.

abergeron: Ok, the code is there so let's leave it like this

src/gpuarray_collectives_cuda_nccl.c
tsirif: [This](https://github.com/tsirif/libgpuarray/blob/feature/nccl/src/gpuarray_buffer_cuda.c#L612) is something I would like to discuss.

When we want to read a `gpudata`, we should wait for previously issued (earlier scheduled) writes to this thread to finish.
When we want to write to a `gpudata`, we should wait for previously issued (earlier scheduled) reads and writes to this thread to finish.

Also, in the case where all kernels run on the same stream, nothing is concurrent. So for the device there are no synchronization issues, because kernels are executed serially with the order they were issued. Synchronization issues occur in this case only between the stream and the host.

Do you agree on these? Shouldn't CUDA_WAIT_(TO)_WRITE also make the stream to wait for `a->wev` too? (And record to `a->wev` also when it finishes)

abergeron: CUDA_WAIT_WRITE does record to `wev` already.

However, we need to wait for previous writes as you say.

If you make this change, please do it in a separate PR, this one is big enough :)

tests/check_array.c
```diff
@@ -143,23 +140,23 @@ START_TEST(test_take1_ok)
...
 
   /* 1 */
-  ck_assert(buf[ 6] == 18);
-  ck_assert(buf[ 7] == 19);
-  ck_assert(buf[ 8] == 20);
-  ck_assert(buf[ 9] == 21);
+  ck_assert(buf[6] == 18);
+  ck_assert(buf[7] == 19);
+  ck_assert(buf[8] == 20);
+  ck_assert(buf[9] == 21);
```
abergeron: Now it's not aligned anymore.

tests/check_array.c
```diff
@@ -245,7 +242,8 @@ START_TEST(test_take1_ok)
 }
 END_TEST
 
-Suite *get_suite(void) {
+Suite *get_suite(void)
+{
```
abergeron: This is an ugly style.

tests/check_array.c
```diff
@@ -22,17 +22,14 @@ START_TEST(test_take1_ok)
...
   GpuArray vidx;
   GpuArray vres;
-  static const uint32_t data[24] = { 0,  1,  2,  3,  4,  5,
-                                     6,  7,  8,  9, 10, 11,
-                                    12, 13, 14, 15, 16, 17,
-                                    18, 19, 20, 21, 22, 23};
-  uint32_t buf[12*24];
+  static const uint32_t data[24] = {0,  1,  2,  3,  4,  5,  6,  7,  8,  9,  10, 11,
+                                    12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23};
+  uint32_t buf[12 * 24];
```
abergeron: Does this still fit in 79 characters?

tests/check_util.c
```diff
@@ -27,13 +29,12 @@ START_TEST(test_type_flags)
   ck_assert_int_eq(gpuarray_type_flags(GA_DOUBLE, -1), GA_USE_DOUBLE);
   ck_assert_int_eq(gpuarray_type_flags(GA_CFLOAT, -1), GA_USE_COMPLEX);
   ck_assert_int_eq(gpuarray_type_flags(GA_CDOUBLE, -1),
-                   GA_USE_DOUBLE|GA_USE_COMPLEX);
-  ck_assert_int_eq(gpuarray_type_flags(GA_HALF, -1), GA_USE_HALF|GA_USE_SMALL);
+                   GA_USE_DOUBLE | GA_USE_COMPLEX);
```
abergeron: This is an highly unusual style.

abergeron:
I'm not a fan of most of the style changes you applied to the test code.  It makes it less readable in my eyes and clashes with the rest of the codebase.

I've pointed the two most egregious examples.

tests/check_array.c
tsirif: I am using these right now:
```
let g:clang_format#code_style = "google"  
let g:clang_format#style_options = {  
      \ "ConstructorInitializerIndentWidth" : 2,  
      \ "ColumnLimit" : 85,  
      \ "BreakBeforeBraces" : "Stroustrup",  
      \ "AccessModifierOffset" : -4,  
      \ "AlwaysBreakTemplateDeclarations" : "true",  
      \ "AllowShortIfStatementsOnASingleLine" : "false",  
      \ "AllowShortCaseLabelsOnASingleLine" : "true",  
      \ "AllowShortLoopsOnASingleLine" : "false",  
      \ "IndentCaseLabels" : "false",  
      \ "Standard" : "C++11"}  
```

tsirif: I have column limit on 85, should I lower this to 80?

tests/check_array.c
tsirif: Aligning in consequent lines like this is not imposed everywhere. e.g. [here](https://github.com/Theano/libgpuarray/blob/master/tests/check_array.c#L119), but I am reverting these back. :)

tests/check_array.c
abergeron: Yes, lower it to 85.  Also BreakBeforeBraces should be "Attach" (if I read correctly).

And I don't know if there is a way to make it not insert spaces next to binary or (`|`), but that is much clearer.

tests/check_array.c
abergeron: It's not a huge deal.  If your tool can fix most other things but this, it would be ok.

Makefile
```diff
@@ -15,27 +15,35 @@ Debug:
...
+	(cd Release && NUM_DEVS=${NUM_DEVS} DEV_NAMES=${DEV_NAMES} cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$(INSTALL_PREFIX))
 endif
 
 relc: Release/Makefile
-	(cd Release && make && $(SUDO) make install)
+ifndef DEVICE
+	cd Release && make && make test && $(SUDO) make install
+else
+	cd Release && make && DEVICE=${DEVICE} make test && $(SUDO) make install
+endif
```
abergeron: I would make a separate test target rather than force the tests to run on each install.

tsirif:
@nouiz @abergeron Please check the latest commit, I think it's ready to be merged. All tests pass. Considering this pull request, the tests are `tests/check_buffer_collectives.c` and `tests/check_collectives.c`. Please examine them to tell me if they are enough to support this PR.

Building & Testing for tests: `check_collectives` and `check_buffer_collectives`  
1. `make debugc`
2. `mpiexec -np <#devices> ./Debug/tests/<test name> <gpu0> <gpu1> ... <gpuN>`

e.g. `mpiexec -np 2 ./Debug/tests/check_buffer_collectives cuda0 cuda1`

*Future details*:
1. Include these tests to `make test-debugc` by including them in tests' CMakeLists.txt, look at [these](https://github.com/tsirif/libgpuarray/blob/feature/nccl/tests/CMakeLists.txt#L101) commented `add_test`. The reason I commented them out is that although tests seems to finish, it hangs there without returning pass/fail to CTest. I think this has something to do with MPI.  
2. Define preprocessor flags for including or not GA_HALF -> ncclHalf as it is not supported [always](https://github.com/NVIDIA/nccl/blob/master/src/nccl.h#L115) (mby this exists already in codebase). Also, include tests for half types (although I am not sure how this can be done, as I do not know how to make half type in host - or should half type results be tested against a float reference less than FLOAT_MAX / 2 in absolute?)

tsirif:
For 2 of previous comment: What exists right now suffices as a half type compatibility checker for nccl suffices as it uses the `nccl.h`-defined `CUDA_HAS_HALF` in order to decide if will expose `GA_HALF` as a valid datatype option for collective operations or not. General half type compatibility checking must be dealt for the whole library and in a opaque implementation specific way. Most probably through a check similar to what is done in `nccl.h`:  

```
#include <cuda_runtime.h>

#if CUDART_VERSION >= 7050
#include <cuda_fp16.h>
#define CUDA_HAS_HALF 1
#else
#undef CUDA_HAS_HALF
#endif
```

abergeron:
I guess something like that will have to be done, but we push it back for later when someone (probably me) will have a serious look at float16 support.

Makefile
```diff
@@ -16,27 +16,46 @@ Debug:
 
 Debug/Makefile: Debug config
 ifndef INSTALL_PREFIX
-	(cd Debug && cmake .. -DCMAKE_BUILD_TYPE=Debug)
+	cd Debug && NUM_DEVS=${NUM_DEVS} DEV_NAMES=${DEV_NAMES} cmake .. -DCMAKE_BUILD_TYPE=Debug
```
abergeron: The parenthesis are required to launch the command in a sub-shell that won't affect the working directory of the main one.  Please leave them.

Makefile
```diff
@@ -16,27 +16,46 @@ Debug:
...
 relc: Release/Makefile
-	(cd Release && make && $(SUDO) make install)
+	cd Release && make
+
+test-relc: relc
+ifndef DEVICE
+	cd Release && make test
+else
+	cd Release && DEVICE=${DEVICE} make test
+endif
```
abergeron: You can also run `DEVICE=whatever make` instead of adding a config option.

abergeron:
The changes to the Makefile are weird.  The main way to build and configure libgpuarray is through cmake and that will stay the only documented way.

The Makefile is mostly there for convinience when you work on the code to avoid retyping all the commands.

It used to be that make rel would build and install the library.  Now it only will only build and the install target forces you to run the tests.  This is inconvenient for development.

abergeron:
The changes also break compilation when no previous version is installed (or an older one is) since the python build phase will try to use the system headers.

The rest looks good, so if you fix the Makefile problems, I'm good with merging.

tsirif:
I think that system headers should be replaced with "<...>", so that a test can be run before any installation. Also, I am in favor of building, then testing the build before installing (optionally) and then install. Makefile.conf should be properly set in the developer's convenience, so that commands should enable him to try to build only, build and test only or build and install only without providing with DEVICE=<...> etc every time.

Do you agree with this?

I will place the parentheses around as it is needed and I will change so that test is not necessary for invoking an install command.

tsirif:
Also I want to fix things, so that a `make test` can run successfully the collectives checks too. For some reason, although the tests are successful, `make test` reports a failure on MPI involved tests. Do you know why would this be? When does a check declared with `add_test` report a success??

tsirif:
@abergeron I don't understand whose compilation will break, if there's no previously installed libgpuarray. If you mean "pygpu"'s and then that `make py` would not suceed, wouldn't that happen anyways? That's because building/installing pygpu requires that libgpuarray is installed first.

tsirif:
I would like to make the following changes in other PRs:  
1. Use quoted includes for libgpuarray headers in contrast with angled. [1](https://msdn.microsoft.com/en-us/library/36k2cdd4.aspx) [2](https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html)
2. Make header guards in uniform style and place comments in `#endif` and `#else`
3. Place `src/gpuarray_types.c` and its header in .gitignore and delete them from src (as they are produced by `gen_types.py` anyway). 
4. Delete `include_directories("${CMAKE_CURRENT_SOURCE_DIR}")` from any CMakeLists.txt. This is actually a smell that cmake/includes are not used as intended, because things included with quoted are can be found from 'current source dir'. That would mean that CMakeLists.txt may need to be placed in any subdirectory containing *.c (e.g in `src/cache`) or reorganizing project's files (into include/src subdir in c project - I have done this many times).
5. Reorder include order like [this](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)  
6. Enforce `GA_EXIT_ON_ERROR` on cuda_wait/record, as we discussed.
7. Fix cuda_wait/record, as we discussed.
8. Other things we have discussed to be done and I don't remember now (I will search in comments you made) :stuck_out_tongue: 

These things concern c libgpuarray and not work to be done in pygpu or platoon.

tsirif:
Finally, I have fixed in the latest commit the issue with collective checks and `make test`. Now, commanding `make test-debugc` or `make test-relc` with the appropriate `Makefile.conf` will build and test everything.

What do you think of the current root Makefile structure?

abergeron:
Can you make the `rel`  target depend in `install-relc` and the `debug` target depend on `install-debugc`.

Then the Makefile will be ok.

abergeron:
For your proposed changes:
1. Ok for the quotes inside the C files, however the public headers (under src/gpuarray) should have angle brackets since when installed the are "system" headers.
2. Ok for the header guard style
3. I would prefer to keep the file in the tree since that enables people who don't have python to build the C part.
4. I'm not sure this is a good idea.  The problem is that the "public" headers should be setup to include from the "public" locations and thus should use angle brackets.  So we can't get rid of that completely.
5. Ok for reordering.
6. Ok, but maybe not with a macro.  Also make sure to not leave the context active when you exit.
7. Wasn't that the previous point?  Otherwise you'll have to remind me.
8. Ok, but don't spend too much time hunting for things to do.  Having a working platoon implementation at the end is the objective.  This also goes for the previous points.  If you spend too much time on those the leave them for another time.

Makefile
```diff
@@ -34,6 +34,8 @@ endif
 install-debugc: debugc
 	(cd Debug && ${SUDO} make install)
 
+debug: install-debugc
+
```
abergeron: This target is already on line 10.

Makefile
```diff
@@ -57,5 +59,7 @@ endif
 install-relc: relc
 	(cd Release && ${SUDO} make install)
 
+rel: install-relc
```
abergeron: This target is already on line 1.

tsirif:
@abergeron How is it now?

abergeron:
Perfect.

PR #196, opened 2016-06-16
Fix bug in synchronization logic for multi-gpu.

abergeron:
fix https://github.com/Theano/Theano/issues/4635

### DIFF
--- a/src/gpuarray_buffer_cuda.c
+++ b/src/gpuarray_buffer_cuda.c
@@ -612,13 +612,16 @@ static int cuda_share(gpudata *a, gpudata *b, int *ret) {
 static int cuda_waits(gpudata *a, int flags, CUstream s) {
   ASSERT_BUF(a);
 
-  if (ISSET(a->ctx->flags, GA_CTX_SINGLE_STREAM))
-    return GA_NO_ERROR;
+  /* Never skip the wait if CUDA_WAIT_FORCE */
+  if (ISCLR(flags, CUDA_WAIT_FORCE)) {
+    if (ISSET(a->ctx->flags, GA_CTX_SINGLE_STREAM))
+      return GA_NO_ERROR;
 
-  /* If the last stream to touch this buffer is the same, we don't
-   * need to wait for anything. */
-  if (a->ls == s)
-    return GA_NO_ERROR;
+    /* If the last stream to touch this buffer is the same, we don't
+     * need to wait for anything. */
+    if (a->ls == s)
+      return GA_NO_ERROR;
+  }
 
   cuda_enter(a->ctx);
   if (flags & CUDA_WAIT_READ) {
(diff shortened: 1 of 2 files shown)


src/private_cuda.h
```diff
@@ -107,7 +107,8 @@ GPUARRAY_LOCAL int cuda_record(gpudata *, int);
 /* private flags are in the upper 16 bits */
 #define CUDA_WAIT_READ  0x10000
 #define CUDA_WAIT_WRITE 0x20000
-#define CUDA_WAIT_MASK  0x30000
+#define CUDA_WAIT_FORCE 0x40000
+#define CUDA_WAIT_MASK  0x70000
```
nouiz: CUDA_WAIT_MASK don't seem used, why do you add it?

src/private_cuda.h
```diff
@@ -107,7 +107,8 @@ GPUARRAY_LOCAL int cuda_record(gpudata *, int);
 /* private flags are in the upper 16 bits */
 #define CUDA_WAIT_READ  0x10000
 #define CUDA_WAIT_WRITE 0x20000
-#define CUDA_WAIT_MASK  0x30000
+#define CUDA_WAIT_FORCE 0x40000
```
nouiz: I think you should also define GPUARRAY_CUDA_WAIT_FORCE as GPUARRAY_CUDA_WAIT_READ/WRITE. The last 2 are used in Theano.

src/private_cuda.h
abergeron: I had plans for it at some point.  Maybe it's not required.

src/private_cuda.h
abergeron: I don't where and how FORCE would be used in theano.  This is only required for cross-device synchronization.

nouiz: What if an op would request data inputs from 2 device and do strange thinks
like that? Or write output to 2 defice or so some manual transfer?

abergeron: Then it might need that.  But we don't have any such ops for now and I don't think we will have any in the future.  I don't plan for "what ifs".

src/gpuarray_buffer_cuda.c
```diff
@@ -1343,9 +1347,18 @@ static int cuda_transfer(gpudata *dst, size_t dstoff,
   ASSERT_BUF(src);
   ASSERT_BUF(dst);
 
+  /* The forced synchronization are there because they are required
+     for proper inter-device correctness. */
+
   cuda_enter(dst->ctx);
-  cuda_waits(src, CUDA_WAIT_READ, dst->ctx->mem_s);
+  /* Make sure we have a rev for the source */
+  cuda_record(src, CUDA_WAIT_READ|CUDA_WAIT_FORCE, src->ctx->mem_s);
```
cmyr: when compiling this:

```
libgpuarray/src/gpuarray_buffer_cuda.c: In function ‘cuda_transfer’:
libgpuarray/src/gpuarray_buffer_cuda.c:1355:3: error: too many arguments to function ‘cuda_record’
   cuda_record(src, CUDA_WAIT_READ|CUDA_WAIT_FORCE, src->ctx->mem_s);
   ^
libgpuarray/src/gpuarray_buffer_cuda.c:667:5: note: declared here
 int cuda_record(gpudata *a, int flags) {
```

abergeron: Fixed.

abergeron:
What's up with this?  Can it be merged?  It fixes problems that could be seen in the wild.

PR #198, opened 2016-06-21
allow concatenate and split to take a negative axis

davidweichiang:
numpy's concatenate and split allow axis to be a negative number. Since _concatenate and _split take an unsigned int, it seemed that the best place to make this change was in operations.concatenate and operations.array_split.


abergeron:
What does it do when axis=-30.  Do we need a check for that?

davidweichiang:
It will do the same thing as before...raise an exception because it can't convert a negative number to unsigned int. That's not very informative, so I can add the check if you wnt.

abergeron:
Perfect! Thanks.

PR #199, opened 2016-06-21
Allow reduction kernels to do broadcasting

davidweichiang:
I'm unsure whether you want this, but it seems useful to me to allow a call to a ReductionKernel to take a broadcast=True argument. For example, if a = [[1,2,3]] and b = [[4,5,6], [7,8,9]], in NumPy we can write sum(a*b, axis=-1) to get [32,50], and the attached change allows a reduction kernel to do the same.

Possible issue: maybe broadcasting on the reduction axis is not useful. The attached change allows this, but maybe it should be disallowed.

### DIFF
--- a/pygpu/reduction.py
+++ b/pygpu/reduction.py
@@ -235,12 +235,15 @@ class ReductionKernel(object):
         return self._find_kernel_ls(self._gen_basic, maxls, nd)
 
     def __call__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


davidweichiang:
Thanks!

abergeron:
ReductionKernel will probably go away soon-ish, but if this helps you in the meantime, why not.

PR #201, opened 2016-06-22
Fix reading and writing to mapped pointers in single stream mode.

abergeron:
This fixes a problem with take1() and check_error=1 where the error would sometimes not be reported or reported for a subsequent call in single stream mode.


lamblin:
I'm curious about the speed impact on the benchmarks, but this fix had to be applied.

PR #206, opened 2016-06-28
Add python bindings for collevtives in pygpu

tsirif:
pygpu part of nccl wrapping

### DIFF
--- a/pygpu/collectives.pyx
+++ b/pygpu/collectives.pyx
@@ -10,12 +10,21 @@ from pygpu.gpuarray import GpuArrayException
 
 
 cdef class GpuCommCliqueId:
+    """Represents a unique id shared among :ref:`GpuComm` communicators which
+    participate in a multi-gpu clique.
+
+    Parameters
+    ----------
+    context: :ref:`GpuContext`, optional
+        Reference to which gpu this `GpuCommCliqueId` object belongs.
+    comm_id: bytes, optional
+        Existing unique id to be passed in this object.
+
     """
-    """
-    def __cinit__(self, GpuContext context=None, bytes bytearr=None):
+    def __cinit__(self, GpuContext context=None, bytes comm_id=None):
         self.context = ensure_context(context)
-        if bytearr is not None:
-            self.set_comm_id(bytearr)
+        if comm_id is not None:
+            self.comm_id = comm_id
         else:
             comm_generate_id(self.context.ctx, self)
 
@@ -29,7 +38,7 @@ cdef class GpuCommCliqueId:
         raise RuntimeError, "Cannot pickle %s object" % self.__class__.__name__
 
     property comm_id:
-        "Unique clique id to be used by each GpuComm in a group of devices"
+        "Unique clique id to be used by each :ref:`GpuComm` in a group of devices"
         def __get__(self):
             cdef bytes res
             res = self.comm_id.internal[:GA_COMM_ID_BYTES]
@@ -44,19 +53,32 @@ cdef class GpuCommCliqueId:
 
 
 cdef class GpuComm:
-    """
+    """Represents a communicator which participates in a multi-gpu clique.
+
+    It is used to invoke collective operations to gpus inside its clique.
+
+    Parameters
+    ----------
+    cid: :ref:`GpuCommCliqueId`
+        Unique id shared among participating communicators.
+    ndev: int
+        Number of communicators inside the clique.
+    rank: int
+        User-defined rank of this communicator inside the clique. It influences
+        order of collective operations.
+
     """
     def __dealloc__(self):
         gpucomm_free(self.c)
-
-    def __reduce__(self):
-        raise RuntimeError, "Cannot pickle %s object" % self.__class__.__name__
 
     def __cinit__(self, GpuCommCliqueId cid not None, int ndev, int rank):
         cdef int err
         err = gpucomm_new(&self.c, cid.context.ctx, cid.comm_id, ndev, rank)
         if err != GA_NO_ERROR:
             raise get_exc(err), gpucontext_error(cid.context.ctx, err)
+
+    def __reduce__(self):
+        raise RuntimeError, "Cannot pickle %s object" % self.__class__.__name__
 
     property count:
         "Total number of communicators inside the clique"
@@ -73,6 +95,27 @@ cdef class GpuComm:
             return gpurank
 
     def reduce(self, GpuArray src not None, op, GpuArray dest=None, int root=-1):
+        """Reduce collective operation for ranks in a communicator world.
+
+        Parameters
+        ----------
+        src: :ref:`GpuArray`
+            Array to be reduced.
+        op: string
+            Key indicating operation type.
+        dest: :ref:`GpuArray`, optional
+            Array to collecti reduce operation result.
+        root: int
+            Rank in `GpuComm` which will collect result.
+
+        Notes
+        -----
+        * `root` is necessary when invoking from a non-root rank. Root caller
+        needs not to provide `root` argument.
+        * Not providing `dest` argument for a root caller will result in creating
+        a new compatible :ref:`GpuArray` and returning result in it.
+
+        """
         if dest is None:
             if root != -1:
                 return comm_reduce_from(self, src, to_reduce_opcode(op), root)
@@ -83,22 +126,90 @@ cdef class GpuComm:
         return comm_reduce(self, src, dest, to_reduce_opcode(op), root)
 
     def all_reduce(self, GpuArray src not None, op, GpuArray dest=None):
+        """AllReduce collective operation for ranks in a communicator world.
+
+        Parameters
+        ----------
+        src: :ref:`GpuArray`
+            Array to be reduced.
+        op: string
+            Key indicating operation type.
+        dest: :ref:`GpuArray`, optional
+            Array to collect reduce operation result.
+
+        Notes
+        -----
+        * Not providing `dest` argument for a root caller will result in creating
+        a new compatible :ref:`GpuArray` and returning result in it.
+
+        """
         if dest is None:
             return pygpu_make_all_reduced(self, src, to_reduce_opcode(op))
         return comm_all_reduce(self, src, dest, to_reduce_opcode(op))
 
     def reduce_scatter(self, GpuArray src not None, op, GpuArray dest=None):
+        """ReduceScatter collective operation for ranks in a communicator world.
+
+        Parameters
+        ----------
+        src: :ref:`GpuArray`
+            Array to be reduced.
+        op: string
+            Key indicating operation type.
+        dest: :ref:`GpuArray`, optional
+            Array to collect reduce operation scattered result.
+
+        Notes
+        -----
+        * Not providing `dest` argument for a root caller will result in creating
+        a new compatible :ref:`GpuArray` and returning result in it.
+
+        """
         if dest is None:
             return pygpu_make_reduce_scattered(self, src, to_reduce_opcode(op))
         return comm_reduce_scatter(self, src, dest, to_reduce_opcode(op))
 
     def broadcast(self, GpuArray array not None, int root=-1):
+        """Broadcast collective operation for ranks in a communicator world.
+
+        Parameters
+        ----------
+        array: :ref:`GpuArray`
+            Array to be reduced.
+        root: int
+            Rank in `GpuComm` which broadcasts its `array`.
+
+        Notes
+        -----
+        * `root` is necessary when invoking from a non-root rank. Root caller
+        does not need to provide `root` argument.
+
+        """
         if root == -1:
             comm_get_rank(self, &root)
         return comm_broadcast(self, array, root)
 
     def all_gather(self, GpuArray src not None, GpuArray dest=None,
                    unsigned int nd_up=1):
+        """AllGather collective operation for ranks in a communicator world.
+
+        Parameters
+        ----------
+        src: :ref:`GpuArray`
+            Array to be gathered.
+        dest: :ref:`GpuArray`, optional
+            Array to receive all gathered arrays from ranks in `GpuComm`.
+        nd_up: unsigned int
+            Used when creating result array. Indicates how many extra dimensions
+            user wants result to have. Default is 1, which means that the result
+            will store each rank's gathered array in one extra new dimension.
+
+        Notes
+        -----
+        * Providing `nd_up` == 0 means that gathered arrays will be appended to
+        the dimension with the largest stride.
+
+        """
         if dest is None:
             return pygpu_make_all_gathered(self, src, nd_up)
         return comm_all_gather(self, src, dest)
--- a/pygpu/collectives.pyx
+++ b/pygpu/collectives.pyx
@@ -66,6 +66,7 @@ cdef class GpuCommCliqueId:
     def __releasebuffer__(self, Py_buffer* buffer):
         if buffer == NULL:
             raise BufferError, "NULL buffer view in releasebuffer"
+
         if buffer.shape != NULL:
             free(buffer.shape)
         Py_DECREF(self)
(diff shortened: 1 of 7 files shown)


pygpu/collectives.pyx
```diff
@@ -0,0 +1,299 @@
...
+    """
+    """
+    def __cinit__(self, GpuContext context=None):
+        self.context = ensure_context(context)
+
+    property comm_id:
+        "Unique clique id to be used by each GpuComm in a group of devices"
+        def __get__(self):
+            cdef bytes res
+            comm_generate_id(self.context.ctx, self)
```
abergeron: You generate a new clique_id each time the property is accessed.  I don't think that is efficient or wise.

You should probably generate it in the constructor and maybe have an optional parameter so that the user can feed in a bytes string to use as the id (like what you do in `__set__`).

pygpu/collectives.pyx
```diff
@@ -0,0 +1,299 @@
...
+from pygpu.gpuarray import GpuArrayException
+
+
+cdef class GpuCommCliqueId:
+    """
+    """
+    def __cinit__(self, GpuContext context=None):
+        self.context = ensure_context(context)
+
+    property comm_id:
```
abergeron: It would probably be a good idea to make that class comparable and hashable on the basis of the content of the struct.

You can also make it pickleable with the byte string I guess.  Make sure that works though and if it doesn't then forbid pickling.

pygpu/collectives.pyx
```diff
@@ -0,0 +1,299 @@
...
+
+    def get_count(self):
+        cdef int gpucount
+        comm_get_count(self, &gpucount)
+        return gpucount
+
+    def get_rank(self):
+        cdef int gpurank
+        comm_get_rank(self, &gpurank)
+        return gpurank
```
abergeron: These might be more python-like as properties.

pygpu/collectives.pyx
```diff
@@ -0,0 +1,299 @@
...
+
+    def reduce(self, GpuArray src not None, op, GpuArray dest=None, int root=-1):
+        if dest is None:
+            if root != -1:
+                return comm_reduce_from(self, src, to_reduce_opcode(op), root)
+            else:
+                return pygpu_make_reduced(self, src, to_reduce_opcode(op))
+        if root == -1:
+            comm_get_rank(self, &root)
```
abergeron: Would it be possible to always gather the rank from the comm and not take the root argument?  This way you can also do argument checking to make sure that the user passed in a `dest` when required

pygpu/collectives.pyx
```diff
@@ -0,0 +1,299 @@
...
+    "product": GA_PROD,
+    "max": GA_MAX,
+    "maximum": GA_MAX,
+    "min": GA_MIN,
+    "minimum": GA_MIN,
+    }
+
+cdef int to_reduce_opcode(op) except -1:
+    if isinstance(op, int):
+        return op
```
abergeron: The integer values are never exposed anywhere so if the user uses some they are probably wrong.

Either expose the `GA_...`  values or remove this part.

pygpu/collectives.pyx
```diff
@@ -0,0 +1,299 @@
...
+    """
+    """
+    def __dealloc__(self):
+        gpucomm_free(self.c)
+
+    def __reduce__(self):
+        raise RuntimeError, "Cannot pickle GpuComm object"
+
+    def __cinit__(self, GpuCommCliqueId cid not None, int ndev, int rank):
+        self.context = cid.context
```
abergeron: Since there is a gpucomm_context operation, we don't need to store another copy of the context here.

abergeron:
I have some small superficial comments, but overall it's looking good.

pygpu/collectives.pyx
tsirif: Nccl generates each time a unique id using process id. Will it be useful to pickle a GpuCommCliqueId object in general?

pygpu/collectives.pyx
tsirif: A non-root participant in reduce collective operation is required to pass `root`'s rank as an argument to indicate where the `src` array will be collected.

tsirif: I am writing documentation to clarify usage

pygpu/collectives.pyx
```diff
@@ -23,12 +23,14 @@ cdef class GpuCommCliqueId:
             res = self.comm_id.internal[:GA_COMM_ID_BYTES]
             return res
 
-        def __set__(self, unsigned char[:] bytearr):
+        def __set__(self, bytearr):
             cdef int length
-            length = bytearr.shape[0]
+            length = len(bytearr)
             if length < GA_COMM_ID_BYTES:
                 raise ValueError, "GpuComm clique id must have length %d bytes" % (GA_COMM_ID_BYTES)
```
abergeron: Check the length after casting to bytes otherwise it can still be too big after.

pygpu/collectives.pyx
```diff
@@ -23,12 +23,14 @@ cdef class GpuCommCliqueId:
             res = self.comm_id.internal[:GA_COMM_ID_BYTES]
             return res
 
-        def __set__(self, unsigned char[:] bytearr):
+        def __set__(self, bytearr):
```
abergeron: It would be ok to require a bytes array on input here.

tsirif: I updated these in the latest commits :)

pygpu/collectives.pyx
abergeron: Ok but since you get get the rank from the comm object you don't have to require that the user passes it in.  This is the sort of "nice interface" things that we usually do for python bindings.

pygpu/collectives.pyx
```diff
@@ -0,0 +1,312 @@
...
+        def __get__(self):
+            return self.comm_id_py
+
+        def __set__(self, bytes bytearr):
+            cdef int length
+            length = len(bytearr)
+            if length < GA_COMM_ID_BYTES:
+                raise ValueError, "GpuComm clique id must have length %d bytes" % (GA_COMM_ID_BYTES)
+            self.comm_id_py = bytearr
+            self.comm_id.internal = <char*>self.comm_id_py
```
abergeron: Does this work?  It seems to me to be doing pointer assignment which should not work.

I may be wrong about this.

pygpu/collectives.pyx
abergeron: Probably not so what you did is ok.

pygpu/collectives.pyx
```diff
@@ -0,0 +1,312 @@
...
+from pygpu.gpuarray import GpuArrayException
+
+
+cdef class GpuCommCliqueId:
+    """
+    """
+    def __cinit__(self, GpuContext context=None, bytes bytearr=None):
+        self.context = ensure_context(context)
+        if bytearr is not None:
+            self.set_comm_id(bytearr)
```
abergeron: Where is `set_comm_id` defined?

tsirif: Mea culpa. I forgot about accessors in python :P

pygpu/collectives.pyx
tsirif: I am not sure, I will check generated code and ensure if it works with tests.
I intend to write code in a way that it will not copy the `bytes bytearr` into a new array, but it will keep a reference and make `self.comm_id.internal` point correctly.

Also, do you think that making `__get__` and `__set__` refer to the `GpuCommCliqueId` object itself would be a good idea?

pygpu/collectives.pyx
tsirif: A non-root caller cannot get the rank of the root which the user desires. `comm_get_rank` will return the rank of the `GpuComm` object in reference to the clique. Reduce operation need user to specify which rank in the clique will act as the root. So, in general it has to be passed explicitly. Not passing it means that the caller itself is the root rank. I am not sure if this API logic is good though. It may have to be reworked. What do you think?

pygpu/collectives.pyx
abergeron: I don't think you can make .internal "point correctly" since it's not a pointer, but a char array.  So you have to copy the data.

What do you mean by "refer to the GpuCommCliqueId object itself"?

pygpu/collectives.pyx
```diff
@@ -10,12 +10,21 @@ from pygpu.gpuarray import GpuArrayException
...
+
     """
-    """
-    def __cinit__(self, GpuContext context=None, bytes bytearr=None):
+    def __cinit__(self, GpuContext context=None, bytes comm_id=None):
         self.context = ensure_context(context)
-        if bytearr is not None:
-            self.set_comm_id(bytearr)
+        if comm_id is not None:
+            self.comm_id = comm_id
```
abergeron: `comm_id` is a bytes object.  `self.comm_id` is a `gpucommCliqueId`.  This won't work.

tsirif: I am confused, won't this trigger the setter?

pygpu/collectives.pyx
abergeron: Oh so the root can be any rank and the user choose?  In that case, yes you need to take it as input.

pygpu/collectives.pyx
abergeron: I'm not sure since you have an attribute and a property for the same name.  In any case you shouldn't make python calls with self in `__cinit__` since the object is not valid yet.

pygpu/collectives.pyx
```diff
@@ -23,10 +23,13 @@ cdef class GpuCommCliqueId:
     """
     def __cinit__(self, GpuContext context=None, bytes comm_id=None):
         self.context = ensure_context(context)
+        if comm_id is None:
+            comm_generate_id(self.context.ctx, &self.comm_id)
+
+    def __init__(self, GpuContext context=None, bytes comm_id=None):
         if comm_id is not None:
             self.comm_id = comm_id
```
abergeron: You still have the problem of duplicated names.

tsirif:
@abergeron How is it now? In Cython I cannot declare `__eq__` as I found while cythonizing, so I implemented a `__richcmp__`. I will write tests tomorrow :)

pygpu/collectives.pyx
```diff
@@ -0,0 +1,432 @@
...
+    def __init__(self, GpuContext context=None, bytes comm_id=None):
+        if comm_id is not None:
+            self.comm_id = comm_id
+
+    def __richcmp__(this, that, int op):
+        if op != 2 and op != 3:
+            raise RuntimeError, "Compare for equal or not equal only."
+        cdef bint res
+        res = type(this) == type(that) and \
+              strncmp(this.c_comm_id.internal, that.c_comm_id.internal, GA_COMM_ID_BYTES) == 0
```
abergeron: Shouldn't that me a memcmp?  There could be zero bytes in the middle of the id.

abergeron:
After the tests this should be ok to go in.

pygpu/tests/test_collectives.py
```diff
@@ -0,0 +1,91 @@
...
+        other_cid1 = GpuCommCliqueId(context=ctx, comm_id=self.cid.comm_id)
+        other_cid2 = GpuCommCliqueId(context=ctx)
+        assert self.cid == other_cid1
+        assert self.cid != other_cid2
+        assert other_cid2 > other_cid1
+        assert other_cid2 >= other_cid1
+        assert self.cid >= other_cid1
+        assert other_cid1 < other_cid2
+        assert other_cid1 <= other_cid2
+        assert other_cid1 <= self.cid
```
abergeron: How can you know for sure the order of the clique IDs?  Is it guaranteed that they compare in the order they were generated?

abergeron:
For the tests you don't have to make the same comparison with MPI as in C.  The only thing we are testing here is that the bindings are ok.  We can assume that the underlying C functions work correctly.

pygpu/tests/test_collectives.py
tsirif: So far, for NCCL yes. It uses internally a static counter which is increased for every call and is sufficed to string id containing process id. If something changes, I will set properly bytestrings.

tsirif:
For `GpuComm`, I am writing one test for each programming path in the `collectives.pyx`. I intend though to do some functional testing because `pygpu_*` introduce functionality (creating new `GpuArray` to be returned).

pygpu/tests/test_collectives.py
abergeron: I don't like counting on that.  Perhaps you could generate some fixed IDs for comparison.  Otherwise add a comment that says NCCL currently returns increasing IDs otherwise this test is way too confusing.

pygpu/collectives.pyx
```diff
@@ -33,6 +36,40 @@ cdef class GpuCommCliqueId:
...
+        buffer.suboffsets = NULL
+        buffer.internal = NULL
+        Py_INCREF(self)
+
+    def __releasebuffer__(self, Py_buffer* buffer):
+        if buffer == NULL:
+            raise BufferError, "NULL buffer view in releasebuffer"
+        if buffer.shape != NULL:
+            free(buffer.shape)
+        Py_DECREF(self)
```
abergeron: Does this really add something that we want or need?  I don't think that there will be people doing a lot of manipulation on the internal of the clique ID.

And if that is not useful I would remove it since it just makes for more code to maintain.

pygpu/collectives.pyx
```diff
@@ -248,7 +256,8 @@ cdef class GpuComm:
         """
         if root == -1:
             comm_get_rank(self, &root)
-        return comm_broadcast(self, array, root)
+        comm_broadcast(self, array, root)
+        return
```
abergeron: This is implicit in python, so no need to add it explicitly.

pygpu/collectives.pyx
```diff
@@ -228,9 +235,10 @@ cdef class GpuComm:
         """
         if dest is None:
             return pygpu_make_reduce_scattered(self, src, to_reduce_opcode(op))
-        return comm_reduce_scatter(self, src, dest, to_reduce_opcode(op))
+        comm_reduce_scatter(self, src, dest, to_reduce_opcode(op))
+        return
 
-    def broadcast(self, GpuArray array not None, int root=-1):
+    def bcast(self, GpuArray array not None, int root=-1):
```
abergeron: I'm not sure the shortened name is better.

tsirif: Ok I will return it to the full.

pygpu/collectives.pyx
tsirif: Ok!

pygpu/collectives.pyx
tsirif: Yes you are right I was thinking about this. I want to expose a buffer
interface so that we can avoid the cost of creating and copying bytestrings
multiple times. Certainly the cost of this isn't that great, but I felt
better avoiding this should we had a way to do this.

I suppose that the common id will be shared among processes through mpi. In
platoon we can use mpi4py which supports objects that expose a buffer
interface as inputs. This will have zero copies other than those that mpi
internally does.

If this benefit is considered less than the cost of having more complex
code, then this will be removed. What do you say??

abergeron: If it helps multi-process sharing then, sure it might be useful so keep it in.

abergeron:
What is happening with this?

tsirif:
@abergeron I think code is ready to be merged. I had a problem with GIL that I did not understand but it seems that it was simple to solve... NCCL does internally some stuff with threads and python crashed with segm fault. I think that placing nogil where needed solved this for now. But we need to be cautious because internals of NCCL may change and therefore we may be need to add nogil to other function calls too.

Please review code and let me know.

nouiz:
if you aren't sure about the gil, we should ask nvidia. The most direct way
would be to ask on the nccl as an issue. Can you do that and report here
the link to that issue?

tsirif:
I will do that. Btw I realized that it was my fault the segm fault and with the corrected code it passes the tests without invoking nogil. But still, I am not sure so I will ask. This is because NCCL does use some threaded things inside. For example, a cpu atomic operation when generating the unique clique id...

abergeron:
NCCL using threaded things does not matter for the GIL.  It only matters if you want to use python objects (which the internals of NCCL do not).

Apart from that I think the code is fine.

PR #207, opened 2016-07-03
fix double free in GpuArray_concatenate

davidweichiang:
The double free only happens when recovering from an exception.

BTW -- it appears that you can't concatenate two arrays with different contexts even if the contexts differ only in their streams. Is this by design?

Thanks!


src/gpuarray_array.c
```diff
@@ -1026,6 +1026,7 @@ int GpuArray_concatenate(GpuArray *r, const GpuArray **as, size_t n,
 
   return GA_NO_ERROR;
  fail:
+  r->dimensions = res_dims;
```
abergeron: Good analysis, but I would put the reset just before the goto instead of here.

nouiz:
thanks.

PR #210, opened 2016-07-06
round allocation size to power of two times BLOCK_SIZE

davidweichiang:
If a block of size (say) 10M is allocated, then a block of 11M, then the previous 10M block can't be used; this change rounds up to a power of two times BLOCK_SIZE to alleviate this problem a bit.

### DIFF
--- a/src/gpuarray_buffer_cuda.c
+++ b/src/gpuarray_buffer_cuda.c
@@ -400,8 +400,12 @@ static int allocate(cuda_context *ctx, gpudata **res, gpudata **prev,
   gpudata *next;
   *prev = NULL;
 
-  if (!(ctx->flags & GA_CTX_DISABLE_ALLOCATION_CACHE))
-    if (size < BLOCK_SIZE) size = BLOCK_SIZE;
+  if (!(ctx->flags & GA_CTX_DISABLE_ALLOCATION_CACHE)) {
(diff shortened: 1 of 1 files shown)


abergeron:
What you say is true, but I'm not sure we want to use only powers of two since it could lead to wasting about half of the memory in the worst case.

If you made this use a multiple of the block_size `(size + BLOCK_SIZE - 1) / BLOCK_SIZE` instead of a power then it should be better.  It can still have some problem cases, but these can be fixed by a large preallocation.

abergeron:
I just realized this this before the splitting so it's ok to allocate a lot more since the rest can be used later.

I'm still not convinced that it is strictly better than the previous behaviour.  It fixes some corner cases and introduces others.

davidweichiang:
Sorry, I just noticed that this was never closed. Could you explain what the new problems introduced would be?

abergeron:
It fixes the problem of allocating 10M, 11M, 12M, ..., but introduces a problem if you allocate 8M+1byte repeatedly where you would essentially allocate twice the memory you should.

It's a tradeoff in both cases and I'm not sure which one is better.  Both have the same workaround of pre-allocating a big block and releasing it which would reserve a large area of gpu memory that can be split up.

nouiz:
Note, in Theano, there is the flag gpuarray.preallocate that allow the user to do the big preallocation  that was discussed.

PR #215, opened 2016-07-12
Fix cmake not detecting sgemmEx.

abergeron:
This was broken in #193

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -111,9 +111,10 @@ if (CUDA_FOUND)
   add_definitions(-DWITH_CUDA)
   include_directories(${CUDADRV_INCLUDE})
 
-  set(BUILD_WITH_BLAS 1 PARENT_SCOPE)
   list(APPEND _GPUARRAY_SRC gpuarray_blas_cuda_cublas.c)
   add_definitions(-DWITH_CUDA_CUBLAS)
+
+  set(CMAKE_REQUIRED_LIBRARIES ${CUDA_CUBLAS_LIBRARIES})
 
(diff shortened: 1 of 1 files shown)


nouiz:
It still seem to work here. Should we add a test to prevent braking this?

abergeron:
It only breaks for new installs/configs.  Also you can't really test this since it's part of the compile configuration process.

We can however dump the test and assume it's there once we stop supporting 6.5.

nouiz:
If the only time the test should fail is with nvcc 6.5, why not just
implement it and let it fail on nvcc 6.5?

abergeron:
What test?  Implement what?

nouiz:
Why not have a test that call sgemmEx?

nouiz:
I'm merging now as this fix a crash, bit I think a quick smoke test would be good, but it isn't urgent.

abergeron:
The problem is that without sgemmEx, there is no hgemm implementation.  So then the hgemm call returns "not supported" and this makes the theano op fail.

There could always be a float16 blas test to make sure it works, but since the call is only available starting with 7.0 I view this as optional functionality so it not being present should not be a reason for tests to fail.

nouiz:
Should we continue to support cuda 6.5? It was released ni august 2016. It
is missing in driver compilation, nccl isn't supported and now sgemmEx.

What do you think of dropping 6.5?

abergeron:
I would very much like to drop it, but I am waiting for the official release of 8.0 for that (which should be soon, in August).  I try to support at least two releases back to give some time to people to upgrade.

nouiz:
seem reasonable.

PR #217, opened 2016-07-13
Add a new preamble macro GA_WARP_SIZE

huanzhang12:
The CUDA runtime constant warpSize is used in a few Theano GPU kernels. This patch defines a new preamble macro, GA_WARP_SIZE, to make kernel code portable between CUDA and OpenCL. In CUDA, it is simply replaced by warpSize; in OpenCL, no such runtime constant is provided; the closet alternative is the kernel property CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE. Technically there are some subtle differences between warpSize and CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE, however on GPUs they both reflect the hardware SIMD size. 

Since it depends on the GPU device, the definition of GA_WARP_SIZE in OpenCL preamble has to be created dynamically. I added necessary code to cl_check_extensions() that creates a dummy kernel, gets the GA_KERNEL_PROP_PREFLSIZE property and updates the preamble. We also need to be aware that warpSize is a runtime constant and the defined macro for OpenCL is a compile-time constant, but there will be no difference if we don't reuse _compiled_ kernel binaries on different devices.

### DIFF
--- a/src/gpuarray_buffer_opencl.c
+++ b/src/gpuarray_buffer_opencl.c
@@ -37,54 +37,7 @@ static gpukernel *cl_newkernel(gpucontext *ctx, unsigned int count,
                                const char *fname, unsigned int argcount,
                                const int *types, int flags, int *ret,
                                char **err_str);
-
-static const char CL_PREAMBLE[] =
-  "#define local_barrier() barrier(CLK_LOCAL_MEM_FENCE)\n"
-  "#define WITHIN_KERNEL /* empty */\n"
-  "#define KERNEL __kernel\n"
-  "#define GLOBAL_MEM __global\n"
-  "#define LOCAL_MEM __local\n"
-  "#define LOCAL_MEM_ARG __local\n"
-  "#define REQD_WG_SIZE(x, y, z) __attribute__((reqd_work_group_size(x, y, z)))\n"
-  "#ifndef NULL\n"
-  "  #define NULL ((void*)0)\n"
-  "#endif\n"
-  "#define LID_0 get_local_id(0)\n"
-  "#define LID_1 get_local_id(1)\n"
-  "#define LID_2 get_local_id(2)\n"
-  "#define LDIM_0 get_local_size(0)\n"
-  "#define LDIM_1 get_local_size(1)\n"
-  "#define LDIM_2 get_local_size(2)\n"
-  "#define GID_0 get_group_id(0)\n"
-  "#define GID_1 get_group_id(1)\n"
-  "#define GID_2 get_group_id(2)\n"
-  "#define GDIM_0 get_num_groups(0)\n"
-  "#define GDIM_1 get_num_groups(1)\n"
-  "#define GDIM_2 get_num_groups(2)\n"
-  "#define ga_bool uchar\n"
-  "#define ga_byte char\n"
-  "#define ga_ubyte uchar\n"
-  "#define ga_short short\n"
-  "#define ga_ushort ushort\n"
-  "#define ga_int int\n"
-  "#define ga_uint uint\n"
-  "#define ga_long long\n"
-  "#define ga_ulong ulong\n"
-  "#define ga_float float\n"
-  "#define ga_double double\n"
-  "#define ga_half half\n"
-  "#define ga_size ulong\n"
-  "#define ga_ssize long\n"
-  "#define load_half(p) vload_half(0, p)\n"
-  "#define store_half(p, v) vstore_half_rtn(v, 0, p)\n"
-  "#define GA_DECL_SHARED_PARAM(type, name) , __local type name[]\n"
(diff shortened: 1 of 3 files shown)


huanzhang12:
Thank you for your feedback. So how about moving the GA_WARP_SIZE check into cl_init()? Also, it seems we need a per-context array for storing the preamble (since each device has a different GA_WARP_SIZE)? If it is OK I can make the changes.

abergeron:
We can have a field for just the GA_WARP_SIZE macro in the context.  No need to duplicate the whole thing.

huanzhang12:
OK, I am working on it :)

huanzhang12:
Please review the new commit and hopefully I didn't break anything this time. Thanks!

src/gpuarray_buffer_opencl.c
```diff
@@ -113,6 +172,27 @@ cl_ctx *cl_make_ctx(cl_context ctx, int flags) {
...
+  /* Create per-context OpenCL preamble */
+
+  // Create a dummy kernel and check GA_KERNEL_PROP_PREFLSIZE
+  rlk[0] = dummy_kern;
+  len = sizeof(dummy_kern);
+  // this dummy kernel does not require a CLUDA preamble
+  m = cl_newkernel((gpucontext *)res, 1, rlk, &len, "kdummy", 0, NULL, 0, &ret, NULL);
+  assert((m != NULL) && "Cannot create a dummy OpenCL kernel!");
+  ret = cl_property((gpucontext *)res, NULL, m, GA_KERNEL_PROP_PREFLSIZE, &warp_size);
+  if (ret != GA_NO_ERROR) warp_size = 128;
```
abergeron: If there is an error assume the warp size is 64, not 128.

src/gpuarray_buffer_opencl.c
```diff
@@ -113,6 +172,27 @@ cl_ctx *cl_make_ctx(cl_context ctx, int flags) {
...
+
+  // Write the preferred workgroup multiple as GA_WARP_SIZE in preamble
+  len = snprintf(NULL, 0, CL_CONTEXT_PREAMBLE, warp_size);
```
abergeron: This will not work on windows.  Please use strb_appendf() which handles this correctly on all platforms.

src/gpuarray_buffer_opencl.c
```diff
@@ -32,6 +32,59 @@ static gpudata *cl_alloc(gpucontext *c, size_t size, void *data, int flags,
...
+  "#define ga_ulong ulong\n"
+  "#define ga_float float\n"
+  "#define ga_double double\n"
+  "#define ga_half half\n"
+  "#define ga_size ulong\n"
+  "#define ga_ssize long\n"
+  "#define load_half(p) vload_half(0, p)\n"
+  "#define store_half(p, v) vstore_half_rtn(v, 0, p)\n"
+  "#define GA_DECL_SHARED_PARAM(type, name) , __local type name[]\n"
+  "#define GA_DECL_SHARED_BODY(type, name)\n";
```
abergeron: Why do you need to move this?

src/private_opencl.h
```diff
@@ -40,6 +40,7 @@ typedef struct _cl_ctx {
   cl_command_queue q;
   char *exts;
   cl_int err;
+  char *preamble;
```
abergeron: Move that above the err field to avoid padding.

src/gpuarray_buffer_opencl.c
```diff
@@ -113,6 +172,27 @@ cl_ctx *cl_make_ctx(cl_context ctx, int flags) {
...
   res->refcnt--; /* Prevent ref loop */
+
+  /* Create per-context OpenCL preamble */
+
+  // Create a dummy kernel and check GA_KERNEL_PROP_PREFLSIZE
+  rlk[0] = dummy_kern;
+  len = sizeof(dummy_kern);
+  // this dummy kernel does not require a CLUDA preamble
+  m = cl_newkernel((gpucontext *)res, 1, rlk, &len, "kdummy", 0, NULL, 0, &ret, NULL);
+  assert((m != NULL) && "Cannot create a dummy OpenCL kernel!");
```
abergeron: Don't use assert as error checking since the assert statements are removed in release mode.

src/gpuarray_buffer_opencl.c
nouiz: I think it would be better to return an error if we can't determine the
warp size. Normally, we should always have that information and if we use
the wrong value, we will probably return bad result. So an error here is
better I think.

huanzhang12:
Thanks for all the comments. I will work on address these issues.

huanzhang12:
Please view the last commit. I hope it look good now.

PR #221, opened 2016-07-15
enable to compile in under C99 compiler

NeutralCode:
In windows, python 3.4 compiled Visual Studio 2010 and it don't support C99.

This PR enable to compile libgpuarray in VS2015.


nouiz:
This seem good, so I'll merge.

Do you know if it is possible to pass a parameter to gcc on linux to assert that we don't break that?

NeutralCode:
'-Wdeclaration-after-statement' option warns that situation.

>-Wdeclaration-after-statement (C only)
    Warn when a declaration is found after a statement in a block. This construct, known from C++, was introduced with ISO C99 and is by default allowed in GCC. It is not supported by ISO C90 and was not supported by GCC versions before GCC 3.0. See Mixed Declarations.

nouiz:
-Wdeclaration-after-statement

PR #223, opened 2016-07-16
Add checks to cuda_exit for cuda_waits and records

tsirif:
Continue #222

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,7 +1,9 @@
 include(CheckFunctionExists)
 
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -DDEBUG")
-add_definitions(-Wdeclaration-after-statement)
+if(CMAKE_COMPILER_IS_GNUCC)
+  add_definitions(-Wdeclaration-after-statement)
+endif()
 
 find_package(CUDA)
 find_package(OpenCL)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,6 +1,7 @@
 include(CheckFunctionExists)
 
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -DDEBUG")
+add_definitions(-Wdeclaration-after-statement)
 
 find_package(CUDA)
 find_package(OpenCL)
--- a/src/private_cuda.h
+++ b/src/private_cuda.h
@@ -55,13 +55,13 @@
     }                                   \
   } while (0)
 
-#define CUDA_EXIT_ON_ERROR(a, cmd)       \
-  do {                                   \
-    (a)->ctx->err = (cmd);               \
-    if ((a)->ctx->err != CUDA_SUCCESS) { \
-      cuda_exit((a)->ctx);               \
-      return GA_IMPL_ERROR;              \
-    }                                    \
+#define CUDA_EXIT_ON_ERROR(ctx, cmd)  \
+  do {                                \
+    (ctx)->err = (cmd);               \
+    if ((ctx)->err != CUDA_SUCCESS) { \
+      cuda_exit((ctx));               \
+      return GA_IMPL_ERROR;           \
+    }                                 \
   } while (0)
 
 typedef struct _cuda_context {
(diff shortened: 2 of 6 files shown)


src/private_cuda.h
```diff
@@ -46,6 +46,24 @@
 /* Keep in sync with the copy in gpuarray/extension.h */
 #define DONTFREE 0x10000000
 
+#define GA_CUDA_EXIT_ON_ERROR(ctx, cmd) \
+  do {                                  \
+    int err = (cmd);                    \
```
abergeron: Does that compile with -Wdeclaration-after-statement without warnings?

tsirif: Ok I am fixing this. Compiling with "-Wdeclaration-after-statement" gives also:

```
/home/tsirif/Development/libgpuarray/src/gpuarray_blas_cuda_cublas.c: In function ‘dgemmBatch’:
/home/tsirif/Development/libgpuarray/src/gpuarray_blas_cuda_cublas.c:686:3: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
   const size_t threshold = 650;
```

What should be done with const declarations?

abergeron: Leave it for now.  I will get around to testing on windows and this will be part of the things I'll have to fix.

src/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 include(CheckFunctionExists)
 
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -DDEBUG")
+add_definitions(-Wdeclaration-after-statement)
```
abergeron: Wrap this if a condition that checks for gcc: `if (CMAKE_COMPILER_IS_GNUCC)`.

src/gpuarray_array_collectives.c
```diff
@@ -10,7 +10,8 @@
  */
 static inline size_t find_total_elems(const GpuArray* array) {
   unsigned int i;
-  size_t total_elems = 1;
+  size_t total_elems;
+  total_elems = 1;
```
abergeron: Was this really triggering the warning?

tsirif: I don't remember. Probably not, but I remembered what you said about mvsc
and assigning when declaring. Except if this only happens in 'for' clauses.
Then it's unnecessary :p

abergeron: Yes for `for` you have to declare the iteration variable beforehand.

As for initializing on declaration, it's not a problem so we should keep doing it.

abergeron: Can you remove the changes that place the initialization on a different line?

abergeron:
I would be ok with this if you revert changes like this:
```
-  int rank = 0;
+  int rank;
+  rank = 0;
```

tsirif:
Ok I reverted all of them. Something to note: When I run the tests in current 'master' branch, three of them fail. All three of them are "test_xsplit". For example:  
```
FAIL: pygpu.tests.test_operations.test_xsplit('d', (2, 2, 4), [3, 6])
  File "/usr/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/tsirif/Development/libgpuarray/pygpu/tests/test_operations.py", line 44, in xsplit
    numpy.testing.assert_allclose(pc, numpy.asarray(pg))
  File "/usr/lib/python3.5/site-packages/numpy/testing/utils.py", line 1359, in assert_allclose
    verbose=verbose, header=header)
  File "/usr/lib/python3.5/site-packages/numpy/testing/utils.py", line 668, in assert_array_compare
    raise AssertionError(msg)
AssertionError:
Not equal to tolerance rtol=1e-07, atol=0

(shapes (0,), (2, 2, 0) mismatch)
 x: array([], dtype=float32)
 y: array([], shape=(2, 2, 0), dtype=float32)
```

The other are:
`FAIL: pygpu.tests.test_operations.test_xsplit('v', (4, 4), [3, 6])` and
`FAIL: pygpu.tests.test_operations.test_xsplit('h', (4, 4), [3, 6])`

abergeron:
The behaviour of the split operations has changed to match what is in numpy 1.11.  If you have an earlier version these failures are to be expected.  Maybe we should skip the tests in that case.

PR #224, opened 2016-07-18
Add methods to read and write to/from existing ndarray in GpuArray

tsirif:
I think I need methods to copy to and from existing buffer in a numpy array in order to minimize copies and avoid creating redundant arrays in Platoon. How does this seem?

### DIFF
--- a/pygpu/gpuarray.pyx
+++ b/pygpu/gpuarray.pyx
@@ -1466,6 +1466,21 @@ cdef class GpuArray:
             raise IndexError, "cannot index with: %s" % (key,)
 
     def write(self, np.ndarray src not None):
+        """Writes host's Numpy array to device's GpuArray.
+
+        :param src: source array in host
+        :type src: np.ndarray
+
+        :raises ValueError: If this GpuArray is not compatible with `src`
+
+        """
+        if not self.flags.behaved:
+            raise ValueError, "Destination GpuArray is not well behaved: aligned and writeable"
+        if not ((self.flags.c_contiguous and src.flags['C_CONTIGUOUS'] and src.flags['ALIGNED']) or \
+                (self.flags.f_contiguous and src.flags['F_CONTIGUOUS'] and src.flags['ALIGNED'])):
+            raise ValueError, "GpuArray and Numpy array do not match in contiguity or Numpy array is not aligned"
+        if self.dtype != src.dtype:
+            raise ValueError, "GpuArray and Numpy array do not have matching data types"
         cdef size_t npsz = np.PyArray_NBYTES(src)
         cdef size_t sz = gpuarray_get_elsize(self.ga.typecode)
         cdef unsigned i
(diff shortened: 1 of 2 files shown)


pygpu/gpuarray.pyx
```diff
@@ -1465,6 +1465,26 @@ cdef class GpuArray:
         else:
             raise IndexError, "cannot index with: %s" % (key,)
 
+    def write(self, np.ndarray src not None):
```
abergeron: The name of the array should be `dst`.

pygpu/gpuarray.pyx
```diff
@@ -1465,6 +1465,26 @@ cdef class GpuArray:
...
+
+    def read(self, np.ndarray dst not None):
+        cdef size_t npsz = np.PyArray_NBYTES(dst)
+        cdef size_t sz = gpuarray_get_elsize(self.ga.typecode)
+        cdef unsigned i
+        for i in range(self.ga.nd):
+            sz *= self.ga.dimensions[i]
+        if sz != npsz:
+            raise ValueError, "GpuArray and Numpy array do not have the same size in bytes"
+        array_read(np.PyArray_DATA(dst), sz, self)
```
abergeron: You might want to check that the array is writeable `dst->flags & NPY_ARRAY_BEHAVED) == NPY_ARRAY_BEHAVED`.

`array_read()` will fail if the source is not contiguous (and will scribble everywhere if the destination is not contiguous).  So you might want to check for that.

Also the check done here for compatibility is very basic.  Do you think it could be extended a bit to cover things like compatible shape?

pygpu/gpuarray.pyx
```diff
@@ -1465,6 +1465,26 @@ cdef class GpuArray:
...
 
+    def write(self, np.ndarray src not None):
+        cdef size_t npsz = np.PyArray_NBYTES(src)
+        cdef size_t sz = gpuarray_get_elsize(self.ga.typecode)
+        cdef unsigned i
+        for i in range(self.ga.nd):
+            sz *= self.ga.dimensions[i]
+        if sz != npsz:
+            raise ValueError, "GpuArray and Numpy array do not have the same size in bytes"
+        array_write(self, np.PyArray_DATA(src), sz)
```
abergeron: This will read out of bounds if the source is not contiguous.  You will also get strange results if the source and destination don't agree on strides.

Finally this will definitely give weird results if the data types don't agree.

I thinks the checks should be extended to cover those.

pygpu/gpuarray.pyx
tsirif: I will fix for the first two. I think we should be flexible in respect to shape :)

pygpu/gpuarray.pyx
```diff
@@ -1465,6 +1465,57 @@ cdef class GpuArray:
...
+
+    def read(self, np.ndarray dst not None):
+        """Reads from this GpuArray into host's Numpy array.
+
+        :param dst: destination array in host
+        :type dst: np.ndarray
+
+        :raises ValueError: If this GpuArray is not compatible with `src`
+
```
nouiz: I would add in the docstring why this is useful as otherwise people can simply use a_gpu_ndarray[...] = an_ndarray and it will work and it could be optimized to be as efficient as the current case.

The difference is mainly that this one only check the total number of element and not the shapes. Also tell that this will always be as fast or faster then  a_gpu_ndarray[...] = an_ndarray.

PR #226, opened 2016-07-20
Make the definition of c_contiguous and f_contiguous stricter.

abergeron:


src/gpuarray_array.c
```diff
@@ -1117,8 +1121,7 @@ int GpuArray_is_c_contiguous(const GpuArray *a) {
   int i;
 
   for (i = a->nd - 1; i >= 0; i--) {
-    if (a->dimensions[i] != 1 && a->strides[i] != size)
-      return 0;
+    if (a->strides[i] != size) return 0;
```
nouiz: I find that too restrictive. What about:
if (a->strides[i] != size) continue;
else if(a->dimensions[i] == 1 && a->strides[i] % size == 0) continue;
else return 0;

abergeron: If I understand the intent of the code `shape = (1, 3, 1, 1)` and `strides = (4, 4, 12, 12)` would pass as C-contiguous when it is obviously not.  This is what I'm trying to fix.

Otherwise I don't get it since the check is written in a confusing way that completely changes the way it was done before.

nouiz: Just to be clear, the problem is that the case you give is only f
contiguous and not c contiguous. But currently it
get detected as both.

My proposed could be reformulated as (but it don't work for your case)

if ((a->dimensions[i] != 1 && a->strides[i] != size) || (a->strides[i] %
size == 0)) return 0;

Instead of trying to find what to do, can you check what the master of
numpy does?

nouiz: ok, it is better to be too strict then not enough.

PR #229, opened 2016-07-22
Feature/get device count

tsirif:
I need it for multi-node support in order to have as a default to use all cuda devices with workers per node if not configured else explicitly :)


abergeron:
Could it be possible to have an equivalent for OpenCL.  Counting the number of devices should not be a cuda-specific interface.

tsirif:
I am not sure how should I express a common interface with opencl and cuda due to the fact that opencl has "platforms" and "devices". Should this function return the sum of devices on all available platforms for opencl? Would this be useful?

We can have python function return a number, representing the number of devices, when we want to count cuda devices and a list of numbers, representing the number of devices on each platform, when we want to count opencl devices. But I think this means writing another extension for opencl which does so in C code. What do you say?

abergeron:
I think having a count of all devices on all platforms would be the closest equivalent.  At worst have some function for OpenCL that just returns an error.  It would be better than to have that as an extension.

Or you could make the API require a platform number (with a single available platform for cuda).

Otherwise we could revise the API to include an explicit platform number and use that.

tsirif:
I like the second more I think. Would a `get_platform_count` be a opencl extension in addition to the "buffer" level function `get_device_count`?

abergeron:
get_platform_count should be an "official" api.  You can make the cuda version always return 1.

tsirif:
With should I prefix the functions `get_platform_count` and `get_device_count` in buffer.c/.h?

tsirif:
@abergeron How is it now?
I also changed FindOpenCL to latest given by CMakeLists.txt and I also added HINTS for find_library for unix based on what "find_path" returned, because I had some problems: headers and library were not belonging to the same installation in my system.

src/gpuarray/buffer.h
```diff
@@ -44,6 +44,29 @@ struct _gpukernel;
 typedef struct _gpukernel gpukernel;
 
 /**
+ * \brief Gets information about the number of available platforms for the
+ * backend specified in `name`.
+ * \param name [const char*] the backend name
+ * \param platcount [int*] will contain number of compatible platforms in host
+ * \return int GA_NO_ERROR, if success
+ */
+GPUARRAY_PUBLIC int get_platform_count(const char* name, int* platcount);
```
abergeron: This needs to have a prefix of at minimum `gpu_` since it's a public interface.

Same for the one below too.

src/private.h
```diff
@@ -74,6 +74,8 @@ typedef struct _partial_gpucomm {
 } partial_gpucomm;
 
 struct _gpuarray_buffer_ops {
+  int (*get_platform_count)(int* platcount);
+  int (*get_device_count)(int platform, int* devcount);
```
abergeron: Make the platform and device count an unsigned int because you can't have a negative count.

abergeron:
Ok, with my two comments.

CMakeModules/FindOpenCL.cmake
```diff
@@ -1,94 +1,150 @@
...
+#
+# Distributed under the OSI-approved BSD License (the "License");
+# see accompanying file Copyright.txt for details.
+#
+# This software is distributed WITHOUT ANY WARRANTY; without even the
+# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the License for more information.
+#=============================================================================
+# (To distribute this file outside of CMake, substitute the full
+#  License text for the above reference.)
```
abergeron: You should probably respect what is said here.

tsirif: What does this mean exactly? Where should I find this?

abergeron: From where you found FindOpenCL.cmake.  I guess from the text that it is bundled with cmake past a certain version, so the copyright file it refers to is probably in the cmake sources.

tsirif:
@abergeron how is it now?

CMakeModules/FindOpenCL.cmake
```diff
@@ -29,8 +29,38 @@
 # implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 # See the License for more information.
 #=============================================================================
-# (To distribute this file outside of CMake, substitute the full
-#  License text for the above reference.)
+# CMake - Cross Platform Makefile Generator
+# Copyright 2000-2016 Kitware, Inc.
+# Copyright 2000-2011 Insight Software Consortium
```
abergeron: These parts shouldn't be here.

CMakeModules/FindOpenCL.cmake
```diff
@@ -20,7 +20,7 @@
 #
 
 #=============================================================================
-# Copyright 2014 Matthaeus G. Chajdas
+# Copyright 2014-2016 Matthaeus G. Chajdas
```
abergeron: This should stay with the original year.

CMakeModules/FindOpenCL.cmake
```diff
@@ -29,8 +29,38 @@
 # implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 # See the License for more information.
 #=============================================================================
```
abergeron: All this text, up to, but not including the copyright notice, should be removed.

PR #233, opened 2016-08-01
Small fixes

abergeron:


src/gpuarray_array.c
```diff
@@ -579,7 +581,8 @@ int GpuArray_setarray(GpuArray *a, const GpuArray *v) {
   tv.dimensions = a->dimensions;
   tv.strides = strs;
   /* This could be optiomized by setting the right flags */
-  tv.flags &= ~(GA_C_CONTIGUOUS|GA_F_CONTIGUOUS);
+  if (tv.nd != 0)
```
nouiz: this need a test in pygpu or libgpuarray.

nouiz:
Merging event without the test as this should fix current  problems

PR #234, opened 2016-08-01
Fix compilation error in device.c

tsirif:


abergeron:
This fix is part of https://github.com/Theano/libgpuarray/pull/233

abergeron:
I'll still merge it since it might be a while until that one gets reviewed and it's annoying in the meantime.

PR #239, opened 2016-08-10
[do not merge] Add an implementation of the buffer api for GpuArray.

abergeron:


abergeron:
This is possibly broken so don't merge.

abergeron:
This apparently overrides the __asarray__ special method and makes conversion to numpy use raw GPU memory (which usually makes the interpreter crash).

So this is definitely not ready for a merge.

abergeron:
This is a bad idea.

abergeron:
Mostly because it made python segfault on access to a numpy array created from a GpuArray.  Also because I'll probably never finish it.

PR #241, opened 2016-08-15
Skip the xsplit test when numpy is older than 1.11.

abergeron:


pygpu/tests/test_operations.py
```diff
@@ -21,18 +21,23 @@ def test_array_split():
...
         numpy.testing.assert_allclose(pc, numpy.asarray(pg))
 
+
 def test_split():
     for spl in (3, [3, 5, 6, 10]):
         yield xsplit, '', (9,), spl
 
+
 def test_xsplit():
+    if numpy.version.version < '1.11':
```
nouiz: grep in theano for numpy_version for how to do that.

PR #243, opened 2016-08-29
make take1 kernel use actual typecode of ind

davidweichiang:
If the index array passed to take1 has a dtype other than int64/ssize_t, it was returning the wrong result; this change uses whatever the index array's actual dtype is.

Thanks!


abergeron:
Yep, you're right.

PR #252, opened 2016-09-07
Check that arguments have some flags that apply

abergeron:
If you forget to tag one of the arguments as either read or write, then it will be completely ignored.

### DIFF
--- a/pygpu/_elemwise.pyx
+++ b/pygpu/_elemwise.pyx
@@ -60,6 +60,8 @@ cdef class arg:
             self.a.flags |= GE_WRITE
         if scalar:
             self.a.flags |= GE_SCALAR
+        if self.a.flags == 0:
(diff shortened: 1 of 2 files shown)


pygpu/_elemwise.pyx
```diff
@@ -60,6 +60,8 @@ cdef class arg:
             self.a.flags |= GE_WRITE
         if scalar:
             self.a.flags |= GE_SCALAR
+	if self.a.flags == 0:
```
nouiz: What if it is not readable, not writable and scalar? It won't raise and error. Is that normal?

abergeron: If is it scalar then it is always readable and never writeable.  So it's ok.

abergeron:
What is happening with this?

PR #269, opened 2016-10-05
Adding ipc memory handle support

hma02:
This PR adds support for doing interprocess memory access on the GPUArray backend. This support is the basis for implementing parallel loading in Theano-MPI, where the training process can access those loaded input images (in shared_x) from loading process without spending the time to load them within the training process.


hma02:
I noticed there's a commit 7fb6061 also included in another PR #239.

abergeron:
Apparently the commits I "lost" where in a branch on github so I made a PR out of that.  The ipc part should work so that PR can be merged.

The buffer part (which is in #239) breaks integration with bumpy so we can't merge that.

hma02:
@abergeron 
I see. 
Is the buffer part (7fb6061) necessary for implementing ipc? If not, maybe we can remove that part from the PR?

Thanks

abergeron:
It's not necessary for ipc, however If I remember correctly, you needed it for MPI.

hma02:
@abergeron 

Yeah, that was for implementing the CUDA-aware MPI allreduce. Since that also needs the as_buffer function which we don't have working yet (also it's easier just to work with NCCL to do communication since pygpu wraps NCCL), so I think we don't need to include that for the purpose of this PR.

PR #276, opened 2016-10-13
Internal scheduler with integer arguments + Factorizer improvements + Assorted bugfixes

obilaniu:
Added, for developer usage, a scheduler function that accepts `uint64_t` arguments rather than `ga_factor_list`s.

Also:

- Refactored and improved factorizer logic. In particular, always attempt an optimal 2-, 3- or 5-smooth factorization depending on `k`-smoothness constraint, instead of 2-smooth (round-up-to-next-power-of-2) only. For very large `n`, the next 5-smooth number is usually less than 1% higher. Compare to the next 2-smooth number, which can be as much as 100% higher.
- Many new testcases and extra checks for older testcases.
- Switched to >2x faster primality checker (BPSW).
- Important inline assembler bugfix, for an error causing miscompilation in Release mode.
- Renamed `gaIFLsnprintf()` to `gaIFLsprintf()` and use `sprintf()`, which is a C89 standard library function.
- Bugfixes to broken `gaIFLappend()`, including a mismatch between the header and source function names and an uninitialized variable `j` warning.
- Cleaned whitespace.

### DIFF
--- a/src/util/integerfactoring.c
+++ b/src/util/integerfactoring.c
@@ -995,12 +995,22 @@ static int      gaIFactorize5Smooth(uint64_t n, ga_factor_list* fl){
 		}
 
 		for(i3=0, p3=1;i3<=40;i3++, p3*=3){
+			/**
+			 * Detect when the product p3*p5 would overflow 2^64.
+			 */
+
+			if(i3){
+				nCurr = (p3/3)*p5;
+				if(nCurr+nCurr < nCurr || nCurr+nCurr+nCurr < nCurr+nCurr){
+					break;
+				}
+			}
 			nCurr = p3*p5;
 
 			/**
 			 * If the current product of powers of 3 and 5 is >= n, then this
-			 * must be the last iteration, but perhaps a pure power of 3 is the
-			 * best choice, so check for this.
+			 * must be the last iteration, but perhaps a pure product of powers
+			 * of 3 and 5 is the best choice, so check for this.
 			 */
 
 			if(nCurr >= n){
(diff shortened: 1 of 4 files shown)


src/util/integerfactoring.c
```diff
@@ -698,22 +1253,145 @@ static uint64_t gaIFLGetSmallestFactorv(int n, const ga_factor_list* fl, int* id
 	return hasFactors ? f : 1;
 }
 
+int      gaIFLsprintf(char* str, const ga_factor_list* fl){
```
abergeron: I already changed that function to use strb_appendf instead.  snprintf is available on MSVC, but you need to to contortions to get it which I'd rather not replicate all over the place.

If you want to print a factor list for debugging you can use either strb_cstr(), followed by the usual printf() or `strb_dump(strb *, FILE *)` to write the content of the strb to the specified FILE handle.

abergeron: Yes I did miss that.  However, I wonder why you need an sprintf-like function.  If it is only for debugging, I would prefer to go with append() only.

obilaniu: @abergeron A factor list constructed by `gaIFactorize()` can _always_ be printed in 128 bytes or less (corresponding to `2**64`). If you're printing something simple (error message or suchlike), and not constructing a huge string with `strb`, it's handy to just dump the factor list into some temporary `char buf[128]` on the stack using `gaIFLsprintf()` and output it as `%s` (string) using any `printf()`-like function. It requires no dynamic memory allocation.

abergeron:
jenkins test this please

PR #279, opened 2016-11-04
Make MaxAndArgmax use internally signed axis numbers.

obilaniu:
This avoids an infinite loop when counting ndX downto 0 with unsigned
integers, as the condition `unsigned >= 0` always holds.

### DIFF
--- a/src/gpuarray_reduction.c
+++ b/src/gpuarray_reduction.c
@@ -27,27 +27,27 @@ struct maxandargmax_ctx{
 	GpuArray*       dstMax;
 	GpuArray*       dstArgmax;
 	const GpuArray* src;
-	unsigned        reduxLen;
-	const unsigned* reduxList;
+	int             reduxLen;
+	const int*      reduxList;
 	
 	/* General. */
 	int             ret;
-	unsigned*       axisList;
+	int*            axisList;
 	gpucontext*     gpuCtx;
 	
 	/* Source code Generator. */
 	const char*     dstMaxType;
 	const char*     dstArgmaxType;
-	unsigned        ndd;
-	unsigned        ndr;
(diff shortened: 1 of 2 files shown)


nouiz:
jenkins test this
add to whitelist

nouiz:
if you can add a test that was failing and now pass, it would be great. I let @abergeron check in more detail the code.

nouiz:
jenkins test this

abergeron:
Nothing was failling before since there was a check on the arguments to prevent it.  However it made the reduce-all-axes case return an error.

src/gpuarray_buffer_cuda.c
```diff
@@ -1305,6 +1305,10 @@ static int cuda_callkernel(gpukernel *k, unsigned int n,
     }
 
     switch (n) {
+    case 0:
+      ctx->err = cuLaunchKernel(k->k, 1, 1, 1, 1, 1, 1, shared,
+                                ctx->s, args, NULL);
+      break;
```
abergeron: I'm not sure we want to support 0-dim at this level.  This is basically useless.

If you really need it special case it where you do the kernel call.

obilaniu: @abergeron I amended the commit, that file is no longer touched. However, I do think it made sense. The pattern is pretty clear:

- 3-dims is a volume XYZ
- 2-dims is a matrix XY with Z=1
- 1-dims is a vector X with Y=Z=1
- 0-dims is a scalar with X=Y=Z=1

abergeron: The pattern is clear but running a 1, 1, 1 kernel is almost always a mistake and I don't want to encourage it.

PR #280, opened 2016-11-05
Eliminate nearly all GCC warnings under -Wall -Wextra -Wno-unused-parameter.

obilaniu:
Eliminated warnings:

- `-Wsign-compare`
- `-Wempty-body`
- <strike>`-Wunused-parameter`</strike>

As well, a possibly logic fault was identified in `src/gpuarray_buffer_opencl.c` involving lack of braces around the branches of an `if/else`.

GCC now builds libgpuarray almost without any warnings (the exception is `gpuarray_reduction.c`, which is the subject of another PR and separately reports under `-Wmissing-field-initializer` even though in that particular case the usage is intentional and correct).

### DIFF
--- a/src/gpuarray_array.c
+++ b/src/gpuarray_array.c
@@ -280,7 +280,7 @@ int GpuArray_index_inplace(GpuArray *a, const ssize_t *starts,
       return GA_VALUE_ERROR;
     }
     if (steps[i] == 0 &&
-	(starts[i] == -1 || starts[i] >= a->dimensions[i])) {
+	(starts[i] == -1 || starts[i] >= (ssize_t)a->dimensions[i])) {
       free(newdims);
       free(newstrs);
       return GA_VALUE_ERROR;
@@ -663,10 +663,10 @@ int GpuArray_reshape_inplace(GpuArray *a, unsigned int nd,
 
     for (ok = oi; ok < oj - 1; ok++) {
       if (ord == GA_F_ORDER) {
-        if (a->strides[ok+1] != a->dimensions[ok]*a->strides[ok])
+        if (a->strides[ok+1] != (ssize_t)a->dimensions[ok]*a->strides[ok])
           goto need_copy;
       } else {
-        if (a->strides[ok] != a->dimensions[ok+1]*a->strides[ok+1])
+        if (a->strides[ok] != (ssize_t)a->dimensions[ok+1]*a->strides[ok+1])
           goto need_copy;
       }
     }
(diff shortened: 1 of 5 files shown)


src/gpuarray_array.c
```diff
@@ -280,7 +280,7 @@ int GpuArray_index_inplace(GpuArray *a, const ssize_t *starts,
       return GA_VALUE_ERROR;
     }
     if (steps[i] == 0 &&
-	(starts[i] == -1 || starts[i] >= a->dimensions[i])) {
+	(starts[i] == -1 || starts[i] >= (ssize_t)a->dimensions[i])) {
```
abergeron: I would cast the other way around just like 6 lines above to be consistent.

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -520,6 +520,25 @@ static int hgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
...
+  (void)lda;
+  (void)B;
+  (void)offB;
+  (void)ldb;
+  (void)beta;
+  (void)C;
+  (void)offC;
+  (void)ldc;
+  (void)batchCount;
+
```
abergeron: This is rather ugly.  Is there another way to convince GCC that unused args are not an error here?

src/gpuarray_buffer_opencl.c
```diff
@@ -1453,11 +1462,12 @@ static int cl_property(gpucontext *c, gpudata *buf, gpukernel *k, int prop_id,
...
 static const char *cl_error(gpucontext *c) {
   cl_ctx *ctx = (cl_ctx *)c;
-  if (ctx == NULL)
+  if (ctx == NULL){
     return get_error_string(err);
-  else
+  }else{
     ASSERT_CTX(ctx);
     return get_error_string(ctx->err);
+  }
```
abergeron: Good catch.  This was luckily working as intended, but still confusing.

src/gpuarray_blas_cuda_cublas.c
obilaniu: @abergeron It's hideously ugly but AFAIK portable to MSVC too. Your alternatives are to
1) Create a compiler-specific `#define UNUSED __attribute__((unused))` and stick that on every parameter and hope MSVC has something like it that goes in the same position (But I think MSVC's solution is to simply not write the name of the argument, which GCC rejects as "parameter name omitted"),
2) Use a flag like  `-Wnounused-parameter` after `-Wall -Wextra` to disable the warning
3) Use compiler-specific `pragma`'s equivalent to `-Wnounused-parameter`.

obilaniu: @abergeron Will nuke the `-Wunused-parameter` commits.

obilaniu: @abergeron They're gone and the tests pass. When `CMAKE_C_FLAGS` is `-Wall -Wextra -Wno-unused-parameter` the only warnings are out of `gpuarray_reduction.c`, which the other PR handles.

PR #282, opened 2016-11-09
Fix for scheduling in some corner cases.

abergeron:
This makes the case reported at https://github.com/Theano/Theano/issues/5154 go from 
~3.6e-2 per call to ~4.3e-4 per call on my card.

We could maybe gain a few more percent (and avoid other corner cases) with a better scheduling, but I am too tired to finish that.


nouiz:
After ~30m of checking how all the scheduling is done, I think it is good enough for now and should always be better then before.

PR #286, opened 2016-11-16
Add dynamic loading of libraries.

abergeron:
This will make the library build with support for everything all the time and will attempt to load the required libraries whenever required.

It has been tested to work on Linux, Windows and macOS.

Arnaud


nouiz:
Can you tell on which version of linux/windows/mac? Mostly to have a record
of that then anything else.

abergeron:
I've tested on macOS 10.10 and 10.12.  On Ubuntu 16.04 and Debian 6 and on Window 7.

src/gpuarray_buffer_opencl.c
```diff
@@ -110,21 +123,24 @@ cl_ctx *cl_make_ctx(cl_context ctx, int flags) {
...
   gpukernel *m;
 
+  e = setup_lib();
+  if (e != GA_NO_ERROR)
+    return NULL;
   id = get_dev(ctx, NULL);
   if (id == NULL) return NULL;
   err = clGetDeviceInfo(id, CL_DEVICE_QUEUE_PROPERTIES, sizeof(qprop),
-                        &qprop, NULL);
+                         &qprop, NULL);
```
nouiz: why those extra spaces? This don't follow the normal convention. There is a few others bellow

src/CMakeLists.txt
```diff
@@ -8,14 +8,13 @@ endif()
...
 find_package(NCCL)
 
-find_package(OpenCL)
-if(OpenCL_FOUND)
+find_package(OpenCL REQUIRED)
 find_package(clBLAS)
 if(NOT CLBLAS_FOUND)
-find_package(CLBlast)
-endif()
+  find_package(CLBlast)
```
nouiz: Why not always include clblas and clblast?

src/gpuarray_buffer_opencl.c
```diff
@@ -811,7 +824,6 @@ static gpukernel *cl_newkernel(gpucontext *c, unsigned int count,
       FAIL(NULL, GA_VALUE_ERROR);
     p = clCreateProgramWithBinary(ctx->ctx, 1, &dev, lengths, (const unsigned char **)strings, NULL, &ctx->err);
     if (ctx->err != CL_SUCCESS) {
-      clReleaseProgram(p);
```
nouiz: why this change? It don't seem related. Is it an unrelated bugfix?

src/private.h
```diff
@@ -264,6 +264,53 @@ GPUARRAY_LOCAL void gpukernel_source_with_line_numbers(unsigned int count,
                                                        size_t *newl,
                                                        strb *src);
 
+static uint16_t float_to_half(float value) {
```
nouiz: why this is needed? I think using the native operation from cuda/opencl is better then using this.

src/gpuarray_buffer_opencl.c
abergeron: A first version of this code added a 'g' prefix to all the functions, thus the extra spaces.  I failed to remove all of them.

src/gpuarray_buffer_opencl.c
abergeron: Yes, it's an unrelelated bugfix.  If we fail to create the program, we don't need to release it (and it could crash if we aren't lucky).

src/CMakeLists.txt
abergeron: The lastest code does.  This was an intermediate step.

src/private.h
abergeron: We can't use either native operations since they only work on the devices.

nouiz:
ok

Issue #205, opened 2016-06-28
ielemwise2 broadcasting too overzealous?


davidweichiang:
If broadcast=True is set, ielemwise2 will not throw an exception if the output array has an axis with dimension 1 but the corresponding input array has dimension >1. Moreover, the result is not what I'd expect (below, I'd expect [6,9,12], but it is [5,7,9]).

```
import numpy
import pygpu

ctx = pygpu.init('cuda')
na = numpy.array([[1,2,3]])
ga = pygpu.array(na, context=ctx)
nb = numpy.array([[1,2,3],[4,5,6]])
gb = pygpu.array(nb, context=ctx)

try:
    na += nb
    print "this should not happen"
except:
    print "this should happen"

try:
    pygpu.elemwise.ielemwise2(ga, "+", gb, broadcast=True)
    print "this should not happen"
except:
    print "this should happen"
```

Output:

```
this should happen
this should not happen
```

davidweichiang:
More specifically, I think it should be an error for a write argument to be broadcastable. However, check_basic doesn't know which arguments are write arguments, so I don't know how to implement the fix.

nouiz:
Thanks for the report. I'm able to reproduce this. I label it as a bug.

nouiz:
fixed. thanks for the report.

PR #287, opened 2016-11-17
Raise an exception to prevent issue #205.

notoraptor:
I suggest this PR to prevent the bug #205
where ielemwise2 does run when broadcast=True
and output array has dimensions smaller than
input array.

I added a checking that raises an exception to
avoid any case when the output array could be
broadcasted, as suggested by davidweichiang on
issue 205.

I also added some tests related to this change:

`DEVICE=cuda0 nosetests pygpu.tests.test_elemwise:test_ielemwise2_output_broadcast`

@nouiz @abergeron

### DIFF
--- a/src/gpuarray_elemwise.c
+++ b/src/gpuarray_elemwise.c
@@ -275,13 +275,12 @@ static int check_basic(GpuElemwise *ge, void **args, int flags,
   GpuArray *a = NULL, *v;
   unsigned int i, j, p, num_arrays = 0, nd = 0, nnd;
   int call32 = 1;
-  int output_ever_found;
 
   /* Go through the list and grab some info */
   for (i = 0; i < ge->n; i++) {
     if (is_array(ge->args[i])) {
       num_arrays++;
-      if (a == NULL) {
+      if (a == NULL || !is_output(a)) {
         a = (GpuArray *)args[i];
         nd = a->nd;
       }
@@ -290,7 +289,8 @@ static int check_basic(GpuElemwise *ge, void **args, int flags,
     }
   }
 
-  if (a == NULL)
+  /* No output arrays, this is an error */
+  if (a == NULL || !is_output(a))
     return GA_VALUE_ERROR;
 
   /* Check if we need to grow the internal buffers */
@@ -302,7 +302,7 @@ static int check_basic(GpuElemwise *ge, void **args, int flags,
   }
 
   /* Now we know that all array arguments have the same number of
-     dimensions */
+     dimensions and that the expected output size is the size of a */
 
   /* And copy their initial values in */
   memcpy(ge->dims, a->dimensions, nd*sizeof(size_t));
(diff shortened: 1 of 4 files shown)


notoraptor:
Ok ! So, I'll move the checking in C side and I'll simplify the test in Python code.

notoraptor:
Hello !

@abergeron , about moving the checking in the C code, I think I have just understood what davidweichiang said in the issue #205 about the check_basic() function :^^: (https://github.com/Theano/libgpuarray/issues/205#issuecomment-230823260) !

In fact, in the C code, we can't determine which arguments are the write arguments. Currently (based on (i)elemwise() functions in Python code), the write argument is the first argument in the args list, but it is a practical case, and GpuElemwise theorically does not disallow to have more than 1 write arguments (it depends on the expression).

So, in the C code, I can't guarantee to check the right write arguments.

Another problem in the C code is that we can't distinguish between an inplace operation (e.g a = a + b) and a not-inplace operation (eg. res = a + b), because it depends only on the expression of the op, which is a string ! But this bug is only related to the case when we have an inplace operation (**i**elemwise2).

It seems that the only really strong solution for these both problems would be to parse the expression string, as it is that expression which in fact determines which args will be written only, read only, or both written and read (and this latter case is what we want to check here).

So, should I parse the expression string ? For the moment I don't find a better solution.

notoraptor:
EDIT: Ah, no, there is `arg.flag` which may contains GE_READ and GE_WRITE.

So I may find read-write args. Sorry.

nouiz:
I let @abergeron review this PR.

notoraptor:
New update, @abergeron !

New tests and all tests passed in Python 2 and Python 3.

```
## Compilation and installation (debug mode for libgpuarray).
cd libgpuarray/

rm -rf Build build && \
mkdir Build && \
cd Build && \
cmake .. -DCMAKE_BUILD_TYPE=Debug && \
make && \
sudo make install && \
sudo ldconfig && \
cd .. && \
python setup.py build && \
python setup.py install

cd ..

## Testing.
# New tests.
DEVICE=cuda0 nosetests -vs pygpu.tests.test_elemwise:test_elemwise_rw_args_not_broadcasted
# All tests.
DEVICE=cuda0 python -c "import pygpu; pygpu.test()"
```

src/gpuarray_elemwise.c
```diff
@@ -321,20 +323,49 @@ static int check_basic(GpuElemwise *ge, void **args, int flags,
   n = 1;
   for (j = 0; j < nd; j++) {
     p = 0;
+    read_write_arrays_found = 0;
```
abergeron: The exact number of arrays doesn't matter, so I would make this a flag like `dim_fixed` or something of the sort.

src/gpuarray_elemwise.c
```diff
@@ -31,6 +31,7 @@ struct _GpuElemwise {
 STATIC_ASSERT(GEN_CONVERT_F16 == GE_CONVERT_F16, same_flags_value_elem1);
 
 #define is_array(a) (ISCLR((a).flags, GE_SCALAR))
+#define is_rw_array(a) (ISSET((a).flags, GE_READ) && ISSET((a).flags, GE_WRITE))
```
abergeron: You should check for write only, we still don't want to broadcast output-only arguments.  Something like
`#define is_output(a) (ISSET((a).flags, GE_WRITE)`.

src/gpuarray_elemwise.c
```diff
@@ -321,20 +323,49 @@ static int check_basic(GpuElemwise *ge, void **args, int flags,
...
+          /* We count the number of read-write arrays found until now
+           * that are not broadcasted. */
+          if (is_rw_array(ge->args[i]))
+            ++read_write_arrays_found;
+        } else {
           if (ISCLR(flags, GE_BROADCAST)) {
             return GA_VALUE_ERROR;
           }
           /* GE_BROADCAST is set */
           if (ge->dims[j] == 1) {
```
abergeron: I would add `&& !dim_fixed` here.  This should take care of dimensions that are fixed with the new flag and prevent change if it were to make an output array broadcasted.

src/gpuarray_elemwise.c
```diff
@@ -321,20 +323,49 @@ static int check_basic(GpuElemwise *ge, void **args, int flags,
...
+            }
+            /* There are no read-write arrays before the current array.
+             * So broadcasting can be done safely. */
             ge->dims[j] = v->dimensions[j];
+            /* We still count the current array if it's a read-write array
+             * (useless in the current implementation, but coherent). */
+            if (is_rw_array(ge->args[i]))
+              ++read_write_arrays_found;
           } else {
             if (v->dimensions[j] != 1) {
```
abergeron: I would add a check here like `|| is_output(ge->args[i])` and it should cover the case below.

src/gpuarray_elemwise.c
notoraptor: When we have not-inplace ops, I think the broadcasting of write-only output is allowed, isn't it ?

For example, something like:
```
matrix_a of shape (1, 4)
matrix_b of shape (5, 4)
matrix_c = elemwise2(matrix_a, '+', matrix_b, matrix_a, broadcast=True)
```

abergeron: No, that should not be allowed, since it doesn't really make sense.  We have 5 output values to store in a single memory location.  How do we decide which do we store?  Do we take the mean? The sum?

notoraptor: Currently, the 1st dimension of matrix_a will be "repeated" 5 times: for example:

```
[[1,2,3,4]]
```
becomes
```
[[1,2,3,4]
[1,2,3,4]
[1,2,3,4]
[1,2,3,4]
[1,2,3,4]]
```
And then this temp matrix is added to matrix_b and then the result is stored in matrix_c (so matrix_c will be a matrix of shape (5,4)). That's how NumPy does it currently on my computer! (and pygpu also, for the moment).

So, do we disallow that behavior for GpuElemwise ?

abergeron: In this case, a and b are read-only so what I propose will not trigger.  But broadcasting on c would not be allowed.

abergeron: I'm referring to the case where you would have c =  a + b, where a is (5, 4) b is (5, 4) and c is (1, 4).  That should not work even if we don't read from c.

notoraptor: Ah ok, I see !

src/gpuarray_elemwise.c
notoraptor: I handle it separately in the code that follows the line.

notoraptor:
New update @abergeron !

notoraptor:
PS: I have also run the CTest and it succeeds:

`DEVICE=cuda0 ctest -V -R test_elemwise`

(but I should notice that test_reduction fails when I run all tests, and I don't know why for the moment).

tests/check_elemwise.c
```diff
@@ -479,6 +479,12 @@ START_TEST(test_basic_broadcast) {
 
   ck_assert_int_eq(GpuElemwise_call(ge, rargs, GE_NOCOLLAPSE), GA_VALUE_ERROR);
 
+  ck_assert_int_eq(GpuElemwise_call(ge, rargs, GE_NOCOLLAPSE | GE_BROADCAST), GA_VALUE_ERROR);
```
abergeron: This is a poor test.  It would pass even if the new broadcast check doesn't do anything, since we try to output (2, 3) and c has shape (1, 6).  The 3 will not match the 6.

notoraptor:
New update @abergeron !

abergeron:
Please get rid of the merge commits.

jenkins test this please

notoraptor:
Corrected, @abergeron !

abergeron:
jenkins test this please

add to whitelist

PR #291, opened 2016-11-22
Small changes for the release

abergeron:


src/CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ set_target_properties(gpuarray PROPERTIES
   COMPILE_FLAGS "-DGPUARRAY_BUILDING_DLL -DGPUARRAY_SHARED"
   INSTALL_NAME_DIR ${CMAKE_INSTALL_PREFIX}/lib
   MACOSX_RPATH OFF
+  VERSION 0.0
```
nouiz: How we can get access to that version in Theano? Does this version if an interface change or any code version upgrade? I think we need to keep that information available in the version in some way.

abergeron: Normally this automatically handled through shared library versioning.

If you change an interface, you bump the first number, if you add new things you bump the second number.

nouiz:
but the PR on theano just remove all versioning stuff. It should use the
new versioning interface instead.

abergeron:
The shared library linker of the OS will handle this if we put the versioning in the library like this.

This is why we don't need custom handling in Theano anymore.

nouiz:
Test for the set strides are missing.

abergeron:
There is a test now (and the same checks that numpy does).

nouiz:
Except for the docstring, we can merge this PR.

Issue #290, opened 2016-11-22
More transparent error messages


khaotik:
Following #289, I spent quite some time finding a rather trivial bug. In file `gpuarray_buffer_cuda.c`, calls to `cuModuleLoadData` should be changed to `cuModuleLoadDataEx`, which can log info/errors. Also the error messages should be shown to the user. Right now, the JIT compilation error would only result in a `GA_IMPL_ERROR`.

PR #293, opened 2016-11-24
better error message for CUDA JIT linking

khaotik:
For #290.

- Now uses `cuModuleLoadDataEx` to link compiled kernel.

- Linker error during `cuModuleLoadDataEx` will be logged in the same manner as compiler error.

To test this (on linux):

1. rename `/path/to/nvidia-driver/libnvidia-ptxjitcompiler.so*` to something invalid

2. `sudo ldconfig`

3. run `check_elemwise` in C tests to see if error message pops up.

error message for me:
```
CUDA kernel link failure::

Linker error log:
fatal   : Unable to load library 'libnvidia-ptxjitcompiler.so.367.44'
```


nouiz:
I'll let @abergeron review this.

jenkins test this

Issue #292, opened 2016-11-23
BLAS dot implementation


khaotik:
TL;DR There should be a way to call vector-vector dot.

Consider the following theano code:
```
T = theano.tensor
x = T.matrix()
y, _ = th.scan(fn=lambda x : T.dot(x,x), sequences=x)
fn = theano.function([x], y)
x_val = numpy.random.rand(1000, 10000).astype(numpy.float32)
timeit( fn(x_val) )
```

On my laptop, the code runs 3.4 sec with the new backend, and 0.85 sec on the old backend with a [small patch](https://github.com/Theano/Theano/pull/5257). Theano don't have shape information at compile time, so the above code was converted into `gemv`, which is much slower than a vector-vector dot on GPU. This is one of the major reasons for the [huge GPU scan overhead](https://github.com/Theano/Theano/issues/1168).

Was about to make a PR like the one for the old backend. I'm a bit unsure what's the best approach to address this problem with the new backend:

1. Modify the `gemv` code in `libgpuarray` to detect row-vector matrix. This is the easiest, however rather hackish/ugly/shortsighted solution. The code must test `alpha==1.0 && beta==0.0` at runtime. This has to be done for all CUDA/OCL/\<potential future api\> implementation.

2. Add `dot` function in `libgpuarray`'s BLAS API, then modify `c_code` in Theano to do the shape detection. Cleaner than the 1st approach. The problems is `libgpuarray` will have to use a whole array object to represent a single dot product result -> another overhead.

3. Make Theano do better shape inference or scan optimization. The best permanent solution, with the potential to bring many other optimizations possible. This might get complicated ...

nouiz:
Thanks for investigating. What is the time in the old back-end without the
patch? Just to make sure it is about 3.4s as in the new back-end. Just to
make sure we don't have an extra overhead in the new back-end. I don't
expect one, but as you are working on this, double checking that would be
great.

3. won't always work. Take the example when the full Theano graph is just a
dot. Then we can't infer any shapes. If the user pass a row and a col for
both matrices, it would be great to still have the optimized version. So
this mean we need a quick check of shapes at run time.

Personally, I would go with 1. It is simple to use and do the work. In
fact, I find it sad that cublas don't do that optimization itself! I don't
find that solution hackish/... I find it as a simple solution that is
clean. It respect the interface and what the user wanted (compute a gemv).
It just does it in a faster way.

@abergeron, do you have objections to approch 1?

abergeron:
[...]

PR #294, opened 2016-11-24
BLAS vector-vector dot

khaotik:
For #292.


src/gpuarray/blas.h
```diff
@@ -8,6 +8,12 @@
 extern "C" {
 #endif
 
+// only for vector-vector dot
+GPUARRAY_PUBLIC int GpuArray_dot( GpuArray *A, GpuArray *B,
```
abergeron: You declare _dot and then use _rdot.  That might not work.

src/gpuarray_buffer_cuda.c
```diff
@@ -1122,15 +1137,46 @@ static gpukernel *cuda_newkernel(gpucontext *c, unsigned int count,
       FAIL(NULL, GA_MEMORY_ERROR);
     }
 
-    ctx->err = cuModuleLoadData(&res->m, bin);
+    // for both info/err log
+    cujit_info_log = (char*)malloc(2*cujit_log_size*sizeof(char));
+    if(cujit_info_log == NULL) {
```
abergeron: Space after if.

src/gpuarray_buffer_cuda.c
```diff
@@ -1122,15 +1137,46 @@ static gpukernel *cuda_newkernel(gpucontext *c, unsigned int count,
       FAIL(NULL, GA_MEMORY_ERROR);
     }
 
-    ctx->err = cuModuleLoadData(&res->m, bin);
+    // for both info/err log
+    cujit_info_log = (char*)malloc(2*cujit_log_size*sizeof(char));
```
abergeron: No need to cast the return of malloc.

abergeron:
This PR seems to be more about using cuModuleLoadDataEx than blas dot.

The blas parts should either be completed or removed.

abergeron:
Ok sorry, I just realized that you just reused the same branch for this PR, so the commits from the other one are in here.  You should probably rebase on the current master.

src/gpuarray_buffer_cuda.c
khaotik: I think it's more friendly to possible inclusion from C++ code, but I'll remove it anyway.

abergeron: The headers should be c++ clean, but the source files will probably not be touched by a c++ compiler.

src/gpuarray_array_blas.c
```diff
@@ -5,6 +5,94 @@
...
+      X->typecode != GA_DOUBLE)
+  return GA_INVALID_ERROR;
+
+  if (X->nd != 1 || X->nd != 1 || Y->nd != 0 ||
+      X->typecode != Y->typecode || X->typecode != Z->typecode)
+    return GA_VALUE_ERROR;
+  if (!(X->flags & GA_ALIGNED) || !(Y->flags & GA_ALIGNED) ||
+      !(Z->flags & GA_ALIGNED))
+    return GA_UNALIGNED_ERROR;
+  if (X->dimensions[0] != n || Y->dimensions[0] != n)
```
abergeron: You don't set `n` before this.  Also you could just compare them to each other.

src/gpuarray_array_blas.c
```diff
@@ -5,6 +5,94 @@
...
+    if (nocopy)
+      return GA_COPY_ERROR;
+    else {
+      err = GpuArray_copy(&copyY, Y, GA_ANY_ORDER);
+      if (err != GA_NO_ERROR)
+	goto cleanup;
+      Yp = &copyY;
+    }
+  }
+  if (Z->strides[0] < 0) {
```
abergeron: Z doesn't have strides, it's a scalar.

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -786,6 +786,97 @@ static int dgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
...
+
+  cuda_enter(ctx);
+
+  GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(X, CUDA_WAIT_READ));
+  GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(Y, CUDA_WAIT_READ));
+  GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(Z, CUDA_WAIT_ALL));
+
+  // we should store dot result on device
+  cublasGetPointerMode(h->h, &pmode);
+  cublasSetPointerMode(h->h, CUBLAS_POINTER_MODE_HOST);
```
abergeron: I think you meant to set the mode to DEVICE here.

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -786,6 +786,97 @@ static int dgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
...
+        gpudata *Z
+        ) {
+    return GA_DEVSUP_ERROR;
+}
+
+static int sdot(
+        size_t N,
+        gpudata *X, size_t offX, size_t incX,
+        gpudata *Y, size_t offY, size_t incY,
+        gpudata *Z) {
```
abergeron: There should be an offset for Z.

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -786,6 +786,97 @@ static int dgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
...
+  ASSERT_BUF(Y);
+  ASSERT_BUF(Z);
+
+  if (LARGE_VAL(N)) return GA_XLARGE_ERROR;
+
+  cuda_enter(ctx);
+
+  GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(X, CUDA_WAIT_READ));
+  GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(Y, CUDA_WAIT_READ));
+  GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(Z, CUDA_WAIT_ALL));
```
abergeron: We should only wait for write on Z since it is not read by dot.

abergeron:
Also, please get rid of the merge commit and do a proper rebase.

abergeron:
Just to note that you still have a merge commit.  Just a different one now.

After a rebase, when you want to push to github, git will tell you to run `git pull`.  This is wrong and you should run `git push -f` instead.

Otherwise, the code progresses nicely.  Good job!

khaotik:
@abergeron 

Mostly done, just a few problems.

Changes

 - BLAS dot for `libgpuarray`, with CUDA/clBLAS/clBLAST backend
   `clblasSdot` need a working buffer of size `N`, probably for sum reduction. I just made a naive implementation (allocate then release). I was about to make a static buffer that is shared between calls, however it's not thread safe. For now, there's some overhead with the naive impl.
 - pygpu binding and tests
 - changed all strides (`inc*` arguments) to type `int` for potential negative strides
   I did this because it's specified in BLAS standard, and is supported in some CPU BLAS libs. Not sure if these GPU libraries would change in future.

src/gpuarray_blas_opencl_clblas.c
```diff
@@ -146,54 +147,145 @@ static int dgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
...
+  cl_int cl_err;
+  cl_uint num_ev = 0;
+  cl_event evl[3];
+  cl_event ev;
+  cl_mem scratch_mem;
+
+  scratch_mem = clCreateBuffer(
+          ctx->ctx, CL_MEM_READ_WRITE, N*sizeof(float), NULL, &cl_err);
+  if (cl_err != CL_SUCCESS)
+      return GA_MEMORY_ERROR;
```
abergeron: The temporary buffer allocation is ok, however please use cl_alloc() to allocate it so we can have one central location for all allocations.

src/gpuarray_blas_opencl_clblast.c
```diff
@@ -436,6 +529,9 @@ GPUARRAY_LOCAL gpuarray_blas_ops clblast_ops = {
   setup,
   teardown,
   error,
+  hdot, /* TODO */
+  sdot, /* TODO */
+  ddot, /* TODO */
```
abergeron: These are not TODO.

src/private.h
```diff
@@ -113,6 +113,19 @@ struct _gpuarray_blas_ops {
   int (*setup)(gpucontext *ctx);
   void (*teardown)(gpucontext *ctx);
   const char *(*error)(gpucontext *ctx);
+
+  int (*hdot)( size_t N,
+    gpudata *X, size_t offX, int incX,
+    gpudata *Y, size_t offY, int incY,
```
abergeron: Don't use int, use `ssize_t` for a signed type.  We don't want to limit the size of arrays at this level.

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -39,8 +39,8 @@ typedef struct _blas_handle {
 
 static const char *code_sgemvBH_N_a1_b1_small =                         \
   "extern \"C\"__global__ void sgemv(const float *A[], size_t lda, "    \
-  "                                  const float *x[], size_t incx, "   \
-  "                                  float *y[], size_t incy, "         \
+  "                                  const float *x[], int incx, "   \
+  "                                  float *y[], int incy, "         \
```
abergeron: Here you changed the data type in the kernel without changing the declared data type.  This will break badly.

src/private.h
khaotik: It seems some BLAS functions are using `int` as stride (such as `*gemv *ger`) while others are using `size_t`. For now I just made a revert to `size_t`. I'm experimenting with `ssize_t` however getting test errors. Still need some work to finish.

abergeron:
jenkins test this please

Issue #296, opened 2016-11-25
GpuElemwise fails for "uneven broadcasts"


kohr-h:
I'm not sure if that is intended use of broadcasting, but `GpuElemwise` fails if I try to compute `out <-- a * x + b * out` where `out` and `x` have shape `(10, 10)`, `a` has shape `(10, 1)` and `b` is a scalar. See [this gist](https://gist.github.com/kohr-h/1d77c030aae32c8665eaf5908774767c) for a full example.

It's a bit hard for me to tell if I'm using the class correctly since the only thing I did basically was look a the code of the `ielemwise2` function and try to generalize it. Help is welcome.

abergeron:
Your code is correct, but there was a bug in the underlying C code which I just fixed: #297.

Thank you for the report and the easy-to-use test case.

kohr-h:
Nice! Thanks for the quick reply.

kohr-h:
Code works for my case.

nouiz:
fixed.

PR #297, opened 2016-11-25
Fix call of elemwise in the basic case when using scalars.

abergeron:
There was an index bug which would cause the library to crash in that case.  I've fixed it and added a test for that.


abergeron:
Fixes #296

nouiz:
ok to merge when jenkins pass

nouiz:
There is a temporary problem with jenkins that made it fail.

nouiz:
jenkins test this

PR #298, opened 2016-11-25
Add offset handling for indices and result in take1()

abergeron:
fix gh-https://github.com/Theano/Theano/issues/5285


abergeron:
I need some tests for the new case.  They will come later.

nouiz:
Is there a bug report for this? If so, can you link to it?

nouiz:
seem good.

nouiz:
I added a link to what it fix.

abergeron:
Now I have a test.

PR #301, opened 2016-11-29
Some Fixes for Windows from conda.

mingwandroid:


abergeron:
I'm ok with the changes, execpt for the stdint.h ones, which should use gpuarray/config.h.

That header could be modified to only include the bundled copy when necessary though.

mingwandroid:
OK, thanks, I'm on it now.

mingwandroid:
The stdint.h issue I had seems to have gone away now.

mingwandroid:
OK, so I had the wrong end of the stick regarding stdint.h, please see https://github.com/Theano/libgpuarray/pull/301/commits/ccc3ba7577dc32b7d316bee6ac76a965046005ef for what I hope is the correct fix.

nouiz:
I don't know what happened, but in this PR, you have commits from another PR.

To get this merged faster, I'll make another PR with just the right commits.

nouiz:
I cherry-picked the commits in https://github.com/Theano/libgpuarray/pull/305

thanks

mingwandroid:
Yeah that's weird, I simply rebased and then force pushed.

PR #303, opened 2016-11-29
Fix typo in take1 offset test.

abergeron:


nouiz:
the fix seem good. Is this a fix to the test not working? I don't understand why this would fix the test failure I have.

nouiz:
ok to merge when jenkins pass

abergeron:
Yes, it fixes the test.

It was failing because we were trying to write more data than what fits into the array due to the larger declaration.

PR #304, opened 2016-11-29
Get the largest allocatable block size

abergeron:
This also includes a location change for blas loading.


lamblin:
Can you temporarily change in travis.yml the way theano is installed, so that it gets it from your updated branch?

src/CMakeLists.txt
```diff
@@ -88,7 +88,7 @@ set_target_properties(gpuarray PROPERTIES
   INSTALL_NAME_DIR ${CMAKE_INSTALL_PREFIX}/lib
   MACOSX_RPATH OFF
   # This is the shared library version
-  VERSION 0.0
+  VERSION 0.1
```
nouiz: I'll come back to the versioning. Now have a new interface. This won't trigger the recompilation and won't give good user warning. If people update Theano but not libgpuarray, then they will get compilation error related to the convolution as GA_CTX_PROP_LARGEST_MEMBLOCK isn't defined. Check jenkins buildbot

http://darjeeling.iro.umontreal.ca:8080/job/Theano%20gpu/550/testReport/junit/theano.gpuarray.tests.test_dnn/TestDnnInferShapes/test_conv_time_on_shape_change_valid_conv/

If we keep it like that, then we will frequently have useless user questions. We will loose our own time and user time. If we bump the major version, I don't think it will give a good user error. We need to fix that. It could be for an 0.9rc2, but it should be before 0.9.

nouiz:
jenkins failed: http://darjeeling.iro.umontreal.ca:8080/job/libgpuarray_PR/63/testReport/junit/nose.failure/Failure/runTest/

src/gpuarray_buffer_cuda.c
```diff
@@ -443,6 +443,21 @@ static void find_best(cuda_context *ctx, gpudata **best, gpudata **prev,
...
 
+static size_t largest_size(cuda_context *ctx) {
+  gpudata *temp;
+  size_t sz, dummy;
+  cuda_enter(ctx);
+  ctx->err = cuMemGetInfo(&sz, &dummy);
+  cuda_exit(ctx);
+   /* We guess that we can allocate at least a quarter of the free size
+     in a single block. This might be wrong though. */
+  sz /= 4;
```
abergeron: This case handles memory that hasn't been preallocated.  We can't query the largest block available for cuMalloc, so I am resorting to a guess here.

src/CMakeLists.txt
abergeron: I don't think that it is a problem that changing the minor version here will not trigger a recompilation.  Everything that worked with 0.0 will work with 0.1.  That is the point of this version scheme.

If we were to change the major version it would make the currently compiled modules unloadable, possibly triggering a recompilation (I'm not 100% sure how Theano deal with C code it can't reload).

If people update Theano and it uses newly introduced symbols, they will need to update libgpuarray also, yes.  Usually this is handled with recommended versions for releases and a guideline of the style: use the latest master of libgpuarray for the latest master of Theano.

My problem with maintaining another version scheme is that it duplicates works and we will forget to bump one or the other for some changes, which will lead to exactly the problems that you describe, except that nobody will expect them.

src/gpuarray_buffer_cuda.c
nouiz: Yes I understand that. My only point is to document that.

abergeron: I consider this to be an implementation detail (and a bad one at that) and I would prefer not to document it since I hope to change it to something better whenever possible.

nouiz: ok that the size / 4 is a detail. But there is no doc about which property to can be queried. I'll make an issue about that.

abergeron: There is documentation that is generated from the headers.  This should list all the defined properties.

src/gpuarray/buffer.h
```diff
@@ -689,6 +689,13 @@ GPUARRAY_PUBLIC gpucontext *gpukernel_context(gpukernel *k);
  */
 #define GA_CTX_PROP_PCIBUSID 19
 
+/**
+ * Get the largest single block of memory that can be allocted.
+ *
+ * Type: `size_t`
+ */
+#define GA_CTX_PROP_LARGEST_MEMBLOCK 20
```
abergeron: I can add it for sure, but I usually only add the values that I need, and I don't need this one currently.

nouiz:
in the C api, but not in the python api.

People should be able to use the python api without looking at the C api,
or at least, have a link from the python api to the relevent part of the c
api.

abergeron:
You want me to add a new context property that exposes this value?  In that case ok.

PR #308, opened 2016-12-02
Bump the required version of cython for the dict change.

abergeron:


nouiz:
ok to merge when jenkins pass. As we have problems with it, I prefer to wait that it work before merging.

abergeron:
This should be good now.

Issue #163, opened 2016-05-04
conda package or Python wheel?


dwf:
Is libgpuarray now stable enough to consider putting up an official conda package for a release version?

nouiz:
The master of Theano can't work with the dev version of libgpuarray. So we
would need to device some versioning that fix the version that Theano
support for correct depencies.

But having a conda install script would be useful I think.

Maybe in a few weeks we can have something more stable. There is a few
speed "regression" compared to current Theano gpu back-end that we need to
fix. I'm not sure if they will request interface change or not.

For the versioning, do you know which restriction we can have? Can you fix
version to 0.1.* ? Then when we break the interface we bump to 0.2.*? We
would just increase the * for new feature or speed up/bugfix.

On Wed, May 4, 2016 at 5:33 PM, David Warde-Farley <notifications@github.com

jakirkham:
I'd be happy to help package `libgpuarray`. We would need to have a formal release to do so though. Honestly a git tag would be sufficient. Though we can probably try from a commit to see if there are needed tweaks before the release is cut. We just wouldn't be able to release the package until there is a source release.

abergeron:
We now have a conda package that will soon be in the defaults channel.

Issue #265, opened 2016-09-30
libgpuarray pip install with Anaconda can't find headers


jessejohns:
Hello,

When I am installing pygpu with Anaconda and using pip install, I have to specify the include directory in **init**.py as follows:

```
def get_include():
    import os.path
    p = '/home/jesse/Programs/anaconda/pygpu'
    assert os.path.exists(os.path.join(p, 'gpuarray_api.h'))
    return p
```

Installing with python setup.py install, everything works out well.

abergeron:
Is the path you are specifying the one to your copy of the source?  In that case the problem is that some versions of pip/setuptools/distutils decide that a .h is not part of the package data (even though I mark it as such) and don't install it.

There are also some versions of Debian which decide to move it in a random directory that doesn't make sense.

I have not been able to figure out why it does that yet, although I haven't been spending too much time on it either since I can't reproduce on any of my machines.  I remember someone writing about MANIFEST.in which could help so you can try that.

jessejohns:
Hi there,

Pip ends up creating its down directory and throws the *.h into anaconda/pygpu and the .py into /anaconda/lib/python2.7/site-packages/pygpu.  I'm using the python/pip from the local anaconda installation and using ubuntu 16.04.  I'm able to reproduce the behavior on RHEL6 as well.

abergeron:
If you could find out why it does that and how to fix it that would help.  The .h file should be alongside the .py file.

Alternatively if there is a way to retrieve the path where pip want to hide files (without hardcoding) then that could be an alternative.

astooke:
Hey I think I just ran into a similar problem, possibly the same.  After doing all the bleeding edge Theano + libgpuarray + pygpu installation instructions (into a conda env), and updating library paths with libgpuarray install location (not the same as where I cloned to), running Theano with gpu would fail like so:

```
Problem occurred during compilation with the command line below:
...
fatal error: gpuarray/array.h: No such file or directory
compilation terminated.
```

Solution for me was less elegant: manually put a folder `gpuarray` with all the header files from `libgpuarray\include` in my anaconda env site-package: `/opt/anaconda2/envs/py3/lib/python3.5/site-packages/pygpu-0.6.0rc1-py3.5-linux-x86_64.egg/pygpu` (this is one of the places shown as always included, -I, in the Theano compilation attempt in the error message).

Maybe I missed a step?

abergeron:
[...]

PR #311, opened 2016-12-08
BLD: add conda recipes

kohr-h:
I saw #163 on the issue tracker and thought it would be a nice project for an evening.

Since your build system differentiates between the C library and the Python bindings, I figured the same split would make sense for conda packages, too. The advantage is that the Python wrapper can *depend* on the C library instead of having to *rebuild* it for every Python version. Instead, only the wrapper itself has to be recompiled.

The recipes are very simple, due to your simple build system (great to have dynamic loaders!), not much that can go wrong in the build phase. To build the conda packages, the following steps are necessary:

0. If necessary, exit from conda environments, builds should be run in the root environment. Enter the top level directory of your `libgpuarray` repo.
1. Build the C library:
   ```
   MPI_LIBRARY=/path/to/libmpi.so conda build conda/libgpuarray
   ```
   The MPI library environment variable is optional. During build, the system will be searched for an MPI library anyway (if left empty), this is just for non-standard locations.
   The conda package is put in the `conda-bld` directory in your Anaconda or Miniconda distribution and will be automatically picked up in the next step as build dependency.
2. Build the Python wrapper for a bunch of Numpy and Python versions:
   ```
   conda build conda/pygpu --python 2.7 --numpy 1.11
   conda build conda/pygpu --python 3.5 --numpy 1.11
   conda build conda/pygpu --python 2.7 --numpy 1.10
   conda build conda/pygpu --python 3.5 --numpy 1.10
   ...
   ```
Done.

Edit: adapted for Numpy version pinning.

### DIFF
--- a/conda/pygpu/meta.yaml
+++ b/conda/pygpu/meta.yaml
@@ -1,0 +1,41 @@
+package:
+    name: "pygpu"
+    version: "0.6.0rc1"
+
+source:
+    git_url: https://github.com/Theano/libgpuarray.git
+    git_rev: master  # TODO: this is for testing. Change to a version tag when done.
+
+build:
+    number: 0
+
+requirements:
+    build:
+        - python
+        - pip
+        - setuptools
+        - cython
+        - mako >=0.7
+        - numpy
+        - libgpuarray ==0.6.0rc1
+
+    run:
+        - python
+        - six
+        - nose
+        - mako >=0.7
+        - numpy
+        - libgpuarray ==0.6.0rc1
+
+test:
+    imports:
+        - pygpu
+
+about:
+    home: https://github.com/Theano/libgpuarray
+    license: MIT
+    license_file: LICENSE
+    summary: "Numpy-like wrapper on libgpuarray for GPU computations"
+
+extra:
+    maintainers:
--- a/conda/pygpu/meta.yaml
+++ b/conda/pygpu/meta.yaml
@@ -1,6 +1,6 @@
 package:
     name: "pygpu"
-    version: "0.6.0rc1"
+    version: "0.6.0"
 
 source:
     git_url: https://github.com/Theano/libgpuarray.git
@@ -17,7 +17,7 @@ requirements:
         - cython
         - mako >=0.7
         - numpy
-        - libgpuarray ==0.6.0rc1
+        - libgpuarray ==0.6.0
 
     run:
         - python
@@ -25,7 +25,7 @@ requirements:
         - nose
         - mako >=0.7
         - numpy x.x
-        - libgpuarray ==0.6.0rc1
+        - libgpuarray ==0.6.0
 
 test:
     imports:
--- a/conda/pygpu/meta.yaml
+++ b/conda/pygpu/meta.yaml
@@ -18,6 +18,8 @@ requirements:
         - mako >=0.7
         - numpy
         - libgpuarray ==0.6.1
+        # Avoid mkl download during build phase
+        - nomkl # [not win]
 
     run:
         - python
@@ -30,6 +32,9 @@ requirements:
 test:
     imports:
         - pygpu
+    requires:
+        # Avoid mkl download during testing
+        - nomkl # [not win]
 
 about:
     home: https://github.com/Theano/libgpuarray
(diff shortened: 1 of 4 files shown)


kohr-h:
Of course, OSX builds will require some tweaking, and Windows builds will need a `bld.bat`.

conda/pygpu/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - mako >=0.7
+        - numpy
+        - libgpuarray ==0.6.0rc1
+
+    run:
+        - python
+        - six
+        - nose
+        - mako >=0.7
+        - numpy
```
kohr-h: This must be `numpy x.x` to pin the Numpy version.

kohr-h:
> Since your build system differentiates between the C library and the Python bindings, I figured the same split would make sense for conda packages, too.

That said, since the build time is rather short anyway, a monolithic package would also be possible.

abergeron:
This is great!

I don't have time to test it this week, but I'll try to take some time to play with it next week and fix it so we can have macOS and Windows packages too.

kohr-h:
> This is great!
> 
> I don't have time to test it this week, but I'll try to take some time to play with it next week and fix it so we can have macOS and Windows packages too.

Cool. For the windows builds, you may want to have a look at the BAT files [here](https://github.com/astra-toolbox/astra-toolbox/tree/master/python/conda). The package layout is similar - a compiled library + python bindings - but there the build process on windows is much more hacky. Here it should "just work" (TM) due to cmake. In the install phase, though, some of the copying around of stuff could be similar.

kohr-h:
I "overheard" a discussion in #265 about conda packages in preparation, plus some discussion [on the Theano issue tracker](https://github.com/Theano/Theano/issues/4705) about a related topic.
So has anybody had the time yet to test building on other platforms?

nouiz:
You can try the current packages:

conda install -c rdonnelly pygpu

@rdonnely told he would try to update them with the beta release of last
week, but he was pretty busy. I don't think he had time for it, but I
hopefully, he will be able to do it this week.

kohr-h:
Okay, nice. I don't think I have any way to test this on a different platform than Linux, but I'll check out the channel anyway.

nouiz:
We tested it on windows and mac. A few other people also did.

kohr-h:
Looks good from what I can see. 

What is think is missing is the `numpy x.x` version pinning in the `pygpu` recipe. It's needed because Numpy is used in the cython code. The ABI needs to match the one of the installed numpy version, but if the recipe just says "numpy" it will take the newest possible, not the one the package was built with.

Also, I know of [another conda quirk](https://github.com/conda/conda-build/issues/857), which is that selectors like `# [win]` don't make it into the final specification file `index.json` that determines the dependencies. Unless they fixed it, the recipe is processed during the build step, and all conditionals like the above are evaluated and substituted according to the *build* environment.
In practice this means that if you build a conda package for windows on Linux (which is nothing unusual with Python packages), and you have `# [not win]` in your *runtime* dependencies, those will end up unconditionally in the final recipe. For example, if you add `nomkl  # [not win]` to the meta.yaml, the final recipe will unconditionally pull in `nomkl` (and fail), provided you built on Linux.
It's a known issue, and I don't know a solution to it other than manually editing the `index.json` in the package, which is very ugly, or simply building the packages on the target platforms. I guess that's necessary here anyway because also the Python part requires a compilation step.
Anyway, it's something to be aware of.

abergeron:
I don't think anaconda support cross-platform building for C projects.  In any case we aren't trying to support that.

For the numpy version thing, I'll admit it's not something I encountered, but I'll raise the point.

kohr-h:
> I don't think anaconda support cross-platform building for C projects. In any case we aren't trying to support that.

You're right, I realized while writing that this is only an issue for pure Python projects when you build for all platforms.

abergeron:
There are official conda packages for pygpu and libgpuarray and we also maintain some on conda-forge.  Given that, this seem redundant.

nouiz:
Maybe we could make them easier to find?

kohr-h:
> Maybe we could make them easier to find?

I was about to ask the same thing. Maybe you want to put the recipes into the repo so that users can build their own packages? I find that very useful for development.

abergeron:
Which ones should be in the package? The official ones or the conda-forge ones?

Also, if we distribute recipes this means we have to maintain them.

nouiz:
Is that what we already do by having conda-forge package that you update?

PR #318, opened 2016-12-28
Ufuncs and reductions

kohr-h:
So here's version 1 of the ufuncs and reductions. Everything works as far as I can see, that means all ufuncs and reductions with all dtypes except float16.

In general, I have added a bunch of TODO notes in the code where I didn't quite know if there's a better way to implement something. Here are some more high-level TODOs:

Remaining open questions / TODOs:
- [ ] Add ufuncs as `ndgpuarray` methods?
- [ ] Expose ufuncs in the top-level `pygpu` namespace?
- [ ] Pre-compile kernels for ufuncs and reductions and a selection of dtypes and ndims? That would make them much faster when called for the first time.
- [ ] Which style should the docstrings adhere to? I used numpydoc style.
- [ ] How about complex math? Currently kernels crash with something like `res = cabs(a)`
- [ ] How to handle ufuncs with multiple outputs? Their value is probably marginal so I didn't bother for now.

Some notes on implementation details, problems and solutions. 

## Functions follow Numpy signature

All functions have the same signature as their Numpy counterparts, which is
- `ufunc(a[, out])` for unary ufuncs,
- `ufunc(a, b[, out])` for binary ufuncs,
- `reduction(a, axis=None, dtype=None, out=None, keepdims=False)` for `sum, prod` and
- `reduction(a, axis=None, out=None, keepdims=False)` for `amin, amax`

### Array-like input
**Problem 1:**
All array arguments are allowed to be array-like, which has  the consequence that in the extreme case when all arguments are not `GpuArray` instances, there is **no object from which to get a context**.
**Current solution:** A default context must be defined in that case.

**Problem 2:** Which class should be chosen for the output?
**Current solution:** `ndgpuarray`

## Ufunc signatures are matched with those of Numpy ufuncs
The `ufunc_dtypes` helper looks at `<ufunc>.types` and tries to find an input-output pair of dtypes that are adequate for the given input dtypes. Arrays are transformed to those types if necessary. This is a bit tricky for `ldexp` which needs a float and and integer array to work - that case is handled a bit hacky when the second argument is actually a scalar and not an array. 

## Some ufuncs would need helper functions
**Problem:** A couple of ufuncs, e.g. `floor_divide`, are currently implemented in a clumsy way that computes things multiple times in some cases. That's because everything has to be a one-liner. If that one-liner could use some external helper code or save intermediate results, the implementation would be more efficient. Some functions cannot be written easily as one-liners, but those are probably also of limited value. See the `MISSING_UFUNCS` list for reference.
**Current solution:** String composition and multiple function evaluation

## Random notes

- How do you handle contributed code?

- Have you ever considered using [pytest](http://pytest.org/) instead of `nose`? The current tests could be easily ported, and the output in case of failure is SOOO much better. Compare this
```
(pygpu_dev) hkohr@sabayon ~/git/libgpuarray $ GPUARRAY_TEST_DEVICE="opencl0:0" nosetests pygpu/tests/test_ufuncs.py -v
*** Testing for Quadro K2100M
test_all (pygpu.tests.test_ufuncs.test_binary_ufuncs) ... FAIL
test_all (pygpu.tests.test_ufuncs.test_reduction) ... ok
test_all (pygpu.tests.test_ufuncs.test_unary_ufuncs) ... ok

FAIL: test_all (pygpu.tests.test_ufuncs.test_binary_ufuncs)
  File "/home/hkohr/git/libgpuarray/pygpu/tests/test_ufuncs.py", line 200, in test_all
    self.check_binary_ufunc_scalar(ufunc, scalar, dtype1)
  File "/home/hkohr/git/libgpuarray/pygpu/tests/test_ufuncs.py", line 276, in check_binary_ufunc_scalar
    atol=rtol, equal_nan=True)
AssertionError
```
to this
```
============================================= FAILURES ==============================================
_____ test_binary_ufuncs_scalar[ binary_ufunc = 'remainder' - scalar = 0 - dtype = np.uint64 ] ______

binary_ufunc = 'remainder', scalar = 0, dtype = dtype('uint64')

    def test_binary_ufuncs_scalar(binary_ufunc, scalar, dtype):
        """Test GpuArray binary ufunc with scalar second operand."""
        # Invalid, don't test
        if np.issubsctype(dtype, np.unsignedinteger) and scalar < 0:
            return
        if binary_ufunc in ('fmod', 'floor_divide') and scalar == 0:
            return
    
        gpuary_ufunc = getattr(gpuary_ufuncs, binary_ufunc)
        npy_ufunc = getattr(np, binary_ufunc)
    
        npy_arr, gpuary_arr = npy_and_gpuary_arrays(shape=(2, 3),
                                                    dtype=dtype,
                                                    positive=True)
    
        try:
            res = np.finfo(dtype).resolution
        except ValueError:
            res = np.finfo(np.promote_types(dtype, np.float16)).resolution
        rtol = 10 * res
    
        try:
            npy_result = npy_ufunc(npy_arr, scalar)
        except TypeError:
            # Make sure we raise the same error as Numpy
            with pytest.raises(TypeError):
                gpuary_ufunc(gpuary_arr, scalar)
        else:
            if npy_result.dtype == np.dtype('float16'):
                # We use float32 as minimum for GPU arrays, do the same here
                npy_result = npy_result.astype('float32')
    
            gpuary_result = gpuary_ufunc(gpuary_arr, scalar)
            assert npy_result.shape == gpuary_result.shape
            assert npy_result.dtype == gpuary_result.dtype
>           assert np.allclose(npy_result, gpuary_result, rtol=rtol, atol=rtol,
                               equal_nan=True)
E           assert False
E            +  where False = <function allclose at 0x7f7e1cd0d598>(array([[0, 0, 0],\n       [0, 0, 0]], dtype=uint64), gpuarray.array([[9223372036854775808, 9223372036854775808, 9223372036854775808],\n       [9223372036854775808, 9223372036854775808, 9223372036854775808]], dtype=uint64), rtol=1.0000000000000002e-14, atol=1.0000000000000002e-14, equal_nan=True)
E            +    where <function allclose at 0x7f7e1cd0d598> = np.allclose

odl/test/space/gpuary_ufuncs_test.py:284: AssertionError
--------------------------------------- Captured stderr call ----------------------------------------
/home/hkohr/git/odl/odl/test/space/gpuary_ufuncs_test.py:267: RuntimeWarning: divide by zero encountered in remainder
  npy_result = npy_ufunc(npy_arr, scalar)
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
============================== 1 failed, 4953 passed in 15.32 seconds ===============================
```
As additional bonus, you don't need all the boilerplate code for iterating through all the parameter combinations, test parametrization is a natural feature of pytest. And no need to indent all test code by 4 characters :-)


abergeron:
This looks very interesting, but I won't have time to review in the coming week.  I'm sorry for the delay.

kohr-h:
> This looks very interesting, but I won't have time to review in the coming week. I'm sorry for the delay.

Don't worry, I'm still finalizing the last part, but in the meanwhile some issues have also gone away magically :-)
I'll bump this for a review sometime soon.

kohr-h:
Bump already.

This is actually in quite good shape now, only a few ufuncs missing and everything works apart from some documented cases and potentially corner cases that aren't tested for yet. I'll leave it at that for now and wait for a review.

kohr-h:
Maybe somebody can look into this at a point in the not-too-distant future?

I'd also like to add that NumPy [now exposes the `__array_ufunc__` interface](https://github.com/numpy/numpy/blob/master/doc/release/1.13.0-notes.rst#__array_ufunc__-added), which means that a class that implements `__array_ufunc__` gets to decide what to do when a NumPy ufunc is called on it. That's very good since somebody calling `np.sin` on a GPU array will then actually call the super-fast GPU elementwise kernel instead of casting to a NumPy array and then calling the NumPy implementation. (In another way that's bad since users may forget who's the workhorse :-) ).

Anyway, that would be a natural follow-up of this PR.

nouiz:
Not this week. It is nips deadline and Arnaud isn't here, but maybe/probably next week.

src/gpuarray_buffer_cuda.c
```diff
@@ -1025,6 +1025,8 @@ static int call_compiler(cuda_context *ctx, strb *src, strb *ptx, strb *log) {
     , "-G", "-lineinfo"
   };
   nvrtcResult err;
+  // DEBUG: print the kernel source
+  // printf("%s\n", src);
```
abergeron: Please remove this

pygpu/ufuncs.py
```diff
@@ -0,0 +1,1145 @@
...
+
+
+UFUNC_SYNONYMS = [
+    ('absolute', 'abs'),
+    # ('conjugate', 'conj'),
+    ('deg2rad', 'degrees'),
+    ('rad2deg', 'radians'),
+    ('true_divide', 'divide'),
+    ('maximum', 'fmax'),  # TODO: differ in NaN propagation in numpy, doable?
+    ('minimum', 'fmin'),  # TODO: differ in NaN propagation in numpy, doable?
```
abergeron: Adjusting NaN propagation is doable with a bit of C code.  If you don't want to do it, I'd rather we don't provide fmax/fmin than provide them with the wrong comportment.

kohr-h:
> I've noted a couple of minor problems, but I can't help but feel the approach is somewhat off.
> 
> This mimics the very surface of the functionality that ufuncs offers. It does not support accumulate, reduce and others. It does not allow the user to define their own ufuncs. It doesn't support the where and order arguments. Most importantly it doesn't unify the interface.
> 
> I understand that your time is limited and you might not have the time to do all those things right now, but the current structure would make those very hard to add. If at least all the ufuncs where instances of a class, then we could modify that class later to provide those things.

Thanks for the quick review. I agree that some flexibility wouldn't hurt. Making the ufuncs class instances that implement `__call__` shouldn't be a big deal.
On the other hand, I don't see how this keeps users from defining their own ufuncs. It's actually easier when they don't have to adhere to a specific class interface. In some sense, the `elemwise` helpers provide exactly that.

Anyway, my approach here was to get the basic stuff working first and not let the amount of code get too much out of hand. 
Certainly `ufunc.reduce` seems relatively easily doable, maybe also `outer`, but for the rest `accumulate`, `at`, `reduceat` and the `where` argument I don't really have a good idea. The last 3 require index arrays with broadcasting and all, which is not trivial as far as I can tell. The `accumulate` function may be doable with some `mako` templating, but I assume it would be rather slow due to bad parallelism.

Another way of playing this would be to just call into NumPy's function (e.g. `at`) in case there is no native implementation, instead of leaving it unimplemented. An upside of that would be that people can write code that works, and which is slow now but becomes fast when someone sits down and implements the feature. In the other scenario, people would use the NumPy ufunc and never learn about the added native feature.

I'll do the minimal fix and make ufuncs classes in any case.

kohr-h:
I made ufuncs class instances now, with a rather lame solution for ufuncs with different signatures (they are different classes; If you know a good solution for creating a callable class with `__call__` signature changing per instance, please let me know.)

I also added some native `ufunc.reduce` methods. For all other methods (like `accumulate`) the code simply calls into the corresponding Numpy function. What's currently missing is a fallback for `at`, I'll add that.

pygpu/ufuncs.py
```diff
@@ -0,0 +1,1330 @@
...
+            if ctx is not None and ary.context != ctx:
+                raise ValueError('cannot mix contexts')
+            ctx = ary.context
+            if cls is None or cls == GpuArray:
+                cls = ary.__class__
+            need_context = False
+
+    if need_context:
+        ctx = get_default_context()
+        cls = ndgpuarray  # TODO: sensible choice as default?
```
abergeron: This is ok for now.

pygpu/ufuncs.py
```diff
@@ -0,0 +1,1330 @@
...
+
+    # TODO: can CPU memory handed directly to kernels?
```
abergeron: No it can't.  Unless it's a scalar.

pygpu/ufuncs.py
```diff
@@ -0,0 +1,1330 @@
...
+    a = _prepare_array_for_reduction(a, out)
+    if a.dtype == numpy.bool:
+        neutral = 1
+    elif numpy.issubsctype(a.dtype, numpy.integer):
+        neutral = numpy.iinfo(a.dtype).max
+    elif numpy.issubsctype(a.dtype, numpy.floating):
+        neutral = 'INFINITY'
+    elif numpy.issubsctype(a.dtype, numpy.complexfloating):
+        raise ValueError('array dtype {!r} not comparable'
+                         ''.format(a.dtype.name))
```
abergeron: Just the else should be enough.

pygpu/ufuncs.py
```diff
@@ -0,0 +1,1330 @@
...
+        raise ValueError('array dtype {!r} not supported'
+                         ''.format(a.dtype.name))
+    return reduce_with_cmp(a, '>', neutral, axis, out, keepdims)
+
+
+# --- Elementwise ufuncs --- #
+
+
+# This dictionary is derived from Numpy's C99_FUNCS list, see
+# https://github.com/numpy/numpy/search?q=C99_FUNCS
```
abergeron: Take care with this because the code you are generating is not C99, but CUDA or OpenCL.  I'm pretty sure at least some of these will not be available, like `cbrt`.

abergeron:
In numpy the ufuncs are in C and they use the fact that python can't know the signature of C functions to trick the interpreter into thinking that there is a fixed number of arguments.

In reality the functions accept any number of arguments and perform a check on them to limit the number according to the instance.

In python it would be possible to "patch in" a __call__ to have a differing signatures, but I don't think this is worth the effort.

Also, if we don't have certain methods because they aren't available on the GPU (like at(), accumulate(), ...) then I'd rather they stay unimplemented for now rather than fallback to the CPU.  This way it can at least attract attention to the fact that it needs more work if somebody needs this.

pygpu/ufuncs.py
```diff
@@ -1153,27 +1149,33 @@ def _reduceat_not_impl(a, indices, axis=0, dtype=None, out=None):
             return NotImplemented
 
         self.accumulate = kwargs.pop('accumulate', _accumulate_not_impl)
-        self.accumulate.__name__ = self.accumulate.__qualname__ = 'accumulate'
+        self.accumulate.__name__ = 'accumulate'
+        if PY3:
+            self.accumulate.__qualname__ = 'accumulate'
```
abergeron: Why is this done different than the others?

abergeron:
jenkins test this please

kohr-h:
Circular dependency, needs fix

abergeron:
jenkins test this please

pygpu/ufuncs.py
kohr-h: Not intentionally, copy-paste error.

abergeron:
jenkins test this please

pygpu/ufuncs.py
kohr-h: So far I haven't found any that don't work, but that will probably depend on the CUDA version and may also be different for OpenCL.

abergeron:
It seems your code is not ready for python2.  Make sure to test it on a python2 install somewhere.

[six](https://pythonhosted.org/six/) can help with portability issues.

kohr-h:
There were some other issues. There is still a failure with `ldexp` (doesn't raise where it should), will look into it.

kohr-h:
Py2 failed due to some string issue, but there was an actual issue with `deg2rad`, missing precision. I also added an optional `context` argument to everything and simplified the ufunc wrapper a bit. Still a bit of remaining work, but we're getting closer.

PR #321, opened 2017-01-09
Add message for /usr/local for macOS.

abergeron:


doc/installation.rst
```diff
@@ -131,6 +131,11 @@ macOS.
 It might be possible to use a version of gcc built using Homebrew or
 MacPorts, but this is untested and unsupported.
 
+It appears that on some versions, /usr/local is not in the default
+compiler paths so you might need to add ``-L /usr/local/lib -I
+/usr/local/include`` to the command line to build the python module.
```
nouiz: Add "The line `python setup.py build -L`". Not everybody will understand which line, there is many lines.

abergeron: I'm saying "the command line to build the python module".  That should be obvious.

nouiz:
This isn't obvious. Many people that use Theano didn't do there undergrad
in computer science nor know how to compile a python module. If we don't
make it more clear, it is sure we will get user question. So I prefer to
make it more clear now and save us time later.

doc/installation.rst
```diff
@@ -131,6 +131,11 @@ macOS.
 It might be possible to use a version of gcc built using Homebrew or
 MacPorts, but this is untested and unsupported.
 
+It appears that on some versions, /usr/local is not in the default
+compiler paths so you might need to add ``-L /usr/local/lib -I
+/usr/local/include`` to the setup.py command.
```
nouiz: I think it is only for the build command. So I would write:

to the ``python setup.py build`` command.

When you amend a commit, this don't trigger a notification. So I just saw this change now.

abergeron: Should be ok now.

PR #323, opened 2017-01-11
Add ABI version handling in C code and Python code

notoraptor:
Add GPUARRAY_ABI_VERSION  and pygpu.gpuarray.abi_version() method to handle ABI version in gpuarray. A new header file (abi_version.h) auto-generated by CMake has been added for this purpose.

These changes must allow to fix Theano issue https://github.com/Theano/Theano/issues/5345 . A related Theano PR is coming.

@abergeron @nouiz


src/gpuarray/abi_version.h
```diff
@@ -0,0 +1,3 @@
+#ifndef GPUARRAY_ABI_VERSION
+#define GPUARRAY_ABI_VERSION "1.0"
+#endif
```
nouiz: I think this file should not be commited. Add it in the .gitignore file to help prevent readding it.

src/gpuarray/config.h
```diff
@@ -1,6 +1,8 @@
 #ifndef GPUARRAY_CONFIG
 #define GPUARRAY_CONFIG
 
+/* The following included file should have been generated by CMake. */
+#include <gpuarray/abi_version.h>
 #define GPUARRAY_API_VERSION 0
```
nouiz: I would remove this line, as it should be in the included file. Otherwise, it is an error and not having it with a wrong value should help with that.

notoraptor: The included file contains GPUARRAY_A**B**I_VERSION only, not A**P**I version !

src/CMakeLists.txt
```diff
@@ -96,13 +96,19 @@ add_library(gpuarray-static STATIC ${GPUARRAY_SRC})
 target_link_libraries(gpuarray ${CMAKE_DL_LIBS})
 target_link_libraries(gpuarray-static ${CMAKE_DL_LIBS})
 
+get_target_property(GPUARRAY_ABI_VERSION gpuarray VERSION)
+FILE(WRITE gpuarray/abi_version.h
+"\#ifndef GPUARRAY_ABI_VERSION\n\#define GPUARRAY_ABI_VERSION \"${GPUARRAY_ABI_VERSION}\"\n\#endif\n"
```
abergeron: Please make it an int like (major * 1000 + minor).  Having it as a string is just super inconvenient for most uses.

.gitignore
```diff
@@ -17,6 +17,7 @@ distribute*tar.gz
 *.so
 *.o
 *.log
+src/gpuarray/abi_version.h
```
abergeron: Group this with the other file in src below.

abergeron:
I'm ok with the idea, just don't use strings.

src/CMakeLists.txt
```diff
@@ -88,17 +88,22 @@ set_target_properties(gpuarray PROPERTIES
   INSTALL_NAME_DIR ${CMAKE_INSTALL_PREFIX}/lib
   MACOSX_RPATH OFF
   # This is the shared library version
-  VERSION 1.0
+  VERSION 1.1
```
abergeron: This doesn't warrant an ABI change.  This is an API change.

notoraptor:
Update. Add CMake variable SOVERSION (auto-generated and equals to the first part of VERSION, so that VERSION still remains the only variable to update when changing gpuarray ABI version).

Now the fix for theano issue https://github.com/Theano/Theano/issues/5345 works ! Theano does not recompile if VERSION changes from 1.X to 1.Y, but recompilation is done if VERSION changes from X.? to Y.? . No more errors in both cases.

src/CMakeLists.txt
```diff
@@ -88,21 +88,35 @@ set_target_properties(gpuarray PROPERTIES
   INSTALL_NAME_DIR ${CMAKE_INSTALL_PREFIX}/lib
   MACOSX_RPATH OFF
   # This is the shared library version
-  VERSION 1.0
+  VERSION 2.91
```
abergeron: Please reset this to the right version (1.0 for now)

notoraptor:
Updated. Back to current version (1.0).

Issue #295, opened 2016-11-24
Bool of gpuarray need fixing


nouiz:
https://github.com/Theano/Theano/issues/5252#issuecomment-261609228

gpuarray don't always evaludate to to the same True/False as numpy ndarray. See the issue above. We need to fix this.

nouiz:
Also, if there is more then 1 value, we need to raise the same error as numpy.

abergeron:
I wouldn't say this is a bug, but rather an enhancement.

nouiz:
One of the goal of libgpuarray/pygpu is to be able to replace numpy with an object on the gpu having:

if gpuarray(0):

take the true branch is a bug in that sence. We are telling people it is an ndarray on the GPU and we don't do that there and this cause wrong result compared to that we told. If we where always raising an error in that case, then adding the support would be an enhancement.

Anyway, this is just words, we agree we should do it :)

nouiz:
fixed.

PR #326, opened 2017-01-12
Add proper conversion to bool.

abergeron:


abergeron:
Fix for https://github.com/Theano/libgpuarray/issues/295

pygpu/gpuarray.pyx
```diff
@@ -1657,6 +1657,16 @@ cdef class GpuArray:
...
+    def __bool__(self):
+        if self.ga.nd == 0:
+            return True
+        if self.ga.nd == 1:
+            if self.ga.dimensions[0] == 0:
+                return False
+            if self.ga.dimensions[0] == 1:
+                return bool(numpy.asarray(self))
+        raise ValueError('The truth value of a multi-element array is ambiguous')
+
```
nouiz: Need tests.

Also, if nd==0, we should not always return True. We need to look at the data as in line 1667.

why not just do something like:

If self.ga.size == 0: return False
elif self.gz.size == 1: return bool(numpy.asarray(self))
raise ValueError(...)

Numpy accept do to bool on a matrix with only 1 element. I don't think it look at nd, but only at size.

nouiz:
jenkins fail

PR #327, opened 2017-01-12
Add simple build test on macOS and linux with travis.

abergeron:


.travis.yml
```diff
@@ -1,8 +1,13 @@
...
-os:
-  - linux
-  - osx
+matrix:
+  include:
+    - os: linux
+      compiler: gcc
+    - os: linux
+      compiler: clang
+    - os: osx
```
nouiz: In the past, travis didn't allowed a project to test on mac and on linux. Can you confirm that it worked somewhere?

nouiz:
travis didn't run.

.travis.yml
abergeron: I tried this PR with my fork on travis and it works.

abergeron:
Of course, we would have to activate it.  But if we do and this isn't merged, we will get a ton spam about failing tests.  I'd rather merge it first and then activate it.

nouiz:
Ok. I merged and activated travis. Here is the first run as I merged
another PR to trigger it:

https://travis-ci.org/Theano/libgpuarray/builds/192416426

Issue #316, opened 2016-12-23
Computations on float16 arrays produce garbage


kohr-h:
Probably related to #218:
```python
>>> a_h = np.ones((2, 3, 4), dtype='float16')
>>> a_d = pygpu.array(a_h, dtype='float16')
>>> reduce1(a_d, op='+', neutral=0, out_type='float16', axis=None)
gpuarray.array(-0.0078125, dtype=float16)
```
There's no option to convert float16 here, but the underlying GpuKernel defaults to `have_half=False` and that flag is not changed in `ReductionKernel` as far as I can see.

Reference:

[Half-precision arithmetic supported for "Pascal" architectures and upward](https://devblogs.nvidia.com/parallelforall/mixed-precision-programming-cuda-8/) (see section "Mixed Precision Performance on Pascal GPUs"; probably CUDA 8 required)

Edit: Forgot to mention that I have a Kepler GPU.

kohr-h:
Actually, the error already happens in conversion from float16 without any computation otherwise:
```python
>>> a_h = np.ones((2, 2), dtype='float16')
>>> a_d = pygpu.array(a_h, dtype='float16')
>>> a_d
gpuarray.array([[ 1.,  1.],
       [ 1.,  1.]], dtype=float16)
>>> a_d.astype('float32')
gpuarray.array([[ 15360.,  15360.],
       [ 15360.,  15360.]], dtype=float32)
```

abergeron:
Yes, that is normal, float16 is not correctly supported everywhere in libgpuarray operations.

We have to add some wrappers to make it work.

nouiz:
Until it is done, we should raise an not implemented error. We should never
return wrong value.

nouiz:
Now we raise an error, so the rest of this PR is to add support for this.

nouiz:
I'm closing and make a new issue with a better title to the current need:

https://github.com/Theano/libgpuarray/issues/316

PR #328, opened 2017-01-12
Raise an error for float16 in ReductionKernel.

abergeron:


abergeron:
Fix for https://github.com/Theano/libgpuarray/issues/316

pygpu/reduction.py
```diff
@@ -150,6 +150,11 @@ def __init__(self, context, dtype_out, neutral, reduce_expr, redux,
         else:
             self.arguments = arguments
 
+        if (self.dtype_out == np.dtype('float16') or
```
nouiz: np. -> numpy.

this make jenkins fail.

pygpu/reduction.py
```diff
@@ -150,6 +150,11 @@ def __init__(self, context, dtype_out, neutral, reduce_expr, redux,
...
+                any(ar.dtype == numpy.dtype('float16')
+                    for ar in self.arguments)):
+            raise UnsupportedError('float16 not supported for the reduction interface')
```
nouiz: Also a test to make sure the error is raised would be great.

PR #332, opened 2017-01-16
FIx the install of the headers (hopefully!).

abergeron:


nouiz:
Why jenkins passed before? Can you change it so that it test the problem?

PR #334, opened 2017-01-25
Generalize MaxAndArgmax to multiple ops and prepare groundwork for small-destination reductions

obilaniu:
Added multiple new reduction ops, enumerated in `enum ga_reduce_op`, and also heavily refactored internally in preparation to add a generator for "small code model" kernels, which will be used for reductions with a small destination.

``` c
typedef enum _ga_reduce_op {
	GA_REDUCE_SUM,             /*        +        */
	GA_REDUCE_PROD,            /*        *        */
	GA_REDUCE_PRODNZ,          /*        * (!=0)  */
	GA_REDUCE_MIN,             /*      min()      */
	GA_REDUCE_MAX,             /*      max()      */
	GA_REDUCE_ARGMIN,          /*     argmin()    */
	GA_REDUCE_ARGMAX,          /*     argmax()    */
	GA_REDUCE_MINANDARGMIN,    /* min(), argmin() */
	GA_REDUCE_MAXANDARGMAX,    /* max(), argmax() */
	GA_REDUCE_AND,             /*        &        */
	GA_REDUCE_OR,              /*        |        */
	GA_REDUCE_XOR,             /*        ^        */
	GA_REDUCE_ALL,             /*     &&/all()    */
	GA_REDUCE_ANY,             /*     ||/any()    */
} ga_reduce_op;
```

***DO NOT MERGE YET:*** Tests still failing on MaxAndArgmax, and no tests yet for any other op.


src/gpuarray_reduction.c
```diff
@@ -188,63 +695,101 @@ static void  appendIdxes                        (strb*              s,
...
 
-	ctx->dstMaxType    = ctx->dstArgmaxType = NULL;
+	ctx->srcTypeStr    = ctx->dstTypeStr    = ctx->dstArgTypeStr =
+	ctx->accTypeStr    = ctx->idxTypeStr    = NULL;
+	ctx->initVal       = NULL;
 	ctx->ndh           = 0;
+	ctx->ndhd          = 0;
+	ctx->ndhr          = 0;
 	ctx->sourceCode    = NULL;
+	ctx->s             = INIT_STRB;
```
abergeron: Strb can be initialized with all-zeros for the same effect as STRB_STATIC_INIT.

src/gpuarray_reduction.c
```diff
@@ -188,63 +695,101 @@ static void  appendIdxes                        (strb*              s,
...
 
-	ctx->hwAxisList[0] = ctx->hwAxisList[1] = ctx->hwAxisList[2] = 0;
-	ctx->blockSize [0] = ctx->blockSize [1] = ctx->blockSize [2] = 1;
-	ctx->gridSize  [0] = ctx->gridSize  [1] = ctx->gridSize  [2] = 1;
-	ctx->chunkSize [0] = ctx->chunkSize [1] = ctx->chunkSize [2] = 1;
+	for(i=0;i<MAX_HW_DIMS;i++){
```
abergeron: STYLE: space between for and (

src/gpuarray_reduction.c
```diff
@@ -256,113 +801,492 @@ static int   maxandargmaxCheckargs              (maxandargmax_ctx*  ctx){
 	ctx->nds = ctx->src->nd;
 	ctx->ndr = ctx->reduxLen;
 	ctx->ndd = ctx->nds - ctx->ndr;
+	strb_ensure(&ctx->s, 5*1024);
```
abergeron: Why do you preallocate so large a buffer?  Is the kernel you're generating really 5kb?

src/gpuarray_reduction.c
```diff
@@ -256,113 +801,492 @@ static int   maxandargmaxCheckargs              (maxandargmax_ctx*  ctx){
...
+
 
-	return ctx->ret;
+
+	return reduxSelectModel(ctx);
```
abergeron: I know that it is convenient to do tail calls everywhere like this, but C doesn't have tail-call optimization (or at least not all the time).

It also feels inappropriate that calling reduxCheckArgs ends up compiling the reduction kernel and calling it.

src/gpuarray_reduction.c
```diff
@@ -256,113 +801,492 @@ static int   maxandargmaxCheckargs              (maxandargmax_ctx*  ctx){
...
+	strb_appendf(&ctx->s, "typedef %s     X;/* The type of the indices: signed 32/64-bit. */\n",   ctx->idxTypeStr);
+	strb_appendf(&ctx->s, "typedef %s     K;/* The type of the accumulator variable. */\n",        ctx->accTypeStr);
+	strb_appends(&ctx->s, "\n");
+	strb_appends(&ctx->s, "\n");
+	strb_appends(&ctx->s, "\n");
+}
+static void  reduxAppendFuncGetInitVal     (redux_ctx*  ctx){
+	strb_appends(&ctx->s, "/**\n");
+	strb_appends(&ctx->s, " * Initial value function.\n");
+	strb_appends(&ctx->s, " */\n");
```
abergeron: I don't know it outputting comments in a generated kernel is useful at all.  Perhaps you can keep this as an actual comment right here.

src/gpuarray_reduction.c
```diff
@@ -256,113 +801,492 @@ static int   maxandargmaxCheckargs              (maxandargmax_ctx*  ctx){
...
+	strb_appends(&ctx->s, "\n");
+	strb_appends(&ctx->s, "WITHIN_KERNEL K    getInitVal(void){\n");
+	strb_appendf(&ctx->s, "\treturn (%s);\n", ctx->initVal);
+	strb_appends(&ctx->s, "}\n");
+	strb_appends(&ctx->s, "\n");
+	strb_appends(&ctx->s, "\n");
+	strb_appends(&ctx->s, "\n");
```
abergeron: Make this a single string.  Don't use an append per line, it is mostly inefficient.

src/gpuarray_reduction.c
```diff
@@ -21,106 +21,613 @@
...
+ */
+
+static int   reduxGetMinInit               (int typecode, const char** property){
+	switch(typecode){
+		case GA_BYTE2:
+		case GA_BYTE3:
+		case GA_BYTE4:
+		case GA_BYTE8:
+		case GA_BYTE16:
+		case GA_BYTE:               *property = "SCHAR_MIN"; break;
```
obilaniu: This might be problematic for the densest `switch`es I have, since that will triple their size for no gain in readability. Are all these STYLE fixes you suggest something you feel increases readability, or are they simply the output of a soulless `lint`er?

abergeron: No they are the result of me reading the code and my eyes bleeding a little.  Also, every other file in libgpuarray respects that style.  It's your code that stands out.

src/gpuarray_reduction.c
obilaniu: Yes, the kernel really is ~5KB for the 3D testcases.

src/gpuarray_reduction.c
obilaniu: In my previous version I tried to do all of the error handling in the highest-level function. That doesn't really scale as I add more and more and deeper functions, since each such function has to be able to detect the condition and return, and all its callers have to also detect a condition, abort and pass on the error until the highest level.

So I instead arranged the code in CPS style, and every function can abort the entire `GpuArray_reduction()` by tail-calling `reduxCleanup()` with whichever is the correct `GA_ERROR_CODE` as argument. A nice way of having clean "exception" handling in C.

CPS is somewhat unidiomatic C, but GCC does optimize my tailcalls from `callq` to `jmpq` in Release mode, or inlines them whole. And there is a net readability win from all the error handling code I've eliminated.

src/gpuarray_reduction.c
```diff
@@ -256,113 +801,492 @@ static int   maxandargmaxCheckargs              (maxandargmax_ctx*  ctx){
...
+	reduxAppendFuncKernel       (ctx);
+	reduxAppendFuncPostKernel   (ctx);
 }
-static void  maxandargmaxAppendTypedefs         (maxandargmax_ctx*  ctx){
+static void  reduxAppendIncludes           (redux_ctx*  ctx){
+	strb_appends(&ctx->s, "/* Includes */\n");
+	strb_appends(&ctx->s, "\n");
+	strb_appends(&ctx->s, "\n");
+	strb_appends(&ctx->s, "\n");
+}
```
obilaniu: I discovered that by myself, but forgot to nuke this function.

src/gpuarray_reduction.c
obilaniu: I think I'll change this to `(strb)STRB_STATIC_INIT`, but it would be lovely if `STRB_STATIC_INIT` incorporated `(strb)` itself.

abergeron: If it works on VS2008 then you can make the change.

src/gpuarray_reduction.c
obilaniu: I take it the answer is indeed that you consider the STYLE fixes do improve readability. I can modify these spots accordingly.

src/gpuarray_reduction.c
obilaniu: It will not, unless VS2008 has C99 extension support for _compound_ initializers (which is probably why I initialized a const variable to `STRB_STATIC_INIT` and used *that* to initialize `ctx->s`: Your static initializer `STRB_STATIC_INIT` only works on _declarations_, not on _expression statements_).

abergeron: Which is why I mention that you can use `memset(ctx->s, 0 sizeof(strb))` here.  In fact you can probably add an strb method to do just that.

obilaniu: The problem with that idea is that it's not future-proof. I have no reason to believe that `memset`ing `strb`'s to zero will always be a correct way of doing it. In C, the typical way this would have been resolved is to offer an `strb_init(strb*)` function; In fact that solution is more popular than static initializers. The current static initializer is currently the only reliable way of doing this, because `strb_clear(strb*)` presumes that the `strb` in question has been initialized in the past (by dint of the fact it unconditionally `free`s a pointer).

abergeron: Which is why I'm suggesting the addition of an strb_init that does the memset.  That way if it needs to change in the future we can change it there.

obilaniu:
@abergeron I've cleaned up the code given your feedback. It's still a WIP and has numerous bugs. However, I'm aware of most of what remains to be done.

src/gpuarray_reduction.c
abergeron: Then at least rename the functions so that it is clear that they do this.

src/util/strb.h
```diff
@@ -77,6 +77,15 @@ static inline int strb_error(strb *sb) {
   return sb->l == (size_t)-1;
 }
 
+/*
+ * Initialize at runtime an strb.
+ */
+
+static inline void strb_init(strb* sb){
+  const strb s = STRB_STATIC_INIT;
+  *sb = s;
```
abergeron: This can also be a `memset(b, 0, sizeof(strb))`.  We can assume that it is ok at this level.

src/gpuarray_reduction.c
obilaniu: @abergeron Alright, how shall I call them then? `reduxDoXYZ()`?

abergeron: I'm not sure.  Something like `reduxCheckArgsAndRest()`?  I don't like it very much ...

src/util/strb.h
obilaniu: @abergeron I think it would be wise to keep it as is, Don't Repeat Yourself and all that. If `STRB_STATIC_INIT` changes in such a way that `memset()` no longer suffices, this function will continue working. Otherwise there's two different places that must be changed, and one can forget one or the other.

At any rate, for the present `strb`, the compiler can and will optimize `memset()` to just
```
xor eax, eax
movq [sb   ], rax
movq [sb+ 8], rax
movq [sb+16], rax
```

abergeron: Will it optimize the current form too?

obilaniu: @abergeron As a matter of fact, a Release-mode build with GCC recognized that `strb`, along with everything else I set to zero explicitly in `reduxCheckArgs()`, could be memsetted to 0:

```
Dump of assembler code for function GpuArray_reduction:
   0x00007ffff74f5300 <+0>:     push   %r15
   0x00007ffff74f5302 <+2>:     xor    %eax,%eax           # Zero RAX
   0x00007ffff74f5304 <+4>:     push   %r14
   0x00007ffff74f5306 <+6>:     push   %r13
   0x00007ffff74f5308 <+8>:     push   %r12
   0x00007ffff74f530a <+10>:    mov    %rcx,%r12
   0x00007ffff74f530d <+13>:    mov    $0x2f,%ecx          # How many words to clear
   0x00007ffff74f5312 <+18>:    push   %rbp
   0x00007ffff74f5313 <+19>:    mov    %edi,%ebp
   0x00007ffff74f5315 <+21>:    push   %rbx
   0x00007ffff74f5316 <+22>:    sub    $0x1b8,%rsp
   0x00007ffff74f531d <+29>:    test   %r12,%r12
   0x00007ffff74f5320 <+32>:    lea    0x30(%rsp),%rdi     # Base pointer for memset
   0x00007ffff74f5325 <+37>:    lea    0x30(%rsp),%rbx
   0x00007ffff74f532a <+42>:    rep stos %rax,%es:(%rdi)   # Memset to 0.
   0x00007ffff74f532d <+45>:    mov    %ebp,0x30(%rsp)
   0x00007ffff74f5331 <+49>:    mov    %rsi,0x38(%rsp)
   0x00007ffff74f5336 <+54>:    mov    %rdx,0x40(%rsp)
   0x00007ffff74f533b <+59>:    mov    %r12,0x48(%rsp)
   0x00007ffff74f5340 <+64>:    mov    %r8d,0x50(%rsp)
   0x00007ffff74f5345 <+69>:    mov    %r9,0x58(%rsp)
   0x00007ffff74f534a <+74>:    movq   $0x1,0x138(%rsp)
   0x00007ffff74f5356 <+86>:    movq   $0x1,0x150(%rsp)
   0x00007ffff74f5362 <+98>:    movq   $0x1,0x168(%rsp)
   0x00007ffff74f536e <+110>:   movq   $0x1,0x140(%rsp)
   0x00007ffff74f537a <+122>:   movq   $0x1,0x158(%rsp)
   0x00007ffff74f5386 <+134>:   movq   $0x1,0x170(%rsp)
   0x00007ffff74f5392 <+146>:   movq   $0x1,0x148(%rsp)
   0x00007ffff74f539e <+158>:   movq   $0x1,0x160(%rsp)
   0x00007ffff74f53aa <+170>:   movq   $0x1,0x178(%rsp)
```

abergeron: Ok. I'm good.

abergeron:
Is there anything missing from this?

obilaniu:
@abergeron Yes; The codegen breaks somewhat at random depending on the kernel's required list of arguments and the tensor dimensionalities involved. The number of cases is large. Where it happens is in cases like generating a `,` to separate arguments to a generated function, but a part of the argument list is empty (e.g. because the `dst` tensor is 0-dim, or the kernel doesn't need a `dstArg` argument, ...), so you get consecutive commas and an angry runtime-compile error.

abergeron:
Ok

obilaniu:
@abergeron @nouiz You're going to be very happy:
- **ALL** reductions are now implemented, using the large code model.
- **ALL** testcases pass.
- The code is a bit cleaner after heavy-duty refactoring of the code-gen functions,
- .. which required that I add a new utility `util/srcgen.h` to make printing lists of arguments or summands easier,
- ... which required that I add `strb_appendv()`, which is to `strb_appendf()` as `vfprintf()` is to `fprintf()`.

Now, the Jenkins bot apparently has an older `libcheck` than Travis, because I'm using `ck_assert_double_eq_tol()` and Jenkins vomits all over that while everything is just fine with Travis.

In several places there are notices `/* BUG: Small code model not implemented */`. At those places (currently unreachable) I must add code-gen for the small code model. This will require great care and some help from the CLUDA language in providing `atomic*()` portably.

Similarly, for the pre-scalar/post-scalar op fusions there are notices at the appropriate places for where they will hook in.

src/util/srcgen.h
```diff
@@ -0,0 +1,106 @@
...
+
+
+/* Data Structure Prototypes & Typedefs */
+struct srcb;
+typedef struct srcb srcb;
+
+
+
+/* Enumerations */
+enum srcb_state{
```
abergeron: Space before `{`.

src/util/srcgen.h
```diff
@@ -0,0 +1,106 @@
...
+	srcb_state  state;
+	int         numElems;
+	const char* sep;
+	const char* empty;
+};
+
+
+
+/* Functions */
+static inline  void srcbInit  (srcb* s, strb* sb){
```
abergeron: No spaces between name and `(` and space before `{`.

src/util/srcgen.h
```diff
@@ -0,0 +1,106 @@
...
+	s->numElems   = 0;
+}
+static inline  void srcbBeginList(srcb* s, const char* sep, const char* empty){
+	s->state      = SRCB_STATE_INLIST;
+	s->numElems   = 0;
+	s->sep        = sep;
+	s->empty      = empty;
+}
+static inline  void srcbEndList(srcb* s){
+	if(s->numElems == 0){
```
abergeron: Space after `if` and space before `{`.

src/util/strb.h
```diff
@@ -77,6 +78,15 @@ static inline int strb_error(strb *sb) {
   return sb->l == (size_t)-1;
 }
 
+/*
+ * Initialize at runtime an strb.
+ */
+
+static inline void strb_init(strb* sb){
```
abergeron: Space before `{`.

src/util/strb.c
```diff
@@ -34,24 +34,35 @@ int strb_grow(strb *sb, size_t n) {
...
+  apSave = ap;
+  s = _vscprintf(f, apSave);
 #else
-  s = vsnprintf(NULL, 0, f, ap);
+  va_copy(apSave, ap);
+  s = vsnprintf(NULL, 0, f, apSave);
 #endif
-  va_end(ap);
-
+  va_end(apSave);
```
abergeron: Move the va_end inside the else branch.

src/gpuarray/reduction.h
```diff
@@ -0,0 +1,157 @@
...
+                                          unsigned        reduxLen,
+                                          const unsigned* reduxList);
+GPUARRAY_PUBLIC int GpuArray_all         (GpuArray*       dst,
+                                          const GpuArray* src,
+                                          unsigned        reduxLen,
+                                          const unsigned* reduxList);
+GPUARRAY_PUBLIC int GpuArray_any         (GpuArray*       dst,
+                                          const GpuArray* src,
+                                          unsigned        reduxLen,
+                                          const unsigned* reduxList);
```
abergeron: All these functions could (and probably should) be macros.

src/gpuarray_reduction.c
```diff
@@ -12,115 +12,712 @@
...
+ * @param [in]  typecode  Typecode of the type whose initializer is to be
+ *                        requested.
+ * @param [out] property  A pointer to a string. On return it will be set to
+ *                        the initializer expression.
+ * @return Zero if successful; Non-zero if the datatype is not supported.
+ */
+
+static int   reduxGetSumInit               (int typecode, const char** property){
+	if (typecode == GA_POINTER ||
+	    typecode == GA_BUFFER){
```
abergeron: You can use typecode < 0.  The "special" values will always be below 0.

src/gpuarray_reduction.c
```diff
@@ -12,115 +12,712 @@
...
+		case GA_FLOAT:
+		case GA_DOUBLE:
+		case GA_QUAD:
+		  *property = "NAN";
+		break;
+		default:
+		  return GA_UNSUPPORTED_ERROR;
+	}
+
+	return GA_NO_ERROR;
```
abergeron: I think that an additional property on the types for min/max values would be better than this.

src/gpuarray/reduction.h
obilaniu: @abergeron One thing that I've really liked about the current design is that I can breakpoint one of these functions without breakpointing on any others. It was very helpful when debugging using the test-suite.

abergeron: That is true, but it costs in code size and number of exported symbols.

For debugging you can always use conditional breakpoints on the op to only stop for those you are interested in.  I don't really mind making debugging a bit harder here.

src/gpuarray_reduction.c
obilaniu: @abergeron Doable by hacking the Python generator script a bit.

src/util/strb.c
obilaniu: The `va_end()` is in the correct place.
- `va_copy()`ing a `va_list` is equivalent to `va_start()`ing it followed by an equivalent number of `va_arg()` calls.
- Every `va_start()` must be matched by a `va_end()`.
- `va_copy()` doesn't exist on older Visual Studios, but in the Visual Studios where it's implemented it's implemented as just a straight assignment.

Therefore `va_end()` should be unconditional or else it's UB (although on x86-64 it's actually a no-op)

abergeron: It is *technically* UB to just assign and reuse va_lists like you did.

If we start from the assumption that a va_list is just a pointer on MSVC and we can just copy it then let's see it through and not use va_end either since it doesn't matter.

obilaniu:
@abergeron @nouiz @lamblin Great success.

Code now passes all tests (apparently deterministically, since no failures in multiple tries) on my machine and `leto13`. But still waiting for Travis results to triple-check.

Key points:

- My kernels are fully deterministic given:
  - An identical number of SMs
  - Identical max local workgroup size
  - Type, dimensions and strides of all input/output tensors.
- For all-reduction of a 100M-element float32 vector to a scalar:
  - Global memory bandwidth:
    - On my laptop's GTX 765M, I achieve 86% of the max in-practice bandwidth (~46.6 GB/s).
    - cuDNN does get about 100% though (53 GB/s).
    - Apart from this best-case, Nvidia Visual Profiler reports that a dozen of my generated kernels get in excess of 20 GB/s, most get > 10 GB/s and a few stragglers get > 3 GB/s.
  - Workspace:
    - For that same reduction, however, I use **exactly**  `2*sizeof(TK0)*BlocksSpawned*D` bytes of memory, where, in this case, `TK0` is `float32`, `BlocksSpawned` is `64*NUMBER_OF_SMs` and `D` is the number of writebacks per block, which in this case is `1`. On a GTX765M, this results in 256 blocks of 1024/4=256 threads being spawned, and the workspace size is just **2KB**. My workspace size is thus **O(k) in the number k of simultaneously-executing threads on the GPU**.
    - But cuDNN requests **400MB** of workspace for the same reduction.
- cuDNN is limited to tensors with no more than `2**31 - 1` elements and strides (measured in elements) representable in a 32-bit signed integer. I am almost completely unconstrained; Practically the only assumption I make is that the tensor will have at most one axis of length > `2**31 - 1`, while the strides (measured in bytes) can be 64-bit integers.
- I enforce a max block size 4 times smaller than the maximum allowed since that gives better performance.
- You may customize almost all types used in my kernel through a new `GpuReductionAttr` API.
- It is in principle possible to add arbitrary pre/post scalar operations. In fact, my code is capable of generalizing `GpuElemwise`.

src/gpuarray_reduction.c
```diff
@@ -6,841 +6,4095 @@
...
+ *                    Reduction Operator.
+ * 
+ * INTRO
+ * 
+ * Generates the source code for a reduction kernel over arbitrarily-ranked,
+ * -shaped and -typed tensors.
+ * 
+ * It is assumed that at most one axis will ever be of length > 2**31-1. The
+ * assumption is believed safe because no GPU or similar accelerator presently
+ * on Earth has the capacity to store or process 2**62-element tensors.
```
obilaniu: @abergeron When a user broadcasts a tensor on two or more axes, each of which are of length >= 2**31, he implicitly accepts a computational cost of 2**62 FLOP's. The most powerful GPUs on Earth right now have a throughput of 1e10 `float32` additions or multiplications per second, and so would chug through such a tensor of 2**62 elements in a _mere_ 15 years. I'm unaware of a realistic environment and usecase for a 15-year computation.

src/gpuarray_reduction.c
```diff
@@ -6,841 +6,4095 @@
...
+ */
+
+static int         reduxGenInferProperties          (GpuReduction*        gr){
+	int i;
+	
+	/**
+	 * Source code buffer preallocation failed?
+	 */
+	
+	if (strb_ensure(&gr->s, 32*1024) != 0){
```
obilaniu: I generate 8 different versions of the core loops depending on where the "split" and "huge" axes are, so the generated source code does get a bit voluminous. Could be made smaller with macros.

src/gpuarray_reduction.c
```diff
@@ -6,841 +6,4095 @@
...
+ *     TPS0 = promotion(TS0)
+ * 
+ * Therefore, it is currently always the case that TK0 == TPS0.
+ * 
+ * In the future this might become wierder when the accumulator is a Kahan
+ * summation, for instance, and then TK0 != promoted(TS0).
+ * 
+ * If the user guaranteed to us through gr->grAttr that TK1 can be made
+ * narrower than 64-bit, this is also where we'd take this into account.
+ * For now we default TK1 to exactly TI0.
```
obilaniu: @abergeron See above comment, but also note that for this kernel I personally assume responsibility for selecting 32/64-bit-length code versions. It's what the `selector` does - It selects one of 8 possible "templates" for Phase 0, or it selects the one version of the fixup code in Phase 1.

src/gpuarray_reduction.c
```diff
@@ -6,841 +6,4095 @@
...
+	for (k=0;k < gr->nds;k++){
+		fn(gr, gr->TU64tc, "TU64",                              "L%d",         k, user);
+	}
+	for (k=0;k < gr->log2MaxBS;k++){
+		fn(gr, gr->TU32tc, "TU32",                              "L%di",        k, user);
+	}
+	
+	/* S0 Lattice */
+	if (reduxGenKernelRequiresLatticeS0(gr)){
+		fn(gr, GA_BUFFER,  "const GLOBAL_MEM char* restrict",   "S0",          0, user);
```
obilaniu: OpenCL 1.0, _§6.1.4 Reserved Data Types_: _"The type qualifiers `const`, `restrict` and `volatile` as defined by the C99 specification are supported. These qualifiers cannot be used with `image2d_t` and `image3d_t` type. Types other than pointer types shall not use the `restrict` qualifier."_

src/gpuarray_reduction.c
```diff
@@ -6,841 +6,4095 @@
...
+	grAttr->maxDstDims = maxDstDims;
+	
+	return GA_NO_ERROR;
+}
+GPUARRAY_PUBLIC int   GpuReductionAttr_sets0type    (GpuReductionAttr*          grAttr,
+                                                     int                        s0Typecode){
+	switch (grAttr->op){
+		case GA_REDUCE_AND:
+		case GA_REDUCE_OR:
+		case GA_REDUCE_XOR:
```
obilaniu: Good idea. Perhaps a better pattern would have been a public `GpuReductionAttr_validate()` function that validates the sanity of all parameters as a whole. I was just too tired to implement such a thing along with logic to detect whether a mandatory reduction attribute has been set or not.

src/gpuarray_reduction.c
```diff
@@ -6,841 +6,4095 @@
...
+		(*gr)->ndd    = (int)grAttr->maxDstDims;
+		(*gr)->ndr    = (int)(grAttr->maxSrcDims-grAttr->maxDstDims);
+		
+		return reduxGenInit(*gr);
+	}else{
+		return GA_MEMORY_ERROR;
+	}
+}
+GPUARRAY_PUBLIC void  GpuReduction_free             (GpuReduction*              gr){
+	reduxGenCleanup(gr, !GA_NO_ERROR);
```
obilaniu: This was an idiotic attempt at freeing the `GpuReduction*` by invoking its cleanup code with an error code, thus pretending its initialization had failed. `reduxGenCleanup()` frees everything but but only if `ret != GA_NO_ERROR`.

Admittedly this could be done better.

src/gpuarray_reduction.c
```diff
@@ -6,841 +6,4095 @@
...
+		break;
+		case GA_HALF:
+		case GA_FLOAT:
+		case GA_DOUBLE:
+		case GA_QUAD:
+		  *property = "NAN";
+		break;
+		default:
+		  return GA_UNSUPPORTED_ERROR;
+	}
```
obilaniu: I haven't tested OpenCL at all since it isn't properly set up on my machine. I've no idea whether the `<type>n` stuff works at all, but I see no obvious reason why they wouldn't work, at least for the arithmetic reductions. Although I agree it might get wierd for e.g. MaxAndArgmax (what does it mean that a `double8` is bigger than another?)

In general the typecode management is a weak point of my code right now.

src/gpuarray_reduction.c
```diff
@@ -6,841 +6,4095 @@
...
+	return axisGetD1Stride(axis)<0 ? -(size_t)axisGetD1Stride(axis):
+	                                     +(size_t)axisGetD1Stride(axis);
+}
+static size_t      axisGetI0Stride                  (const axis_desc*     axis){
+	return axis->i0S;
+}
+static void        axisSetI0Stride                  (axis_desc*           axis,
+                                                     size_t               i0S){
+	axis->i0S = i0S;
+}
```
obilaniu: Some getters and setters are more justified than others, but my code went through several iterations where certain attributes were computed automatically based on others, and this flipped often enough that I made functions to protect the rest of the code from that concern. I'll have to extirpate many of them now that the design is pretty much frozen.

There are also various places in the code I chose to stick with getters mostly because I liked the uniformity of all attribute accesses going through a getter.

Could absolutely be changed.

obilaniu:
@abergeron In case this interests you I rebased against `master`, so there are now no conflicts and the codebase now builds, runs and passes tests again.

On Jenkins I see that for some reason or other `ck_assert_ptr_nonnull()` no longer works properly. It did use to, AFAIK. Was `libcheck` downgraded? For now I simply re-`#define` it to another `libcheck` API, but I'm still surprised.

Given your fp16-related changes I may have to review certain data-access macros, but otherwise this PR is still as good as it was last month.

PR #335, opened 2017-01-25
Disk cache

abergeron:
This add an on-disk cache for kernels which speeds up recompilation.

From my light testing it shaves off about 1 second of a 3 seconds program that compiles an elemwise kernel and runs it, which means that it should be a net gain in most situations.

The cache is disabled by default and enabled when the environment variable GPUARRAY_CACHE_PATH is set to some path.  The cache should be fine for use over NFS with multiple processes.  Entries that were writing while a system crash happens can get corrupted or lost, but the cache should recover by ignoring and replacing bad entries. There is currently no built-in pruning mechanism, but there is a python script which can act as such.  It is safe to run it while other processes are accessing the cache.

If there is a problem accessing the given cache path, the cache will be disabled.

This still needs testing on Windows and macOS to make sure it works.  Especially on windows the semantics for touching files that are opened by other processes is significantly different, which may require adjustment.

Finally DO NOT MERGE until the 0.6 release is out the door.  This can be part of an eventual 0.6.1 release when it has been properly tested.

### DIFF
--- a/bin/gpuarray-cache
+++ b/bin/gpuarray-cache
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+
+import os
+
+def clean(max_size):
+    content = []
+    for root, dirs, files in os.walk(os.environ.get('GPUARRAY_CACHE',
+                                                    '~/.gpuarray/cache/')):
+        for file in files:
+            fpath = os.path.join(root, file)
+            st = os.stat(fpath)
+            content.append((st.st_atime, st.st_size, fpath))
+
+    content.sort()
+    cur_size = 0
+    for _, size, path in content:
+        cur_size += size
+        if cur_size > max_size:
+            os.remove(path)
+
+
+SUFFIXES = {'B': 1, 'K': 1 << 10, 'M': 1 < 20, 'G': 1 << 30, 'T': 1 << 40,
+            'P': 1 << 50, 'E': 1 << 60, 'Z': 1 << 70, 'Y': 1 << 80}
+
+
+def get_size(s):
+    i = 0
+    while i < len(s) and (s[i].isdigit() or s[i] == '.'):
+        i += 1
+    num = s[:i]
+    suf = s[i:]
+    num = float(num)
+    if suf != "":
+        letter = suf.strip().upper()
+        if letter not in SUFFIXES:
+            raise ValueError("can't interpret %r" % init)
+        mult = SUFFIXES[letter]
+    else:
+        mult = 0
+    return int(num * mult)
+
+
+if __name__ == '__main__':
+    import argparse
+
+    parser = argparse.ArgumentParser(description='libgpuarray cache maintenance utility')
+    parser.add_argument('-s', '--max_size', help='Set the maximum size for pruning')
+    args = parser.parse_args()
+
+    clean(get_size(args.max_size))
+
--- a/bin/gpuarray-cache
+++ b/bin/gpuarray-cache
@@ -1,11 +1,11 @@
 #!/usr/bin/env python
 
 import os
+import sys
 
-def clean(max_size):
+def clean(max_size, path):
     content = []
-    for root, dirs, files in os.walk(os.environ.get('GPUARRAY_CACHE',
-                                                    '~/.gpuarray/cache/')):
+    for root, dirs, files in os.walk(path):
         for file in files:
             fpath = os.path.join(root, file)
             st = os.stat(fpath)
@@ -25,18 +25,18 @@ SUFFIXES = {'B': 1, 'K': 1 << 10, 'M': 1 << 20, 'G': 1 << 30, 'T': 1 << 40,
 
 def get_size(s):
     i = 0
-    while i < len(s) and (s[i].isdigit() or s[i] == '.'):
-        i += 1
-    num = s[:i]
-    suf = s[i:]
+    s = s.strip()
+    if s[-1].upper() in SUFFIXES:
+        num = s[:-1]
+        suf = s[-1].upper()
+    else:
+        num = s
+        suf = ""
     num = float(num)
     if suf != "":
-        letter = suf.strip().upper()
-        if letter not in SUFFIXES:
-            raise ValueError("can't interpret %r" % init)
-        mult = SUFFIXES[letter]
+        mult = SUFFIXES[suf]
     else:
-        mult = 0
+        mult = 1
     return int(num * mult)
 
 
@@ -46,6 +46,10 @@ if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='libgpuarray cache maintenance utility')
     parser.add_argument('-s', '--max_size', help='Set the maximum size for pruning (in bytes with suffixes: K, M, G, ...)')
     args = parser.parse_args()
+    path = os.environ.get('GPUARRAY_CACHE_PATH', None)
+    if path is None:
+        print("You need to set GPUARRAY_CACHE_PATH so that this programs knows which path to clean.")
+        sys.exit(1)
 
-    clean(get_size(args.max_size))
+    clean(get_size(args.max_size), path)
 
--- a/bin/gpuarray-cache
+++ b/bin/gpuarray-cache
@@ -19,7 +19,7 @@ def clean(max_size):
             os.remove(path)
 
 
-SUFFIXES = {'B': 1, 'K': 1 << 10, 'M': 1 < 20, 'G': 1 << 30, 'T': 1 << 40,
+SUFFIXES = {'B': 1, 'K': 1 << 10, 'M': 1 << 20, 'G': 1 << 30, 'T': 1 << 40,
             'P': 1 << 50, 'E': 1 << 60, 'Z': 1 << 70, 'Y': 1 << 80}
 
 
@@ -44,7 +44,7 @@ if __name__ == '__main__':
     import argparse
 
     parser = argparse.ArgumentParser(description='libgpuarray cache maintenance utility')
-    parser.add_argument('-s', '--max_size', help='Set the maximum size for pruning')
+    parser.add_argument('-s', '--max_size', help='Set the maximum size for pruning (in bytes with suffixes: K, M, G, ...)')
     args = parser.parse_args()
 
     clean(get_size(args.max_size))
--- a/make.bat
+++ b/make.bat
@@ -1,0 +1,6 @@
+del bld
+mkdir bld
+cd bld
+cmake .. -G "NMake Makefiles"
+cmake --build . --config Release
+cd ..
--- a/make.bat
+++ b/make.bat
@@ -1,3 +1,7 @@
+REM This helps repetitive builds on windows
+REM It needs the compiler you want to use to be available in the shell
+REM and it will build a release version
+
 del bld
 mkdir bld
 cd bld
--- a/make.bat
+++ b/make.bat
@@ -1,6 +1,6 @@
 del bld
 mkdir bld
 cd bld
-cmake .. -G "NMake Makefiles"
+cmake .. -G "NMake Makefiles" -DCMAKE_BUILD_TYPE=Release
 cmake --build . --config Release
 cd ..
--- a/src/cache/disk.c
+++ b/src/cache/disk.c
@@ -8,15 +8,9 @@
 
 #include "cache.h"
 #include "private_config.h"
-#include "util/strb.h"
 #include "util/skein.h"
 
 #define HEXP_LEN (128 + 2)
-
-typedef int (*kwrite_fn)(strb *res, cache_key_t key);
-typedef int (*vwrite_fn)(strb *res, cache_value_t val);
-typedef cache_key_t (*kread_fn)(const strb *b);
-typedef cache_value_t (*vread_fn)(const strb *b);
 
 typedef struct _disk_cache {
   cache c;
--- a/src/cache/disk.c
+++ b/src/cache/disk.c
@@ -1,0 +1,301 @@
+#include <dirent.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <sys/time.h>
+#include <sys/stat.h>
+
+#include "cache.h"
+#include "private_config.h"
+#include "util/strb.h"
+#include "util/skein.h"
+
+#define HEXP_LEN (128 + 2)
+
+typedef int (*kwrite_fn)(strb *res, cache_key_t key);
+typedef int (*vwrite_fn)(strb *res, cache_value_t val);
+typedef cache_key_t (*kread_fn)(const strb *b);
+typedef cache_value_t (*vread_fn)(const strb *b);
+
+typedef struct _disk_cache {
+  cache c;
+  cache * mem;
+  kwrite_fn kwrite;
+  vwrite_fn vwrite;
+  kread_fn kread;
+  vread_fn vread;
+  int dirfd;
+} disk_cache;
+
+
+static unsigned long long ntohull(const char *in) {
+  return ((unsigned long long)in[0] << 56 | (unsigned long long)in[1] << 48 |
+          (unsigned long long)in[2] << 40 | (unsigned long long)in[3] << 32 |
+          (unsigned long long)in[4] << 24 | (unsigned long long)in[5] << 16 |
+          (unsigned long long)in[6] << 8 | (unsigned long long)in[7]);
+}
+
+static void htonull(unsigned long long in, char *out) {
+  out[0] = in >> 56;
+  out[1] = in >> 48;
+  out[2] = in >> 40;
+  out[3] = in >> 32;
+  out[4] = in >> 24;
+  out[5] = in >> 16;
+  out[6] = in >> 8;
+  out[7] = in;
+}
+
+static int mkstempat(int dfd, char *template) {
+  static const char letters[] =
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+  size_t length;
+  char *XXXXXX;
+  struct timeval tv;
+  unsigned long long  randnum, working;
+  int i, tries, fd;
+
+  length = strlen(template);
+  if (length < 6) {
+    errno = EINVAL;
+    return -1;
+  }
+  XXXXXX = template + length - 6;
+  if (strcmp(XXXXXX, "XXXXXX") != 0) {
+    errno = EINVAL;
+    return -1;
+  }
+
+  /* This is kind of crappy, but the point is to not step on each
+     other's feet */
+  gettimeofday(&tv, NULL);
+  randnum = ((unsigned long long) tv.tv_usec << 16) ^ tv.tv_sec ^ getpid();
+
+  for (tries = 0; tries < TMP_MAX; tries++) {
+    for (working = randnum, i = 0; i < 6; i++) {
+      XXXXXX[i] = letters[working % 62];
+      working /= 62;
+    }
+    fd = openat(dfd, template, O_RDWR | O_CREAT | O_EXCL, 0600);
+    if (fd >= 0 || (errno != EEXIST && errno != EISDIR))
+      return fd;
+
+    randnum += (tv.tv_usec >> 10) & 0xfff;
+  }
+  errno = EEXIST;
+  return -1;
+}
+
+static int key_path(disk_cache *c, const cache_key_t key, char *out) {
+  strb kb = STRB_STATIC_INIT;
+  unsigned char hash[64];
+  int i;
+
+  if (c->kwrite(&kb, key)) return -1;
+  if (Skein_512((unsigned char *)kb.s, kb.l, hash)) return -1;
+  if (snprintf(out, 6, "%02x%02x/%02x%02x",
+               hash[0], hash[1], hash[2], hash[3]) != 5)
+    return -1;
+  for (i = 4; i < 64; i += 4) {
+    if (snprintf(out+(i * 2 + 1), 9, "%02x%02x%02x%02x",
+                 hash[i], hash[i+1], hash[i+2], hash[i+3]) != 8)
+      return -1;
+  }
+  return 0;
+}
+
+static int write_entry(disk_cache *c, const cache_key_t k,
+                       const cache_value_t v) {
+  char hexp[HEXP_LEN];
+  char tmp_path[] = "tmp.XXXXXXXX";
+  strb b = STRB_STATIC_INIT;
+  size_t kl, vl;
+  int fd, err;
+
+  if (key_path(c, k, hexp)) return -1;
+
+  if (!strb_ensure(&b, 16)) return -1;
+  b.l = 16;
+  c->kwrite(&b, k);
+  kl = b.l - 16;
+  c->vwrite(&b, v);
+  vl = b.l - kl - 16;
+  htonull(kl, b.s);
+  htonull(vl, b.s + 8);
+  if (strb_error(&b)) {
+    strb_clear(&b);
+    return -1;
+  }
+
+  fd = mkstempat(c->dirfd, tmp_path);
+  if (fd == -1) {
+    strb_clear(&b);
+    return -1;
+  }
+
+  err = strb_write(fd, &b);
+  strb_clear(&b);
+  close(fd);
+  if (err) {
+    unlinkat(c->dirfd, tmp_path, 0);
+    return -1;
+  }
+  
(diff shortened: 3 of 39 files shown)


src/util/skein.c
```diff
@@ -0,0 +1,309 @@
+/***********************************************************************
```
nouiz: I didn't review this commit.

src/gpuarray/buffer.h
```diff
@@ -328,9 +328,9 @@ GPUARRAY_PUBLIC int gpudata_move(gpudata *dst, size_t dstoff,
  * \returns the new buffer in dst_ctx or NULL if no efficient way to
  *          transfer could be found.
  */
-GPUARRAY_LOCAL int gpudata_transfer(gpudata *dst, size_t dstoff,
-                                    gpudata *src, size_t srcoff,
-                                    size_t sz);
+GPUARRAY_PUBLIC int gpudata_transfer(gpudata *dst, size_t dstoff,
```
nouiz: What is the consequense of this? SHould we bump version number? How this affect Theano?

src/gpuarray/config.h
```diff
@@ -12,19 +12,15 @@
   #else
    #define GPUARRAY_PUBLIC __declspec(dllimport)
   #endif
-  #define GPUARRAY_LOCAL
```
nouiz: Should this commit and the previous should be included in 0.6? If so, make a new PR for this.

bin/gpuarray-cache
```diff
@@ -0,0 +1,51 @@
...
+    else:
+        mult = 0
+    return int(num * mult)
+
+
+if __name__ == '__main__':
+    import argparse
+
+    parser = argparse.ArgumentParser(description='libgpuarray cache maintenance utility')
+    parser.add_argument('-s', '--max_size', help='Set the maximum size for pruning')
```
nouiz: The help isn't clear. Does it mean we keep that number of element? Does this take into account the date? creation/usage date?

bin/gpuarray-cache
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
+
+import os
+
+def clean(max_size):
+    content = []
+    for root, dirs, files in os.walk(os.environ.get('GPUARRAY_CACHE',
+                                                    '~/.gpuarray/cache/')):
```
nouiz: Why set a default here when if GPUARRAY_CACHE isn't set, we don't have a default?

If it work well on Linux with NFS, I would enable it by default on Linux. Until it is more tested on mac and windows, we should wait for those platform.

bin/gpuarray-cache
nouiz: I understood the size as number of element. You probably mean by disk space. Be more clear and give the unit.

bin/gpuarray-cache
```diff
@@ -0,0 +1,51 @@
...
+
+    content.sort()
+    cur_size = 0
+    for _, size, path in content:
+        cur_size += size
+        if cur_size > max_size:
+            os.remove(path)
+
+
+SUFFIXES = {'B': 1, 'K': 1 << 10, 'M': 1 < 20, 'G': 1 << 30, 'T': 1 << 40,
```
nouiz: M is missing a <

bin/gpuarray-cache
```diff
@@ -0,0 +1,51 @@
...
+            'P': 1 << 50, 'E': 1 << 60, 'Z': 1 << 70, 'Y': 1 << 80}
```
nouiz: I would not note this scrip to survive up to Y size for the cache alone! Talk about being future proof!

bin/gpuarray-cache
```diff
@@ -0,0 +1,51 @@
...
+
+
+def get_size(s):
+    i = 0
+    while i < len(s) and (s[i].isdigit() or s[i] == '.'):
+        i += 1
```
nouiz: This is too complicated then needed, but it work. Just doing if s[-1] in "BKM"... would do it.

src/gpuarray_buffer_cuda.c
```diff
@@ -49,18 +49,72 @@ static int cuda_records(gpudata *, int, CUstream);
...
+    return NULL;
+  }
+  memcpy(k->bin_id, b->s, 64);
+  strb_appendn(k->src, b->s+64, b->l-64);
+  return k;
+}
+
+static int kernel_write(strb *res, strb *bin) {
+  strb_appendb(res, bin);
+  return strb_error(res);
```
nouiz: What it should return? Normally a write tell the number of bytes written. I don't think this is what happen here.

src/gpuarray_buffer_cuda.c
```diff
@@ -49,18 +49,72 @@ static int cuda_records(gpudata *, int, CUstream);
...
+}
+
+static strb *kernel_read(const strb *b) {
```
nouiz: I don't get what this do. Is the name correct? What it from from? It look more like a copy method then a read method.

src/gpuarray_buffer_cuda.c
```diff
@@ -152,11 +208,38 @@ cuda_context *cuda_make_ctx(CUcontext ctx, int flags) {
...
+                                 (cache_eq_fn)strb_eq,
+                                 (cache_hash_fn)strb_hash,
                                  (cache_freek_fn)strb_free,
                                  (cache_freev_fn)cuda_freekernel);
   if (res->kernel_cache == NULL)
     goto fail_cache;
+
+  cache_path = getenv("GPUARRAY_CACHE_PATH");
+  if (cache_path != NULL) {
+    mem_cache = cache_lru(64, 8,
```
nouiz: Can you explain this mem cache? What it is for?

src/private_cuda.h
```diff
@@ -68,6 +68,7 @@ typedef struct _cuda_context {
   CUstream mem_s;
   gpudata *freeblocks;
   cache *kernel_cache;
+  cache *disk_cache;
```
nouiz: Why in the context? Do you have a sub path per arch?

src/gpuarray/error.h
```diff
@@ -36,6 +36,7 @@ enum ga_error {
   GA_COMM_ERROR,
   GA_XLARGE_ERROR,
   GA_LOAD_ERROR,
+  GA_DEPRECATED_ERROR,
```
nouiz: This will request a version bump somewhere.

src/gpuarray_buffer_cuda.c
```diff
@@ -996,22 +996,22 @@ static int detect_arch(const char *prefix, char *ret, CUresult *err) {
   return GA_NO_ERROR;
 }
 
-static void *call_compiler(const char *src, size_t len, const char *arch_arg,
-                           size_t *bin_len, char **log, size_t *log_len,
-                           int *ret) {
+static int call_compiler(cuda_context *ctx, strb *src, strb *ptx, strb *log) {
```
nouiz: version bump.

src/gpuarray_buffer_cuda.c
```diff
@@ -1020,41 +1020,115 @@ static void *call_compiler(const char *src, size_t len, const char *arch_arg,
...
 
-  buf = malloc(buflen);
-  if (buf == NULL) {
-    nvrtcDestroyProgram(&prog);
-    FAIL(NULL, GA_MEMORY_ERROR);
+  if (strb_ensure(ptx, buflen) == 0)
+    err = nvrtcGetPTX(prog, ptx->s+ptx->l);
+
+end:
+  nvrtcDestroyProgram(&prog);
```
nouiz: Now you always destroy the program. Why? It the past it was only in case of error.

src/gpuarray_buffer_cuda.c
```diff
@@ -1789,7 +1783,6 @@ const gpuarray_buffer_ops cuda_ops = {cuda_get_platform_count,
                                       cuda_freekernel,
                                       cuda_kernelsetarg,
                                       cuda_callkernel,
-                                      cuda_kernelbin,
```
nouiz: I think this break the ABI.

src/gpuarray/buffer.h
abergeron: This might warrant a minor bump because there are new functions exposed in the API, yes.

src/gpuarray/config.h
abergeron: It's mostly a cosmetic change since we use -fvisibility=hidden anyway.

bin/gpuarray-cache
abergeron: It should work well on mac too, it's windows I have more doubts about.

Also, maybe there should be a default, I just didn't put one in the C code.

bin/gpuarray-cache
abergeron: I want to be really sure :)

bin/gpuarray-cache
abergeron: ok

bin/gpuarray-cache
abergeron: ok

abergeron: Although we could do number of elements too.  Or age.  Or whatever.

src/gpuarray_buffer_cuda.c
abergeron: This is not really a write.  It just appends to the strb buffer given.  So it returns if it worked or not.

src/gpuarray_buffer_cuda.c
abergeron: This is an internal function so it has no impact on the API/ABI

src/gpuarray_buffer_cuda.c
abergeron: In this case it does more or less a copy.  The point is that it has to fit into an API that asks it to read data from the input strb and deserialize the data.  It just happens that the data in this case is another strb.

src/gpuarray_buffer_cuda.c
abergeron: No I moved that list to stritcly internal to avoid ABI breaks when it changes.

src/private_cuda.h
abergeron: No, but I don't want to introduce contention on cache access and global locks.

src/gpuarray/error.h
abergeron: Yes, an API bump probably.

src/gpuarray_buffer_cuda.c
abergeron: It's to have an in-memory read cache for the disk cache.

abergeron:
I finally got the Windows code working!

My test case of importing Theano (which compiles 6 blas kernels) goes from 5 to 2 seconds total time with this on Windows. On linux we go from 10 to 4 seconds (not the same machine).

There are still some details to tweak such as possibly setting a default value for the cache path and some code cleanup.

src/gpuarray_buffer_cuda.c
```diff
@@ -937,41 +1021,123 @@ static void *call_compiler(const char *src, size_t len, const char *arch_arg,
...
-    buf = malloc(buflen);
-    if (buf == NULL) goto end2;
-    err2 = nvrtcGetProgramLog(prog, (char *)buf);
-    if (err2 != NVRTC_SUCCESS) goto end2;
-    if (log_len != NULL) *log_len = buflen;
-    *log = (char *)buf;
-    buf = NULL;
+  if (nvrtcGetProgramLogSize(prog, &buflen) == NVRTC_SUCCESS) {
+    strb_appends(log, "NVRTC compile log::\n");
+    if (strb_ensure(log, buflen) == 0)
```
nouiz: This isn't enough. It should be buflen + 1 for line 1029

src/gpuarray_buffer_cuda.c
```diff
@@ -937,41 +1021,123 @@ static void *call_compiler(const char *src, size_t len, const char *arch_arg,
...
   err = nvrtcGetPTXSize(prog, &buflen);
   if (err != NVRTC_SUCCESS) goto end;
 
-  buf = malloc(buflen);
-  if (buf == NULL) {
-    nvrtcDestroyProgram(&prog);
-    FAIL(NULL, GA_MEMORY_ERROR);
+  if (strb_ensure(ptx, buflen) == 0) {
+    err = nvrtcGetPTX(prog, ptx->s+ptx->l);
+    if (err == NVRTC_SUCCESS) ptx->l = buflen;
```
nouiz: I think it should be "ptx->l += buflen;". But in practice I think ptx->l will be zero most of the time. Still better safe then this hard to debug problem.

src/gpuarray_buffer_cuda.c
```diff
@@ -937,41 +1021,123 @@ static void *call_compiler(const char *src, size_t len, const char *arch_arg,
...
+  err = make_bin(ctx, &ptx, bin, log);
+  if (err != GA_NO_ERROR) return err;
+  if (ctx->disk_cache) {
+    pk = memdup(&k, sizeof(k));
+    if (pk == NULL)
+      return GA_NO_ERROR;
+    cbin = strb_alloc(bin->l);
+    if (cbin == NULL) {
+      free(pk);
+      return GA_NO_ERROR;
```
nouiz: I don't like that. The cache is just skipped and the user don't have any information about that.

src/gpuarray_buffer_cuda.c
```diff
@@ -937,41 +1021,123 @@ static void *call_compiler(const char *src, size_t len, const char *arch_arg,
...
+    }
+    strb_appendb(cbin, bin);
+    if (strb_error(cbin)) {
+      free(pk);
+      strb_free(cbin);
+      return GA_NO_ERROR;
```
nouiz: idem

src/gpuarray_buffer_cuda.c
```diff
@@ -1210,16 +1306,16 @@ static gpukernel *cuda_newkernel(gpucontext *c, unsigned int count,
...
     TAG_KER(res);
-    psb = memdup(&sb, sizeof(strb));
-    if (psb == NULL) {
-      cuda_freekernel(res);
-      strb_clear(&sb);
-      FAIL(NULL, GA_MEMORY_ERROR);
+    psrc = memdup(&src, sizeof(strb));
+    if (psrc != NULL) {
+      /* One of the refs is for the cache */
+      res->refcnt++;
```
nouiz: Why it isn't the cache that add the ref if it succed?

src/private_cuda.h
nouiz: Add that reason as a comment.

src/util/strb.h
```diff
@@ -159,7 +159,21 @@ static inline void strb_appendb(strb *sb, strb *sb2) {
  *
  * A format error will place the strb in error mode.
  */
-GPUARRAY_LOCAL void strb_appendf(strb *, const char *f, ...);
+void strb_appendf(strb *, const char *f, ...);
+
+/*
+ * Reads from the file specified by the given file descriptor.
```
nouiz: add that it read up to sz bytes.

src/util/strb.h
```diff
@@ -159,7 +159,21 @@ static inline void strb_appendb(strb *sb, strb *sb2) {
...
+ *
+ * A read error will place the strb in error mode.
+ */
+void strb_read(strb *, int fd, size_t sz);
```
nouiz: I think there is an error in this line that prevent compilation... why only "strb *," and no name for that parameter? Why does jenkins passed?

bin/gpuarray-cache
```diff
@@ -0,0 +1,51 @@
...
+    for root, dirs, files in os.walk(os.environ.get('GPUARRAY_CACHE',
+                                                    '~/.gpuarray/cache/')):
+        for file in files:
+            fpath = os.path.join(root, file)
+            st = os.stat(fpath)
+            content.append((st.st_atime, st.st_size, fpath))
+
+    content.sort()
+    cur_size = 0
+    for _, size, path in content:
```
abergeron: content.sort() does that since I store the tuples with (atime, size, path)

src/gpuarray_blas_opencl_clblas.c
```diff
@@ -100,7 +100,7 @@ static int sgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
     ARRAY_INIT(C[i]);
     err = clblasSgemm(convO(order), convT(transA), convT(transB), M, N, K,
                       alpha, A[i]->buf, offA[i], lda, B[i]->buf, offB[i], ldb,
-                      beta, C[i]->buf, offB[i], ldc, 1, &ctx->q,
+                      beta, C[i]->buf, offC[i], ldc, 1, &ctx->q,
```
abergeron: Yes, it's a bugfix.

make.bat
```diff
@@ -0,0 +1,6 @@
+del bld
```
abergeron: ok

src/gpuarray/buffer.h
abergeron: I'll do it.

bin/gpuarray-cache
abergeron: I simplified this code.

src/gpuarray_buffer_cuda.c
abergeron: We are cutting the final '\0' of the log at line 1028 so there will be enough space.

Also even if there wasn't enough space strb_appendc() will extend the buffer if required.

src/gpuarray_buffer_cuda.c
abergeron: Yes.

src/gpuarray_buffer_cuda.c
abergeron: Would you prefer to error out in case we can't add the kernel to the cache for some reason?

I can make that change.

src/gpuarray_buffer_cuda.c
abergeron: Because I would have to teach the cache how to add references (which is not possible for just any object) where as now I just have to teach it how to free objects.

src/private_cuda.h
abergeron: ok

src/util/strb.h
abergeron: ok

src/util/strb.h
abergeron: We don't need to specify parameter names on declaration, so this wouldn't be a problem.  However it jars with the style of the rest of the file so I'll fix it.

src/gpuarray_buffer_cuda.c
nouiz: Add on line 1028, that this is what that line does

src/gpuarray_buffer_cuda.c
nouiz: I would prefer a warning, but I think it isn't easy. So a memory error seem more appropriate as this is the only cause of this that I can think of.

src/gpuarray_buffer_cuda.c
nouiz: ok

src/cache/disk.c
```diff
@@ -0,0 +1,459 @@
...
+  cache * mem;
+  kwrite_fn kwrite;
+  vwrite_fn vwrite;
+  kread_fn kread;
+  vread_fn vread;
+  const char *dirp;
+} disk_cache;
+
+
+static unsigned long long ntohull(const char *_in) {
```
abergeron: Like ntohl but for unsigned long long (64 bits).

It converts from network byte order to host byte order.

src/cache/disk.c
```diff
@@ -0,0 +1,459 @@
...
+  out[1] = (unsigned char)(in >> 48);
+  out[2] = (unsigned char)(in >> 40);
+  out[3] = (unsigned char)(in >> 32);
+  out[4] = (unsigned char)(in >> 24);
+  out[5] = (unsigned char)(in >> 16);
+  out[6] = (unsigned char)(in >> 8);
+  out[7] = (unsigned char)(in);
+}
+
+static int catp(char *path, const char *dirp, const char *rpath) {
```
abergeron: ok

src/cache/disk.c
```diff
@@ -0,0 +1,459 @@
...
+    return -1;
+  }
+  if (strlcat(path, rpath, PATH_MAX) >= PATH_MAX) {
+    errno = ENAMETOOLONG;
+    return -1;
+  }
+  return 0;
+}
+
+static int openp(const char *dirp, const char *rpath, int flags, int mode) {
```
abergeron: Ok.

src/gpuarray_buffer_cuda.c
abergeron: ok

src/gpuarray_buffer_cuda.c
abergeron: With disk_cache it could also be a write error of some kind.

nouiz:
I have a segfault when I try this PR with the cache enabled, but not when
the cache isn't enabled. This is for the pygpu tests, the cache was reusing
the output of libgpuarray tests.

nouiz:
I also have the segfaul of pygpu tests with an empty cache. In that case
the segfaul happen earlier.

I let you investigate this more.

src/cache/disk.c
nouiz: The name is very cryptic. If you check htonull, this made me think we talk about null ptr!

Can you just add for both a 1 line comment that tell what you wrote here.

abergeron:
I had to rebase for some unrelated problems, but I found and fixed the cause of the segfault.

The whole test suite at once passes.  Just for fun I made a second run of the testsuite with the cache full and it took 52s instead of 700s.

abergeron:
What is remaining:
 - Decide if erroring out when there are problems adding to the cache is a good idea
 - Maybe add some versioning to the caching for debug/non-debug, new cuda releases.

abergeron:
Also, in release mode the tests go from 542 seconds to 30.

nouiz:
How does it handle NFS? I ran the libgpuarray and pygpu tests in different config and they pass. So the only thing left for me is to understand how NFS is handled

src/gpuarray_buffer_cuda.c
```diff
@@ -50,10 +50,17 @@ static int detect_arch(const char *prefix, char *ret, CUresult *err);
...
+  uint8_t version;
+  uint8_t debug;
+  uint8_t major;
+  uint8_t minor;
+  uint32_t reserved;
   char bin_id[64];
   strb src;
 } kernel_key;
 
+#define KERNEL_KEY_MM (sizeof(kernel_key) - sizeof(strb))
```
nouiz: What MM mean? Add this in comment.

nouiz:
With the cache, test_elemwise fail:
```
$ GPUARRAY_CACHE_PATH=~/.theano/gpuarray_cache/ DEVICE=cuda0 make test
Running tests...
      Start  1: test_types
 1/11 Test  #1: test_types .......................   Passed    0.01 sec
      Start  2: test_util
 2/11 Test  #2: test_util ........................   Passed    0.01 sec
      Start  3: test_util_integerfactoring
 3/11 Test  #3: test_util_integerfactoring .......   Passed    0.51 sec
      Start  4: test_reduction
 4/11 Test  #4: test_reduction ...................   Passed    5.91 sec
      Start  5: test_array
 5/11 Test  #5: test_array .......................   Passed    2.37 sec
      Start  6: test_blas
 6/11 Test  #6: test_blas ........................   Passed    4.28 sec
      Start  7: test_elemwise
 7/11 Test  #7: test_elemwise ....................***Failed    9.82 sec
      Start  8: test_error
 8/11 Test  #8: test_error .......................   Passed    0.01 sec
      Start  9: test_buffer
 9/11 Test  #9: test_buffer ......................   Passed    2.46 sec
      Start 10: test_buffer_collectives
10/11 Test #10: test_buffer_collectives ..........   Passed    0.83 sec
      Start 11: test_collectives
11/11 Test #11: test_collectives .................   Passed    0.55 sec

91% tests passed, 1 tests failed out of 11

Total Test time (real) =  26.84 sec

The following tests FAILED:
          7 - test_elemwise (Failed)
Errors while running CTest
Makefile:84: recipe for target 'test' failed
make: *** [test] Error 8
```

abergeron:
I fixed the bug that made the test fail.

nouiz:
I,ll try again tomorrow, but I didn't saw a speed up in pygpu tests with the current state of the PR with this line:

GPUARRAY_DISK_CACHE=~/.theano/gpuarray_cache DEVICE=cuda0 python -c "import pygpu;pygpu.test()"

abergeron:
Yes, because the name of the variable is GPUARRAY_CACHE_PATH, not GPUARRAY_DISK_CACHE

PR #340, opened 2017-02-02
fix GA_DECL_SHARE_PARAM for apple opencl

ar90n:
I found that GA_DECL_SHARED_PARAM makes an error in Apple OpenCL environment.
I think it seems that this issue is caused by the array declaration not decaying into pointer.
So I modified the array declaration into pointer declaration.

### DIFF
--- a/src/gpuarray_buffer_opencl.c
+++ b/src/gpuarray_buffer_opencl.c
@@ -316,7 +316,7 @@ static const char CL_PREAMBLE[] =
   "#define ga_ssize long\n"
   "#define load_half(p) vload_half(0, p)\n"
   "#define store_half(p, v) vstore_half_rtn(v, 0, p)\n"
-  "#define GA_DECL_SHARED_PARAM(type, name) , __local type name[]\n"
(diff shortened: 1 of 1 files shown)


ar90n:
Sorry, I forgot to write the error log.

```bash
[CL_DEVICE_NOT_AVAILABLE] : OpenCL Error : Error: Build Program driver returned (-2)
OpenCL Warning : clBuildProgram failed: could not build program for 0x1021c00 (AMD Radeon Pro 450 Compute Engine) (err:-2)
[CL_BUILD_ERROR] : OpenCL Build Error : Compiler build log:
<program source>:44:276: error: parameter may not be qualified with an address space
KERNEL void kSoftmax (const ga_size M, const ga_size N, const GLOBAL_MEM ga_float * x, const ga_size offset_x, const ga_ssize sx0, const ga_ssize sx1, GLOBAL_MEM ga_float * sm, const ga_size
 offset_sm, const ga_ssize sm_s0, const ga_ssize sm_s1 GA_DECL_SHARED_PARAM(ga_float,buf))
```

Issue #336, opened 2017-01-30
SIP breaks library loading (with the system python) on macOS


2shou:
I have a new macbook pro(2016) and tried to install theano with gpu support on it.
but, after I finished installing libgpuarray and theano(0.9.0beta1.dev), and test by [check1.py](http://deeplearning.net/software/theano/tutorial/using_gpu.html#testing-theano-with-gpu), it shows like this:
```
bogon:Downloads 2shou$ THEANO_FLAGS=mode=FAST_RUN,device=opencl0:0,floatX=float32 python check1.py
python: dlopen: OpenCL.framework/OpenCL: No such file or directory
Segmentation fault: 11
```
is there anything wrong?

abergeron:
This should fix the segfault: https://github.com/Theano/libgpuarray/pull/338

However it will do nothing for the root of the problem which is that the OpenCL library can't be loaded for some reason.

 * Which version of macOS that you have?
 * What is the output of `mdfind -name OpenCL.framework`?

2shou:
@abergeron 
my macos version is 10.12.3.
```
bogon:~ 2shou$ mdfind -name OpenCL.framework
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/OpenCL.framework
/System/Library/Frameworks/OpenCL.framework
```

abergeron:
I have the same version and it loads and runs fine.

Can you try to run this: `DYLD_PRINT_LIBRARIES_POST_LAUNCH=1 python -c 'import pygpu; pygpu.init("opencl0:0")'` and print the output

2shou:
it shows the same:
```
2shou$ DYLD_PRINT_LIBRARIES_POST_LAUNCH=1 python -c 'import pygpu; pygpu.init("opencl0:0")'
python: dlopen: OpenCL.framework/OpenCL: No such file or directory
```

abergeron:
This should print a large list of libraries that are loaded by python (even before it gets to trying and failing on OpenCL).  If it really prints nothing, something is wrong with your version of python.  At this point, though I don't know what.

2shou:
@abergeron 
I have tried to run this and it printed a large list of libraries but crashed finally:
```
2shou$ DEVICE=opencl0:0 python -c "import pygpu;pygpu.test()"
pygpu is installed in /Library/Python/2.7/site-packages/pygpu-0.6.0-py2.7-macosx-10.12-intel.egg/pygpu
[...]

PR #342, opened 2017-02-03
SIP breaks library loading with system python.

abergeron:
This was reported in #336.  It's not really possible to fix without a rather large effort so this just adds a warning about it in the docs.


nouiz:
Is there a way do detect this or extend an exception to also tell this? Merging how to help people find the informatin.

abergeron:
The only way to detect this is to realize that you can't load dynamic libraries (without the full path) and you are on macOS.

There isn't a convenient place where we have all the information to do that.

PR #348, opened 2017-02-07
GpuArray are now castable into int

MarcCote:
This PR allows `GpuArray` objects to be cast into `int` where appropriate.

~~This PR fixes some issues Theano scan had with the new GPU backend (see https://github.com/Theano/Theano/issues/5503).~~

### DIFF
--- a/pygpu/gpuarray.pyx
+++ b/pygpu/gpuarray.pyx
@@ -1668,6 +1668,12 @@ cdef class GpuArray:
         else:
             raise ValueError('The truth value of a multi-element array is ambiguous')
 
+    def __int__(self):
+        if self.size == 1:
+            return int(numpy.asarray(self))
+        else:
+            raise ValueError('only length-1 arrays can be converted to Python scalars')
+
     def _empty_like_me(self, dtype=None, order='C'):
         """
         _empty_like_me(dtype=None, order='C')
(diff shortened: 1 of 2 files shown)


abergeron:
This does fix some problems, but I don't think is it the proper fix for the issue you raised in Theano.  The values are of the wrong type and they shouldn't be and I don't want to just paint over cracks.

MarcCote:
I totally agree with you. Should those values be Python int instead of gpuarry (or ndarray on CPU)? https://github.com/Theano/Theano/pull/5508 forces `n_steps` to be on the CPU but it is still a ndarray.

nouiz:
I agree with @abergeron the error you have show other problems. We should fix them before merging your PR. I think your PR is good to have, I just don't want to hide problems.

Are you able to share code that reproduce the problem you have? It will speed up the fixing of them.

thanks.

MarcCote:
I'm trying to come up with a small example like the other one.

abergeron:
The problem this alleviates is fixed in a Theano PR, so I'll close this.

MarcCote:
Is the gpuarray suppose to behave like a ndarray? If so, wouldn't it need to be castable to int?

abergeron:
Yes and no.  We aim for a numpy-like API, but not for 100% drop-in support.  In this case (and for float, hex, index, ...) I don't feel that hiding the costs of transfer behind magic python features is a good idea.

PR #347, opened 2017-02-07
Add the extra ldtype argument to __array__.

abergeron:

PR #350, opened 2017-02-07
Make sure not to change the public cython C api.

abergeron:
Follow up to #347 to restore the old API.


abergeron:
Since this breaks current Theano and the change was unintentional, I'll merge.

nouiz:
seem good.

PR #355, opened 2017-02-13
Fix the blas loading in OpenCL.

abergeron:


src/gpuarray_buffer_opencl.c
```diff
@@ -1245,10 +1245,14 @@ static int cl_property(gpucontext *c, gpudata *buf, gpukernel *k, int prop_id,
...
-    if ((e = load_libclblas()) == GA_NO_ERROR)
+    if ((e = load_libclblas()) == GA_NO_ERROR) {
       *((gpuarray_blas_ops **)res) = &clblas_ops;
-    if ((e = load_libclblast()) == GA_NO_ERROR)
+      return e;
+    }
+    if ((e = load_libclblast()) == GA_NO_ERROR) {
       *((gpuarray_blas_ops **)res) = &clblast_ops;
+      return e;
+    }
```
nouiz: I'm good with that, I'll merge. But why you pick clblas before clblast? Just because we wrapped clblas before? Is one of them better then the other? Mostly out of curiosity.

abergeron: Mostly because clBLAS was wrapped before.  We should make a mechanism to select the preferred one eventually, but I don't have the time right now.

PR #359, opened 2017-02-20
Infinity in CUDA kernels

kohr-h:
The `INFINITY` C99 macro seems to be unavailable in CUDA kernels, while OpenCL has it. I added it to the default preamble in the same manner as the `NAN` macro.


kohr-h:
You may want to run the tests on CUDA and OpenCL. I ran them only on CUDA, but had some other kernel code using `INFINITY` which ran fine on OpenCL, so it's only an educated guess that the test will succeed there.

PR #360, opened 2017-02-22
Overhaul the documentation to make it easier to navigate and get rid of all the warnings

abergeron:
This also makes travis build the doc and (hopefully) error out if any new warnings sneak in there.


doc/Doxyfile
```diff
@@ -879,7 +865,7 @@ IGNORE_PREFIX          =
 # If the GENERATE_HTML tag is set to YES (the default) Doxygen will
 # generate HTML output.
 
-GENERATE_HTML          = YES
```
nouiz: why change this? you call make html that will trigger html, so this is a strange change.

nouiz: @abergeron told me IRL that this is a 2 step process, another step after will generate the html.

doc/_static/version_switch.js
```diff
@@ -0,0 +1,104 @@
+// Create version selector for documentation top bar.
```
nouiz: @slefrancois can you review this part?

src/gen_types.py
```diff
@@ -165,7 +165,6 @@ def add_type(name, sz):
  * List of all built-in types.
  */
 enum GPUARRAY_TYPES {
-  GA_POINTER = -2,
```
nouiz: should we bump a version number?

src/gpuarray/kernel.h
```diff
@@ -100,7 +102,7 @@ GPUARRAY_PUBLIC int GpuKernel_sched(GpuKernel *k, size_t n,
  * \param n dimensionality of the grid/blocks
  * \param gs sizes of launch grid
  * \param ls sizes of launch blocks
- * \param amount of dynamic shared memory to allocate
+ * \param shared of dynamic shared memory to allocate
```
nouiz: shared of -> shared amount of

src/gen_types.py
abergeron: We can bump the API version but that would entail also bumping Theano since it checks for version 1.

src/gpuarray/kernel.h
abergeron: ok

src/gen_types.py
nouiz: I would prefer that for safety.

nouiz:
travis failed.

doc/_static/version_switch.js
```diff
@@ -0,0 +1,104 @@
...
+  if (current_version != "libgpuarray" && current_version != "html") {
+    ver = current_version.replace("libgpuarray_versions/", "")
+    versions_dir[ver] = current_version
+  }
+
+  function build_vswitch() {
+  // Build HTML string for version selector, based on ReadTheDocs theme's versions.html
+
+    var vlabel = current_version.replace("libgpuarray_versions/", "");
+    if (vlabel == 'theano') {
```
slefrancois: `vlabel == 'libgpuarray'`

slefrancois:
For the version switcher, we will need to update the server-side scripts to generate the version folders and the version list for libgpuarray.

lamblin:
Rebased after PR on error messages introduced conflicts. If everything works, we can hopefully merge tomorrow, in synchronization with @slefrancois.

lamblin:
Actually, the build scripts on travis need some fixing as well.

PR #366, opened 2017-03-02
Issue 365  getitem none

### LINKED ISSUE #365: Support indexing with None for new axes
It would be nice to be able to index arrays like `x[:, None]` to create new artificial axes, since broadcasting is already supported:
```python
>>> x = pygpu.gpuarray.array([1, 2, 3], cls=pygpu.ndgpuarray)
>>> x.reshape((3, 1)) + x.reshape((1, 3))
gpuarray.array([[2, 3, 4],
       [3, 4, 5],
       [4, 5, 6]])
```
But the indexing gives an error:
```python
>>> x[:, None]

  File "<ipython-input-14-9b3b3d2c380e>", line 1, in <module>
    x[:, None]

  File "pygpu/gpuarray.pyx", line 1839, in pygpu.gpuarray.GpuArray.__getitem__ (pygpu/gpuarray.c:23530)

  File "pygpu/gpuarray.pyx", line 1884, in pygpu.gpuarray.GpuArray.__cgetitem__ (pygpu/gpuarray.c:24063)

IndexError: too many indices
```
Are there any big headaches with this type of addition to `__getitem__`?

kohr-h:
I ran this manually plus the old getitem tests, works for me. You'll notice that I moved quite a bit of logic from `__cgetitem__` to `__getitem__` simply because I am faster when I can do this kind of stuff in Python. If you feel this is gonna cost too much efficiency (which I cannot say for sure, but probably it's no issue), you may want to push it back to C. Although that would likely mean all of it.

Closes #365

### DIFF
--- a/pygpu/gpuarray.pyx
+++ b/pygpu/gpuarray.pyx
@@ -1836,7 +1836,83 @@ cdef class GpuArray:
             raise TypeError, "len() of unsized object"
 
     def __getitem__(self, key):
-        return self.__cgetitem__(key)
+        cdef unsigned int i
+
+        if key is Ellipsis:
+            return self.__cgetitem__(key)
+
+        # A list or a sequence of list should trigger "fancy" indexing.
+        # This is not implemented yet.
+        # Conversely, if a list contains slice or Ellipsis objects, it behaves
+        # the same as a tuple.
+        if isinstance(key, list):
+            if any(isinstance(k, slice) or k is Ellipsis for k in key):
+                return self.__getitem__(tuple(key))
+            else:
+                raise NotImplementedError, "fancy indexing not supported"
+
+        try:
+            iter(key)
+        except TypeError:
+            key = (key,)
+        else:
+            if all(isinstance(k, list) for k in key):
+                raise NotImplementedError, "fancy indexing not supported"
+
+            key = tuple(key)
+
+        # Need to massage Ellipsis here, to avoid packing it into a tuple.
+        if key.count(Ellipsis) > 1:
+            raise IndexError, "cannot use more than one Ellipsis"
+
+        # The following code replaces an Ellipsis found in the key by
+        # the corresponding number of slice(None) objects, depending on the
+        # number of dimensions.  As example, this allows indexing on the last
+        # dimension with a[..., 1:] on any array (including 1-dim).  This
+        # is also required for numpy compat.
+        try:
+            ell_idx = key.index(Ellipsis)
+        except ValueError:
+            pass
+        else:
+            # Need number of axes minus missing dimensions extra slice(None)
+            # objects, not counting None entries and the Ellipsis itself
+            num_slcs = self.ga.nd - (len(key) - key.count(None) - 1)
+            fill_slices = (slice(None),) * num_slcs
+            key = key[:ell_idx] + fill_slices + key[ell_idx + 1:]
+
+        # Remove the None entries for indexing
+        getitem_idcs = tuple(k for k in key if k is not None)
+
+        # For less than 1 index, fill up with slice(None) to the right.
+        # This allows indexing a[1:] in multi-dimensional arrays, where the
+        # slice is applied along the first axis only. It also allows
+        # a[()], which simply is a view in Numpy.
+        if len(getitem_idcs) <= 1:
+            getitem_idcs = (getitem_idcs +
+                            (slice(None),) * (self.ga.nd - len(getitem_idcs)))
+
+        # Slice into array, then reshape, accommodating for None entries in key
+        sliced = self.__cgetitem__(getitem_idcs)
+        if key.count(None) == 0:
+            # Avoid unnecessary reshaping if there was no None
+            return sliced
+        else:
+            new_shape = []
+            i = 0
+            if sliced.shape:
+                for k in key:
+                    if isinstance(k, int):
+                        continue
+                    elif k is None:
+                        new_shape.append(1)
+                    else:
+                        new_shape.append(sliced.shape[i])
+                        i += 1
+            # Add remaining entries from sliced.shape if existing (happens
+            # for 1 index or less if ndim >= 2).
+            new_shape.extend(sliced.shape[i:])
+            return sliced.reshape(new_shape)
 
     cdef __cgetitem__(self, key):
         cdef ssize_t *starts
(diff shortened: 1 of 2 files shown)


kohr-h:
Another remark: I changed the behavior of indexing with lists and sequences of lists that contain only integers from "same as tuple" to "raise exception". The reason is that in Numpy, this type of indexing is fundamentally different from indexing with tuples (or sequences containing slices or Ellipsis):
```python
>>> x = np.array([1, 2])
>>> x[[0, 1, 1, 0, 1]]
array([1, 2, 2, 1, 2])
x[[slice(None)]]  # same as x[:]
array([1, 2])
```

pygpu/gpuarray.pyx
```diff
@@ -1836,19 +1836,82 @@ cdef class GpuArray:
...
+                    continue
+                elif k is None:
+                    new_shape.append(1)
+                else:
+                    new_shape.append(sliced.shape[i])
+                    i += 1
+        # Add remaining entries from sliced.shape if existing (happens
+        # for 1 index or less if ndim >= 2).
+        new_shape.extend(sliced.shape[i:])
+        return sliced.reshape(new_shape)
```
abergeron: Please avoid the reshape if there where no None.  This is a performance issue.

pygpu/gpuarray.pyx
```diff
@@ -1836,19 +1836,82 @@ cdef class GpuArray:
...
+        try:
+            ell_idx = key.index(Ellipsis)
+        except ValueError:
+            pass
+        else:
+            # Need number of axes minus missing dimensions extra slice(None)
+            # objects, not counting None entries and the Ellipsis itself
+            num_slcs = self.ga.nd - (len(key) - key.count(None) - 1)
+            fill_slices = (slice(None),) * num_slcs
+            key = key[:ell_idx] + fill_slices + key[ell_idx + 1:]
```
abergeron: Moving the Ellipsis processing outside of \_\_cgetitem__ breaks `a[...] = 1` or similar things.

pygpu/gpuarray.pyx
```diff
@@ -1836,19 +1836,82 @@ cdef class GpuArray:
...
+
+        # Remove the None entries for indexing
+        getitem_idcs = tuple(k for k in key if k is not None)
```
abergeron: We might need a similar thing in the \_\_setitem__ code (yes, numpy apparently just ignores the None for that case).

pygpu/gpuarray.pyx
```diff
@@ -1836,19 +1836,82 @@ cdef class GpuArray:
...
+
+        # For less than 1 index, fill up with slice(None) to the right.
+        # This allows indexing a[1:] in multi-dimensional arrays, where the
+        # slice is applied along the first axis only. It also allows
+        # a[()], which simply is a view in Numpy.
+        if len(getitem_idcs) <= 1:
+            getitem_idcs = (getitem_idcs +
+                            (slice(None),) * (self.ga.nd - len(getitem_idcs)))
```
abergeron: Does `a[0] = [1, 2, 3, 4]` (where `a` is shape (3, 4)) still works after moving this here?

pygpu/gpuarray.pyx
```diff
@@ -1836,19 +1836,82 @@ cdef class GpuArray:
...
+
+        # Slice into array, then reshape, accommodating for None entries in key
+        sliced = self.__cgetitem__(getitem_idcs)
+        new_shape = []
+        i = 0
```
abergeron: Don't forget to add `cdef unsigned int i`at the top.

abergeron:
I'm ok with the behaviour change wrt indexing with lists.  It's better to not behave slightly differently.

pygpu/gpuarray.pyx
kohr-h: Right, I wasn't expecting that it uses `__cgetitem__`. That needs to change then, I'll check.

pygpu/gpuarray.pyx
kohr-h: I now use the view returned in `__getitem__` to do the same thing as before, so no need for new code.

pygpu/gpuarray.pyx
kohr-h: Works

pygpu/gpuarray.pyx
kohr-h: Done

pygpu/gpuarray.pyx
kohr-h: Ok, avoiding unnecessary reshaping now.

pygpu/gpuarray.pyx
```diff
@@ -1936,7 +1938,7 @@ cdef class GpuArray:
             free(steps)
 
     def __setitem__(self, idx, v):
-        cdef GpuArray tmp = self.__cgetitem__(idx)
+        cdef GpuArray tmp = self.__getitem__(idx)
```
abergeron: Don't do that.  The point of \_\_cgetitem__ was to avoid python overhead here (which is significant).

kohr-h:
> I'm ok with the behaviour change wrt indexing with lists.

Actually, I saw that there's the `take1` function which implements a part of that behavior. I think you can use this function to implement indexing with lists or sequences of lists, by converting the input to an array, then flatten it using index strides and feed the flat array into `take1`.
In Numpy I would do
```python
def __getitem__(self, idx):
    # Assuming idx is a list or sequence of lists
    idx_arr = np.asarray(idx, dtype=int)
    if self.ndim == 1:
        if idx_arr.ndim == 2:
            # Given as single-element list of index list
            flat_idx_arr = idx_arr[0]
        else:
            flat_idx_arr = idx_arr
    else:
        strides = np.concatenate(np.cumprod(self.shape[1:]), [1])
        flat_idx_arr = np.sum(idx_arr * strides[:, None], axis=0)
    return self.take1(flat_idx_arr)
```
Is that easily doable?

abergeron:
That would be doable, yes.  But let's keep this for another PR.

It might be better to write different kernels to handle the multi dimension advanced indexing though.  That is somewhere on my todo list.

Also, I would like to add support for setting elements with a list or multiple lists, which cannot use the current code.

kohr-h:
> That would be doable, yes. But let's keep this for another PR.
> 
> It might be better to write different kernels to handle the multi dimension advanced indexing though. That is somewhere on my todo list.
> 
> Also, I would like to add support for setting elements with a list or multiple lists, which cannot use the current code.

Sure, I was just wondering since I stumbled upon the function. A poor man's version with index flattening would be quite nice, better than nothing. On the other hand, as you say, `__setitem__` cannot use the current code since the `__getitem__` does not return a view in that case, so users may be surprised if setting is not supported.

kohr-h:
So, old `__cgetitem__` code back in place, and `__setitem__` uses it again. I added the same checks for fancy indexing attempts to make sure that both getters and setters behave the same way. Are there further bottlenecks to worry about?

nouiz:
Before having this PR merged, I would have liked to get some profiling. @abergeron you did speed up with cgetitem of code that was slow down otherwise in the new back-end. Do you remember that use case? We could profile it again before and after this PR. There is much python code in getitem before cgetitem is called. This could slow down computation a lot.

abergeron:
First, it's already been merged.

Second, there is no python code.  All the code is Cython and will get converted to C.  The only issue is that if you call python methods that can't be converted you pay a performance penalty.

I'm reasonably confident that this isn't the case here.

PR #371, opened 2017-03-06
Remove the driver black list

nouiz:


nouiz:
I merge now. jenkins passed and trivis only test mac and is super slow to execute.

PR #376, opened 2017-03-13
Remove the blacklist code 

abergeron:
It's horrible and I really hope we never need to do that again.

We can always look at version history if it's really required.


nouiz:
I'm pretty sure we will need it again and I don't like the idea to look in the history. This is more complicated. I also like the idea of printing the driver version.

Why not remove the code in "#if 0", and allow to query the driver version? This way it will make it easier to get the user information while debuging?

I'm not strongly opposed to this PR, so if don't like the idea to allow printing de driver version, you can merge it.

PR #386, opened 2017-03-20
triu/tril functions

tfjgeorge:
@notoraptor 

I added 2 functions in `pygpu/array_tools.py` for triu/tril and tests.


pygpu/array_tools.py
```diff
@@ -0,0 +1,47 @@
...
+        le = '>'
+    else:
+        le = '<'
+    src = tmpl.substitute(cols=cols, le=le)
+    spec = [GpuArray, 'uint32']
+    k = GpuKernel(src, "extract_tri", spec, context=ctx)
+    return k
+
+
+def triu(A, ctx, inplace=True):
```
notoraptor: As `A` is a GPU array, `triu` and `tril` can deduce `ctx` arg from `A`, that is `A.context`. So, rewrite both functions so that they take only 2 args (A and inplace).

pygpu/array_tools.py
```diff
@@ -0,0 +1,47 @@
...
+    if not inplace:
+        A = A.copy()
+    upper = True
+    if A.flags['F_CONTIGUOUS']:
+        upper = False
+    k = _generate_kernel(ctx, A.shape[0], upper)
```
notoraptor: Based on your kernel code, the 2nd parameter of `_generate_kernel` should be the number of elements on the first dimension of the array (generally called the "Leading Dimension'), i.e number of rows for a C-array, and number of columns for a F-array. So, it is not always `A.shape[0]`. It currently works because you test with square matrices, but if you test on non-square matrices, some errors will occur.

Same remarks for `tril`.

pygpu/tests/tests_array_tools.py
```diff
@@ -0,0 +1,45 @@
+from pygpu.array_tools import (tril, triu)
+from .support import (gen_gpuarray, context)
+import numpy
+
+
+def test_triu_inplace():
+    ac, ag = gen_gpuarray((10, 10), 'float32', ctx=context)
```
notoraptor: As all your tests run with square matrices, there are some cases not tested which will currently raise errors. I suggest you test non-square matrices instead in all your 6 tests. Or a square-matrix AND a non-square matrix in each test.

tfjgeorge:
@notoraptor I updated the commit with your remarks

pygpu/basic.py
```diff
@@ -0,0 +1,53 @@
+from string import Template
+from .gpuarray import GpuArray, GpuKernel
+
+
+def _generate_kernel(ctx, cols, upper=True):
+    tmpl = Template("""
+    KERNEL void extract_tri(ga_float *a, ga_uint N) {
+        unsigned int idx = blockIdx.y*blockDim.x*gridDim.x+
+                           blockIdx.x*blockDim.x+threadIdx.x;
```
abergeron: Use the cluda defines here:
threadIdx -> LID
blockDim -> LDIM
blockIdx -> GID
gridDim -> GDIM
.x -> _0
.y -> _1

pygpu/basic.py
```diff
@@ -0,0 +1,53 @@
+from string import Template
+from .gpuarray import GpuArray, GpuKernel
+
+
+def _generate_kernel(ctx, cols, upper=True):
+    tmpl = Template("""
+    KERNEL void extract_tri(ga_float *a, ga_uint N) {
```
abergeron: Add a GLOBAL_MEM before the ga_float.

notoraptor:
I suggest to raise an error if input ndim > 2, as tri/triu/tril seem to expect 2D tensors.

For 1D tensor, NumPy creates a square matrix by broadcasting the vector, and then applies triu/tril on the matrix. Is it what we want ? In such cases, `inplace=True` could not work.

@abergeron

abergeron:
Numpy does support triu/tril for 3d tensors.  It applies the operation to each 2d sub-tensor.  But I'm ok with an exception in that case.

I would also be ok with an exception in the 1d case but we have to do these checks.

I am also worried about what happens in the non-contiguous case.

notoraptor:
Ok. @tfjgeorge add also a checking that A is contiguous.

For shape checking, use `A.ndim` instead of `len(A.shape)`.

tfjgeorge:
I dont know how to check if A is contiguous ?

notoraptor:
@tfjgeorge `if (A.flags['F_CONTIGUOUS'] or A.flags['C_CONTIGUOUS'])` should work.

pygpu/basic.py
```diff
@@ -44,6 +47,9 @@ def triu(A, inplace=True):
 def tril(A, inplace=True):
     if len(A.shape) != 2:
```
notoraptor: Just use `A.nidm` instead of `len(A.shape)` here (same for triu), and all will be ok !

PR #388, opened 2017-03-24
Stop requiring nose to import pygpu and add the dependency on six.

abergeron:


pygpu/__init__.py
```diff
@@ -14,10 +14,8 @@ def get_include():
...
-    test = main.NoseTester().test
-else:
-    def test():
-        raise ImportError("The nose module is not installed."
-                          " It is needed for pygpu tests.")
+
+def test():
+    from .tests import main
+    if hasattr(main, "NoseTester"):
+        main.NoseTester().test()
```
nouiz: Should raise an error as before.

abergeron: It will raise the same error as before.  The previous error message was never encountered since it would just fail to import nose all the time.

nouiz: This make the test command fail:
```
$ DEVICE=cuda0 python -c "import pygpu;pygpu.test()"
  File "<string>", line 1, in <module>
  File "/u/bastienf/.local/lib64/python2.7/site-packages/pygpu-0.6.2-py2.7-linux-x86_64.egg/pygpu/__init__.py", line 21, in test
    main.NoseTester().test()
  File "/Tmp/lisa/os_v5/anaconda/lib/python2.7/site-packages/numpy/testing/nosetester.py", line 187, in __init__
    raise AssertionError
AssertionError
```

PR #391, opened 2017-03-28
Fix the NoseTester invocation so that it actually works now.

abergeron:
This should "unbreak" the commits in #388


abergeron:
This also fixes #392.

PR #395, opened 2017-03-31
Fix wrong result of GpuArray_rgemmBatch_3d when the inputs aren't in …

nouiz:
…C order.

see https://github.com/Theano/Theano/issues/5795


nouiz:
Don't merge. This don't always fix the problem.

abergeron:
Better fix in #398

PR #398, opened 2017-04-03
Fix contiguous detection to properly handle F-contiguous input.

abergeron:
Fixes https://github.com/Theano/Theano/issues/5795

### DIFF
--- a/src/gpuarray_array_blas.c
+++ b/src/gpuarray_array_blas.c
@@ -439,18 +439,18 @@ int GpuArray_rger(double alpha, GpuArray *X, GpuArray *Y, GpuArray *A,
   return err;
 }
 
-static inline int is_2d_contiguous(const GpuArray *a) {
+static inline int is_last_2d_contiguous(const GpuArray *a) {
   size_t size = GpuArray_ITEMSIZE(a);
 
   if (GpuArray_IS_C_CONTIGUOUS(a))
     return 1; // C contiguous
 
-  if (a->strides[1] <= 0 || a->strides[2] <= 0)
+  if (a->strides[a->nd - 2] <= 0 || a->strides[a->nd - 1] <= 0)
     return 0;
 
-  if (a->strides[1] == size)
+  if (a->strides[a->nd - 2] == size)
     return 2; // F contiguous
-  if (a->strides[2] == size)
+  if (a->strides[a->nd - 1] == size)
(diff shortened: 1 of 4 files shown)


src/gpuarray_array_blas.c
```diff
@@ -439,6 +439,23 @@ int GpuArray_rger(double alpha, GpuArray *X, GpuArray *Y, GpuArray *A,
   return err;
 }
 
+static inline int is_2d_contiguous(const GpuArray *a) {
```
nouiz: The name isn't clear to what it does. Currently, it always expect a 2d inputs. If you want to keep that name,use ndim-1 and ndim-2 for which dimensions you checks. Otherwise, rename.

abergeron: It expects a 3d input and makes sure that the last two are contiguous.

abergeron: I'm ok with a rename, but to what?

nouiz: is_last_2d_contiguous and change the to use ndim -1 and ndim-2 in the code.

tests/check_blas.c
```diff
@@ -14,18 +14,99 @@ void teardown(void);
...
+  ck_assert_fbuf_eq(data, res, sizeof(res));
+}
+END_TEST
+
+START_TEST(test_gemmBatch_3d_F) {
+  GpuArray A;
+  GpuArray B;
+  GpuArray C;
+
+  size_t dims[3] = {2, 32, 32};
```
nouiz: typo, should be 2,3,3 how it can have passed ?

tests/check_blas.c
```diff
@@ -14,18 +14,99 @@ void teardown(void);
...
+}
+END_TEST
+
+START_TEST(test_gemmBatch_3d_S) {
+  GpuArray A;
+  GpuArray B;
+  GpuArray C;
+  ssize_t t;
+
+  size_t dims[3] = {2, 32, 32};
```
nouiz: should be 2,3,3, why the test passed?

abergeron:
I'm working on this, it's WIP.

abergeron:
Now it's final.

nouiz:
Don't forget to raise the version number

abergeron:
We should bundle this and the disk cache in 0.6.4

PR #401, opened 2017-04-11
Force the use of managed memory.

nouiz:
TODO: Make a user option of this.

### DIFF
--- a/src/loaders/libcuda.fn
+++ b/src/loaders/libcuda.fn
@@ -30,6 +30,7 @@ DEF_PROC_V2(cuMemGetInfo, (size_t *free, size_t *total));
 DEF_PROC_V2(cuMemAlloc, (CUdeviceptr *dptr, size_t bytesize));
 DEF_PROC_V2(cuMemFree, (CUdeviceptr dptr));
 DEF_PROC_V2(cuMemAllocHost, (void **pp, size_t bytesize));
+DEF_PROC(cuMemAllocManaged, (CUdeviceptr* dptr, size_t bytesize, unsigned int flags));
 DEF_PROC(cuMemFreeHost, (void *p));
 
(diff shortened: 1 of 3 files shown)


nouiz:
@abergeron how do you suggest to give the run time option to the user? Could this information be in the context and passed at context init?

src/loaders/libcuda.fn
```diff
@@ -30,6 +30,7 @@ DEF_PROC_V2(cuMemGetInfo, (size_t *free, size_t *total));
 DEF_PROC_V2(cuMemAlloc, (CUdeviceptr *dptr, size_t bytesize));
 DEF_PROC_V2(cuMemFree, (CUdeviceptr dptr));
 DEF_PROC_V2(cuMemAllocHost, (void **pp, size_t bytesize));
+DEF_PROC(cuMemAllocManaged, (CUdeviceptr* dptr, size_t bytesize, unsigned int flags));
```
nouiz: I tried to use DEF_PROC_V2 here and it failed. Why? How this work? I followed the code but didn't understood.

nouiz: I found why.

abergeron:
Since this doesn't seem to make a difference in speed (if you don't oversubscribe) I'm inclined to enable this by default.

However @nouiz wants to have some sort of user warning if you actually exceed the memory available on the gpu to warn about potential slowdown.  This could be a simple fprintf(), but I dislike chatty libraries for things that are rather common.

nouiz:
I did only the timing for big size matrices on Pascal. I need to time for
small matrices in case it add overhead and test it on pre Pascal GPU.

nouiz:
I tested on an 750. The speed was about the same, but I wasn't able to allocate as much memory! So we can't activate this by default.

nouiz:
@abergeron told that maybe we could use it automatically/always only for Pascal GPU.

slefrancois:
jenkins test this (https)

slefrancois:
jenkins test this (https)

PR #406, opened 2017-04-13
Require only CMake 2.8 (which appears to work just fine)

jeremybarnes:
Less recent versions of Ubuntu (we have 14.04.3) come with CMake 2.8.  Modifying the version, everything appears to build just fine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 2.8)
 
 PROJECT(libgpuarray C)
 


abergeron:
It works ok on linux, but breaks on mac so I want to keep the current minimum (and maybe windows).

One way to get cmake 3.0 for ubuntu 14.04 is through conda.

Otherwise, you can keep using a local modification like this.

nouiz:
Can we make the minimun OS dependent?

jeremybarnes:
Is there any more info on how it breaks on Mac?  I would be willing to look for a workaround if there was interest.  It is certainly possible to get later versions, but it causes friction in corporate environments, hence the interest in working out of the box.

abergeron:
2.8 fails to find the compiler on newer mac OS because they moved the default location.  There may be other problems but once I read that it was fixed in 3.0 I didn't spend much time investigating.

I just tried and apparently we can put cmake_minimum_required() in a conditional so that could be a option.

Issue #411, opened 2017-04-19
fp16 appears to break gpu collectives


astooke:
Hi,
  
  Is this known / being worked on?  When I try to call collectives under theano.config.floatX="float16", I get the error:

```
File "pygpu/collectives.pyx", line 257, in pygpu.collectives.GpuComm.broadcast (pygpu/collectives.c:5022)
  File "pygpu/collectives.pyx", line 362, in pygpu.collectives.comm_broadcast (pygpu/collectives.c:6060)
pygpu.gpuarray.GpuArrayException: b'Invalid value or operation'
```

NCCL can definitely do half-precision.

Really hoping for that free 2x speedup ;)

nouiz:
@tsirif any idea? A quick look in the code see to indicate it is supported.

Note, you won't get a 2x speed up from the Theano side. Theano only support float16 storage, not float16 computation. But it will speed up the memory transfer and memory access.

tsirif:
It is trivial to include the correspondence between the types in `gpuarray_collectives_cuda_nccl.c` in [this](https://github.com/Theano/libgpuarray/blob/master/src/gpuarray_collectives_cuda_nccl.c#L168) and I believe it would not need any other modifications from C's side. In pygpu however I am not sure how can I establish such a correspondence between a Numpy's type and float16 (maybe we should handle a custom type for float16 all the way...). Also, among the types defined in `gpuarray/types.h`, which one corresponds float16? Is it [GA_FLOAT16](https://github.com/Theano/libgpuarray/blob/master/src/gpuarray/types.h#L117)? Is it in use/could it be used?

nouiz:
It is the GA_HALF and ga_float16 c type that correspond to float16.

nouiz:
The other PR was merged, so to my knowledge all should work. So closing. If you find other problem, tell us.

PR #414, opened 2017-04-20
Fix float 16 for cuda collectives

tsirif:
#411

### DIFF
--- a/src/gpuarray_collectives_cuda_nccl.c
+++ b/src/gpuarray_collectives_cuda_nccl.c
@@ -179,10 +179,8 @@ static inline ncclDataType_t convert_data_type(int typecode) {
   case GA_DOUBLE: return ncclDouble;
   case GA_LONG: return ncclInt64;
   case GA_ULONG: return ncclUint64;
-  #ifdef CUDA_HAS_HALF
   case GA_HALF: return ncclHalf;
   case GA_FLOAT16: return ncclHalf;
-  #endif
   }
   return nccl_NUM_TYPES;
 }
--- a/src/gpuarray_collectives_cuda_nccl.c
+++ b/src/gpuarray_collectives_cuda_nccl.c
@@ -175,11 +175,14 @@ static inline ncclDataType_t convert_data_type(int typecode) {
   switch (typecode) {
   case GA_BYTE: return ncclChar;
   case GA_INT: return ncclInt;
-  case GA_HALF: return ncclHalf;
   case GA_FLOAT: return ncclFloat;
   case GA_DOUBLE: return ncclDouble;
   case GA_LONG: return ncclInt64;
   case GA_ULONG: return ncclUint64;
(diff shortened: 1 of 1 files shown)


src/gpuarray_collectives_cuda_nccl.c
```diff
@@ -175,11 +175,14 @@ static inline ncclDataType_t convert_data_type(int typecode) {
   switch (typecode) {
   case GA_BYTE: return ncclChar;
   case GA_INT: return ncclInt;
-  case GA_HALF: return ncclHalf;
   case GA_FLOAT: return ncclFloat;
   case GA_DOUBLE: return ncclDouble;
   case GA_LONG: return ncclInt64;
   case GA_ULONG: return ncclUint64;
+  #ifdef CUDA_HAS_HALF
```
nouiz: I'm not sure what to do if CUDA_HAS_HALF isn't defined. I think it should always be defined as we only support recent enough cuda version. So I would just remove the ifdef. Do you agree with that?

otherwise, it look good. thanks.

tsirif: This is defined in `nccl.h` and is enabled if there is compatibility. Should I remove it?

nouiz: I remember that, but I think GA_HALF is always defined. So this cause a problem if libgpuarray support it while nccl don't. Here we don't give a good error message. Here, we request cuda 7 or more recent:

http://deeplearning.net/software/libgpuarray/installation.html#run-requirements

And CUDA_HAS_HALF is always true in that case. So I would remove the ifdef.

abergeron: Yes don't use an ifdef switch since we load the libraries dynamically and we don't want to disable it for future loads.

nouiz:
I pushed the small change to this PR. If travis pass, I'll merge.

thanks

src/gpuarray_collectives_cuda_nccl.c
```diff
@@ -175,11 +175,12 @@ static inline ncclDataType_t convert_data_type(int typecode) {
   switch (typecode) {
   case GA_BYTE: return ncclChar;
   case GA_INT: return ncclInt;
-  case GA_HALF: return ncclHalf;
   case GA_FLOAT: return ncclFloat;
   case GA_DOUBLE: return ncclDouble;
   case GA_LONG: return ncclInt64;
   case GA_ULONG: return ncclUint64;
+  case GA_HALF: return ncclHalf;
+  case GA_FLOAT16: return ncclHalf;
```
abergeron: GA_FLOAT16 means a vector of 16 floats.  This is horribly inaccurate.

tsirif: I had asked on #411 about the types. Excuse me for the inconvenience.

PR #389, opened 2017-03-24
Add better error reporting from the backends

abergeron:
This allows the backend to clarify errors by adding error messages that include relevant information.

This should make it much easier to track down problems reported by users and even allows a trace of all the errors encountered in debug mode.

The code depends on the disk_cache PR since that modifies a lot of the error handling and I didn't want to do the work twice.  I will rebase once that is in.

PR #415, opened 2017-04-20
Try to fix compilation errors on Windows/Python 2.7/64 bits.

notoraptor:
The last master version of libgpuarray fails to compile on Windows 8.1 64 bits for Python 2.7 with Microsoft C++ Compiler 2008. This PR tries to fix all errors I have found. **Not yet tested on Linux**.

@abergeron @nouiz

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ lib
 .*.sw[po]
 *~
 *.pyc
+*.pyd
 *.pyo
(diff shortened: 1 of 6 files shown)


notoraptor:
PS: Seems another user has found similar problems: https://github.com/Theano/libgpuarray/pull/389#issuecomment-295781542

PR #416, opened 2017-04-20
Tell that people can use conda to install libgpuarray and pygpu.

nouiz:


doc/installation.rst
```diff
@@ -13,6 +13,17 @@ It should also work on any decently recent OS not listed here. If you
 get an error during the build on your favorite OS, please report it
 and we will attempt to fix it.
 
+Conda
+-----
+
+The easiest way to install libgpuarray is with conda::
+
+  conda install pygpu
```
abergeron: Maybe we want to mention that conda-forge uses libgpuarray as the package name.

nouiz: I checked and in both conda and conda-forge, there is two packges: libgpuarray and pygpu. The next line also talk about the libgpuarray package. So I don't understand what you want to change.

I just added a commit that tell about conda-forge.

PR #419, opened 2017-04-25
python interface for rgemmBatch_3d

khaotik:
Basically c/p from gemm code.

In tests, I removed the cases where destination array is fortran-contiguous, because it would trigger an error at [here](https://github.com/Theano/libgpuarray/blob/master/src/gpuarray_array_blas.c#L540).

```c
cC = is_last_2d_contiguous(C);
  if (!cC) {
    err = GA_VALUE_ERROR;
    goto cleanup;
  }
```

The function `is_last_2d_contiguous` can't detect a rank-3 fortran-contiguous array. Because the least strided index would be 0 while this only test for the last two indices. For example:

```python
>>> x = np.random.rand(3,4,5).astype('float32')
>>> y = np.asfortranarray(x)
>>> x.strides
(80, 20, 4)
>>> y.strides
(4, 12, 48)
```

The batched gemm needs an 3d array with the 2nd stride being 4, so it's neither c-contig or f-contig as whole. I'm not sure what's the most convenient solution here.

### DIFF
--- a/pygpu/blas.pyx
+++ b/pygpu/blas.pyx
@@ -92,14 +92,14 @@ def gemv(double alpha, GpuArray A, GpuArray X, double beta=0.0,
         transA = cb_no_trans
 
     if A.ga.nd != 2:
-        raise TypeError, "A is not a matrix"
+        raise TypeError("A is not a matrix")
     if transA == cb_no_trans:
         Yshp = A.ga.dimensions[0]
     else:
         Yshp = A.ga.dimensions[1]
     if Y is None:
         if beta != 0.0:
-            raise ValueError, "Y not provided and beta != 0"
+            raise ValueError("Y not provided and beta != 0")
         Y = pygpu_empty(1, &Yshp, A.ga.typecode, GA_ANY_ORDER, A.context, None)
         overwrite_y = True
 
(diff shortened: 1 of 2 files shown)


pygpu/blas.pyx
```diff
@@ -153,3 +167,47 @@ def ger(double alpha, GpuArray X, GpuArray Y, GpuArray A=None,
...
+        Cshp[1] = A.ga.dimensions[1]
+    else:
+        Cshp[1] = A.ga.dimensions[2]
+    if transB == cb_no_trans:
+        Cshp[2] = B.ga.dimensions[2]
+    else:
+        Cshp[2] = B.ga.dimensions[1]
+    if C is None:
+        if beta != 0.0:
+            raise ValueError, "C not provided and beta != 0"
```
nouiz: This is the old python syntax, can you use "raise ValueError(...)"

pygpu/blas.pyx
```diff
@@ -153,3 +167,47 @@ def ger(double alpha, GpuArray X, GpuArray Y, GpuArray A=None,
...
+        C = pygpu_empty(3, Cshp, A.ga.typecode, GA_ANY_ORDER, A.context, None)
+        overwrite_c = True
```
nouiz: Instead of that line, you can remove that line and make the next line "elif"

pygpu/blas.pyx
```diff
@@ -153,3 +167,45 @@ def ger(double alpha, GpuArray X, GpuArray Y, GpuArray A=None,
...
+        Cshp[1] = A.ga.dimensions[2]
+    if transB == cb_no_trans:
+        Cshp[2] = B.ga.dimensions[2]
+    else:
+        Cshp[2] = B.ga.dimensions[1]
+    if C is None:
+        if beta != 0.0:
+            raise ValueError("C not provided and beta != 0")
+        C = pygpu_empty(3, Cshp, A.ga.typecode, GA_ANY_ORDER, A.context, None)
+    else:
```
abergeron: This should check for `overwrite_c`.

abergeron:
jenkins test this please

PR #430, opened 2017-05-05
Use secure HTTPS URLs where possible

paulmenzel:
Not all URLs could be converted, as there are still Web sites out there not having HTTPS set up correctly.

### DIFF
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -9,7 +9,7 @@
 
   (function() {
     var ga = document.createElement('script'); ga.type = 'text/javascript'; ga.async = true;
-    ga.src = ('https:' == document.location.protocol ? 'https://ssl' : 'http://www') + '.google-analytics.com/ga.js';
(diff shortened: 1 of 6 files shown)


doc/_templates/layout.html
```diff
@@ -9,7 +9,7 @@
 
   (function() {
     var ga = document.createElement('script'); ga.type = 'text/javascript'; ga.async = true;
-    ga.src = ('https:' == document.location.protocol ? 'https://ssl' : 'http://www') + '.google-analytics.com/ga.js';
+    ga.src = 'https://ssl.google-analytics.com/ga.js';
```
nouiz: @slefrancois What do you think of that? I think it is good, but I'm not 100% sure.

slefrancois: I'm not a web design expert, but it seems reasonable that google sites would be available in secure connections.

nouiz:
thanks

PR #436, opened 2017-05-16
Use "right" ld* with 1D matrices

lamblin:
This is re-applying an old fix that was in Theano, see https://github.com/Theano/Theano/blob/master/theano/tensor/blas.py#L649-L663

This first commit fixes Theano/Theano#5730, but there are many other places lda/b/c would be considered "incorrect".

Another option might be to consider the other contiguity pattern here, but I don't know how to make sure we choose the right one.


nouiz:
look good, just miss test. I'm fine to have them in Theano or to merge now and wait for @abergeron to do them later if you prefer. If so, just merge this PR.

lamblin:
The test is in https://github.com/Theano/Theano/issues/5950

slefrancois:
jenkins test this for earlgrey

nouiz:
I'll merge now as travis error are due to there software stack and jenkins pass.

PR #440, opened 2017-05-18
Add error messages in gpuarray_array.c

lamblin:
Follow-up to #389.

### DIFF
--- a/src/gpuarray_array.c
+++ b/src/gpuarray_array.c
@@ -76,8 +76,7 @@ static int ga_extcopy(GpuArray *dst, const GpuArray *src) {
                                       (cache_freev_fn)GpuElemwise_free,
                                       ctx->err);
     if (ctx->extcopy_cache == NULL)
-      return error_set(ctx->err, GA_MISC_ERROR,
-                       "No context cache");
+      return ctx->err->code;
     if (cache_add(ctx->extcopy_cache, aa, k) != 0)
       return error_set(ctx->err, GA_MISC_ERROR,
                        "Could not store GpuElemwise copy kernel in context cache");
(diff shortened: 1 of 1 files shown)


src/gpuarray_array.c
```diff
@@ -62,21 +63,24 @@ static int ga_extcopy(GpuArray *dst, const GpuArray *src) {
...
     }
     if (ctx->extcopy_cache == NULL)
       ctx->extcopy_cache = cache_twoq(4, 8, 8, 2, extcopy_eq, extcopy_hash,
                                       extcopy_free,
                                       (cache_freev_fn)GpuElemwise_free,
                                       ctx->err);
     if (ctx->extcopy_cache == NULL)
-      return GA_MISC_ERROR;
+      return error_set(ctx->err, GA_MISC_ERROR,
+                       "No context cache");
```
abergeron: cache_twoq should have its own error that you should forward.

src/gpuarray_array.c
```diff
@@ -440,27 +457,35 @@ int GpuArray_take1(GpuArray *a, const GpuArray *v, const GpuArray *i,
...
+  if (!GpuArray_IS_C_CONTIGUOUS(i))
+    return error_set(ctx->err, GA_INVALID_ERROR, "Index array (i) not C-contiguous");
 
   /* Check that the dimensions match namely a[0] == i[0] and a[>0] == v[>0] */
   if (v->nd == 0 || a->nd == 0 || i->nd != 1 || a->nd != v->nd ||
       a->dimensions[0] != i->dimensions[0])
-    return GA_INVALID_ERROR;
+    return error_fmt(ctx->err, GA_INVALID_ERROR, "Dimension mismatch. "
+                     "v->nd = %llu, a->nd = %llu, i->nd = %llu, a->dimensions[0] = %llu, i->dimensions[0] = %llu",
+                     v->nd, a->nd, i->nd, a->nd > 0 ? a->dimensions[0] : 0, i->nd > 1 ? i->dimensions[0] : 1);
```
abergeron: Why does `i->nd` has to be bigger than 1?  Also why is it swapped out for 1?

abergeron:
This apparently segfaulted jenkins.

src/gpuarray_array.c
lamblin: Should have been `i->nd > 0` indeed.
But I'm not sure what to print in the case where `nd == 0`. Does `dimensions` even exist? I wanted to print 1 because there is only one element in total and it would have fitted `%llu`, but maybe we need to rework how the errors are detected in two phases.

src/gpuarray_array.c
lamblin: So something like `return ctx->err->code;`?

src/gpuarray_array.c
```diff
@@ -62,21 +63,24 @@ static int ga_extcopy(GpuArray *dst, const GpuArray *src) {
...
                                       (cache_freev_fn)GpuElemwise_free,
                                       ctx->err);
     if (ctx->extcopy_cache == NULL)
-      return GA_MISC_ERROR;
+      return error_set(ctx->err, GA_MISC_ERROR,
+                       "No context cache");
     if (cache_add(ctx->extcopy_cache, aa, k) != 0)
-      return GA_MISC_ERROR;
+      return error_set(ctx->err, GA_MISC_ERROR,
+                       "Could not store GpuElemwise copy kernel in context cache");
```
lamblin: Does `cache_add` set its own error as well?

lamblin:
There may be other places I grab the context from an array that might not be there, it could be worth double-checking that.

src/gpuarray_array.c
abergeron: Yes.

PR #442, opened 2017-05-23
Mixed fixes

abergeron:
Some accumulated small fixes.  Most notably float16 is changed to a struct to avoid bad results.

### DIFF
--- a/pygpu/_elemwise.pyx
+++ b/pygpu/_elemwise.pyx
@@ -49,6 +49,9 @@ cdef class arg:
         memset(&self.a, 0, sizeof(gpuelemwise_arg))
 
     def __init__(self, name, type, read=False, write=False, scalar=False):
+        # Make sure to clear previous storage
+        # __init__ may be called more than once
+        free(self.a.name)
         self.a.name = strdup(to_bytes(name))
         if self.a.name is NULL:
             raise MemoryError
@@ -62,6 +65,9 @@ cdef class arg:
             self.a.flags |= GE_SCALAR
         if self.a.flags == 0:
             raise ValueError('no flags specified for arg %s' % (name,))
+
+    def __dealloc__(self):
+        free(self.a.name)
 
     property name:
         def __get__(self):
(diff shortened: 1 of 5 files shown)


src/gpuarray_buffer_cuda.c
```diff
@@ -449,11 +449,11 @@ static const char CUDA_PREAMBLE[] =
     "#define ga_ulong unsigned long long\n"
     "#define ga_float float\n"
     "#define ga_double double\n"
-    "#define ga_half ga_ushort\n"
+    "typedef struct _ga_half { ga_ushort data; } ga_half;\n"
```
nouiz: version bump? Do Theano tests still pass?

abergeron: I don't know about Theano, we would have to try.

abergeron:
The latest commits will require some changes to Theano (mainly to remove the atomicAdd/Exch implementation in GpuAdvancedIncSubtensor1_dev20).  These functions cheat a bit so they are affected by the structure change.

The tests with Theano are running so we'll see what else falls out.

abergeron:
There are 25 test failures in Theano.  Two of those seem to have nothing to do with float16 but the others look like fallout from the struct change.

Given that, I'm not willing to make the change in a minor release.  I would rather push this back to 0.7.  The other fixes in the PR should be fine though.  I can isolate them.

nouiz:
Ok to move the float16 struct change to another PR and 0.7

But I think we should do that soon. Keep the failed test in Theano with
that changes. They are indication of places that could give wrong result.
Can you look at them to see if you can spot such stuff?

abergeron:
I've removed the float16 commits from this.  I'll check the failed tests in Theano for wrong results.

abergeron:
It seems that all of the failures in Theano come from something like this:

```
kernel void f(ga_half *res) {
  res[0] = __float2half_rn(value);
}
```
And the error is something like `Can't store 'unsigned short' in 'struct _ga_half'`

So there doesn't appear to be any wrong-results stuff hiding there, but we might want to flesh out the struct a bit to avoid having to do those conversions manually.  I'll leave that for later.

PR #446, opened 2017-06-05
Add offset in DEBUG

abergeron:
Adds an offset to all GpuArrays in DEBUG to help catch errors where the offset is not handled appropriately.

Also fix the meaning of .gpudata/.data/.base_data to better emulate what PyCUDA/PyOpenCL does.

### DIFF
--- a/pygpu/gpuarray.pyx
+++ b/pygpu/gpuarray.pyx
@@ -2229,13 +2229,40 @@ cdef class GpuArray:
         def __get__(self):
             return self.ga.offset
 
-    property gpudata:
-        "Return a pointer to the raw backend object."
+    property data:
+        """Return a pointer to the raw OpenCL buffer object.
+
+        This will fail for arrays that have an offset.
+        """
         def __get__(self):
+            if self.context.kind != b"opencl":
+                raise TypeError("This is for OpenCL arrays.")
+            if self.offset != 0:
+                raise ValueError("This array has an offset.")
             # This wizadry grabs the actual backend pointer since it's
             # guarenteed to be the first element of the gpudata
             # structure.
             return <size_t>((<void **>self.ga.data)[0])
+
+    property base_data:
+        "Return a pointer to the backing OpenCL object."
+        def __get__(self):
+            if self.context.kind != b"opencl":
+                raise TypeError("This is for OpenCL arrays.")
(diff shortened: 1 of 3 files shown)


nouiz:
Note, triu, tril wasn't bugged. As we where working on newly allocated memory, it was working well as it should normally don't have an offset. Having it support offset is good as it make Theano tests when we enable that debug config.

pygpu/gpuarray.pyx
```diff
@@ -2229,13 +2229,40 @@ cdef class GpuArray:
         def __get__(self):
             return self.ga.offset
 
+    property data:
+        """Return a pointer to the raw OpenCL buffer object.
+
+        This will fail for arrays that have an offset.
```
nouiz: Add a comment that we support pyopencl interface that raise an error if offset isn't 0. It took me some times to understand why you chosed that interface instead of just adding the offset as gpudata

pygpu/gpuarray.pyx
```diff
@@ -2229,13 +2229,40 @@ cdef class GpuArray:
...
     property gpudata:
         "Return a pointer to the raw backend object."
         def __get__(self):
+            if self.context.kind != b"cuda":
+                raise TypeError("This is for CUDA arrays.")
             # This wizadry grabs the actual backend pointer since it's
             # guarenteed to be the first element of the gpudata
             # structure.
-            return <size_t>((<void **>self.ga.data)[0])
+            return <size_t>((<void **>self.ga.data)[0]) + self.offset
```
nouiz: I'm fine with this, but why make gpudata different then the data attribute? What is your reasoning?

pygpu/gpuarray.pyx
abergeron: It's not possible to just add the offset in the OpenCL case.

I chose the name `data` to help compatibility with PyOpenCL, but the exception would be there regardless.

pygpu/gpuarray.pyx
abergeron: The backend pointer for cuda is actually a pointer, but for OpenCL it's an opaque struct that we can't just add the offset to.  This is why there are two different attributes.

abergeron: Also this helps compatibility with PyCUDA which also does this.

PR #447, opened 2017-06-05
Use the new GemmStridedBatched functions from cublas for the 3d gemm.

abergeron:
This may or may not be what we want since it will break the functionality for everyone that is not running cuda 8.

However the code is much cleaner this way.  It might be possible to hack in a way to fallback to the old code if desired, but that feels like a lot of work for little gain.


src/gpuarray/util.h
```diff
@@ -98,6 +98,90 @@ GPUARRAY_PUBLIC void gpuarray_elemwise_collapse(unsigned int n,
                                                 unsigned int *nd,
                                                 size_t *dim, ssize_t **strs);
 
+
+typedef uint16_t ga_half_t;
```
nouiz: Why do that here? We want to move to a struct. Can you use the struct here?

src/gpuarray/buffer_blas.h
```diff
@@ -115,15 +115,30 @@ GPUARRAY_PUBLIC int gpublas_hgemmBatch(
   float beta, gpudata **C, size_t *offC, size_t ldc,
   size_t batchCount, int flags);
 
-//TODO: float should be half
-GPUARRAY_PUBLIC int gpublas_hgemmStridedBatch(
+GPUARRAY_PUBLIC int gpublas_hgemm3D(
```
nouiz: I don't like much the new name. why that won't be a tensordot? 3d is the part I don't like too much.

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -573,11 +573,11 @@ static int hgemm3D(cb_order order, cb_transpose transA, cb_transpose transB,
   GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(C, CUDA_WAIT_ALL));
   err = cublasHgemmStridedBatched(h->h,
 				  convT(transA), convT(transB),
-				  M, N, K, &halpha,
+				  M, N, K, (__half *)&halpha,
```
nouiz: That conversion work? Why don't use you use fct to convert float to half?

nouiz:
@lamblin do you think we can drop batched gemm for cuda 7.5? I think it is fine.

@abergeron why do you tell we need to hack the old code? Can we test at the right place if we have the new cuda fct? If so, that would make detecting it easy. Otherwise, we could call the new fct and if they fail fall back to the old one...

src/gpuarray_blas_cuda_cublas.c
abergeron: There is a conversion above with a function.  This is just a cast so the compiler is happy.

abergeron:
We could do a fallback at the level of buffer *gemm3D that would make temporary pointer arrays and use the old batch functions.

The alternative I had in mind was to have GpuArray_rgemmBatch_3d realise that the new functions don't work and fallback to the old code with the pointer lists.

Either will introduce lots of fiddly code that we probably won't test much, so I would avoid it if possible.

nouiz:
Poke Pascal about that. I'm not sure he will see my questions in the
comments.

lamblin:
After discussion IRL, I think it is best not to drop support for cuda 7.5 right away.
Since the old functions are there, we can fall back (if the dtype is not float16) to the old implementation for now.
We can clean that up later, for instance when we release 0.7 and bump requirements.

abergeron:
I managed to make the fallback self-contained so it's not as ugly as I thought.

Now it should work as before on cuda < 8.

lamblin:
There seems to be compilation errors when compiling the extensions for pygpu.

abergeron:
Should be fixed now.

slefrancois:
jenkins test this

src/gpuarray_array_blas.c
```diff
@@ -625,50 +622,90 @@ int GpuArray_rgemmBatch_3d(cb_transpose transA, cb_transpose transB, double alph
...
+      B_datas[i] = Bp->data;
+      C_datas[i] = Cp->data;
+      A_offsets[i] = (Ap->offset + i * Ap->strides[0]) / elsize;
+      B_offsets[i] = (Bp->offset + i * Bp->strides[0]) / elsize;
+      C_offsets[i] = (Cp->offset + i * Cp->strides[0]) / elsize;
+    }
+
+    switch (C->typecode) {
+      case GA_HALF:
+        err = gpublas_hgemmBatch(o, transA, transB, m, n, k, (float)alpha,
```
nouiz: cuda8 don't define that method: http://docs.nvidia.com/cuda/cublas/#cublas-lt-t-gt-gemmbatched

So what should we do? Remove it?

abergeron: That's ok, it will just fail.

lamblin: If that call is expected to always fail, maybe we should remove it and fail explicitly...

abergeron: It might stop failing in the future.  Or it might work for OpenCL.

I would prefer to keep it as-is so that if someone in the future want to implement this they don't have to chase conditions up and down the stack.

nouiz: ok to keep it for now.

src/gpuarray/buffer_blas.h
```diff
@@ -115,6 +115,30 @@ GPUARRAY_PUBLIC int gpublas_hgemmBatch(
   float beta, gpudata **C, size_t *offC, size_t ldc,
   size_t batchCount, int flags);
 
+GPUARRAY_PUBLIC int gpublas_hgemm3D(
```
nouiz: There is still the question of the fct name. Why not use the cublas name? Also, I think this 3d can be confonding, as what is a 3d gemm? People could think about tensordot. So there is many interpretation, that isn't good.

lamblin: FWIW, I'm fine with the 3D name.

nouiz:
There is problems with the corresponding Theano PR. Tests fail in
debugmode. I prefer to wait for both to be ready before merging any of
them, just in case.

lamblin:
I launched the tests on Theano's master branch just to check, and I've got the following issue, at least when libgpuarray was compiled in Debug.
It does not seem to be related to this PR, and hopefully compiling in release would work.

[libgpuarray_compile_error.txt](https://github.com/Theano/libgpuarray/files/1086944/libgpuarray_compile_error.txt)

abergeron:
The problem is because you are compiling on a Pascal and the atomicAdd implementation for double should be wrapped in `#ifdef __CUDA_ARCH__ < 600`.  This was fixed in some places, but apparently not everywhere.

Alternatively we could try to push that down so that libgpuarray provides these alternate implementations correctly.

lamblin:
Oh, OK. Strange that it happens when libgpuarray is compiled in Debug, but not in Release, though.

abergeron:
It should also happen in Release.

lamblin:
I have a fix in https://github.com/Theano/Theano/pull/6057, at least temporarily.

src/gpuarray/util.h
```diff
@@ -98,6 +98,90 @@ GPUARRAY_PUBLIC void gpuarray_elemwise_collapse(unsigned int n,
                                                 unsigned int *nd,
                                                 size_t *dim, ssize_t **strs);
 
+
+typedef struct _ga_half_t { uint16_t h; } ga_half_t;
+
+/* code strongly inspired from
+   https://github.com/numpy/numpy/blob/master/numpy/core/src/npymath/halffloat.c#L246 */
+
+static inline ga_half_t ga_float2half(float f) {
```
jakirkham: This seems to cause issues for Visual Studio. Actually only the most recent version of MSVC seems to be ok. Please see this [failed build]( https://ci.appveyor.com/project/conda-forge/libgpuarray-feedstock/build/1.0.9/job/fcetfa7a6ux8bfe3#L388 ) for details. Also this [SO question]( https://stackoverflow.com/q/32851292 ) and its [answer]( https://stackoverflow.com/a/32959521 ) seem pertinent.

jakirkham: Have raised this as issue ( https://github.com/Theano/libgpuarray/issues/461 ).

PR #448, opened 2017-06-06
Use blocking_sync instead of yield when configured for multi-threading

f0k:
As discussed in https://github.com/Theano/Theano/issues/5706, this PR changes the CUDA host/device synchronization to use blocking synchronization instead of yielding when configured for multi-threading (e.g., when `gpuarray.sched=multi` in Theano). This lowers CPU utilization and power consumption by a large margin while the CPU is waiting for the GPU.

The alternative would be to introduce a third option so one could select between spinning (the current default), yielding and blocking, but in https://github.com/Theano/Theano/issues/5706 we figured that yielding may not be that useful anyway.

Closes https://github.com/Theano/Theano/issues/5706. (Not sure if that works across project boundaries? Probably not.)


f0k:
Thanks for merging! And it actually closed the other issue automatically. Neat. I'll see if I find time to update the `gpuarray.sched` documentation later.

PR #449, opened 2017-06-06
Add all the missing error messages and stop filtering in pygpu.

abergeron:
I hope I did not miss a spot.

PR #452, opened 2017-06-08
Restore the IndexError in pygpu_index.

abergeron:
fix regression reported in https://github.com/Theano/libgpuarray/pull/449/files/e2f1f07accad1561db8177d85f119dfd3bcdfa2a#diff-ff48e8755d3dfca5cabbf877f62ff560

### DIFF
--- a/pygpu/gpuarray.pyx
+++ b/pygpu/gpuarray.pyx
@@ -1412,7 +1412,10 @@ cdef GpuArray pygpu_index(GpuArray a, const ssize_t *starts,
                           const ssize_t *stops, const ssize_t *steps):
     cdef GpuArray res
     res = new_GpuArray(type(a), a.context, a.base)
-    array_index(res, a, starts, stops, steps)
+    try:
+        array_index(res, a, starts, stops, steps)
+    except ValueError, e:
+        raise IndexError, "index out of bounds"
     return res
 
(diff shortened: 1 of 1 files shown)


pygpu/gpuarray.pyx
```diff
@@ -1412,7 +1412,10 @@ cdef GpuArray pygpu_index(GpuArray a, const ssize_t *starts,
                           const ssize_t *stops, const ssize_t *steps):
     cdef GpuArray res
     res = new_GpuArray(type(a), a.context, a.base)
-    array_index(res, a, starts, stops, steps)
+    try:
+        array_index(res, a, starts, stops, steps)
+    except ValueError, e:
```
nouiz: Why this is needed? Can you add a test to make sure this don't get broken again?

nouiz:
As this is a regression fix, I'll merge now. But adding the test after would be great.

nouiz:
And probably changing the error at the C level to be of the right type.

nouiz:
Should we run Theano gpu tests once to make sure to not break it again?

lamblin:
Ideally, I think we should include the tests in `theano/gpuarray` (at least) in the automatic buildbot for libgpuarray. This would make sure such issues are detected.

abergeron:
Instead of having the Theano PRs depend on the master of libgpuarray, we should depend on a tag and/or the latest release.

This way a nightly buildbot or some such catch these issues.

PR #453, opened 2017-06-12
Remove old unused stuff + flake8

abergeron:


src/gpuarray_buffer_cuda.c
```diff
@@ -1270,10 +1270,10 @@ static void _cuda_freekernel(gpukernel *k) {
   }
 }
 
-static gpukernel *cuda_newkernel(gpucontext *c, unsigned int count,
-                                 const char **strings, const size_t *lengths,
-                                 const char *fname, unsigned int argcount,
-                                 const int *types, int flags, char **err_str) {
+static int cuda_newkernel(gpukernel **k, gpucontext *c, unsigned int count,
```
nouiz: Is this a public interface?

pygpu/tools.py
```diff
@@ -168,58 +168,6 @@ def check_args(args, collapse=False, broadcast=False):
     return n, nd, dims, tuple(strs), tuple(offsets)
 
 
-class Counter(dict):
-    'Mapping where default values are zero'
-    def __missing__(self, key):
-        return 0
-
-
-def lfu_cache(maxsize=20):
```
nouiz: Why we use lru if we have lfu implemented?

pygpu/tests/support.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import print_function
 
-import os, sys
+import os
+import sys
 import numpy
```
nouiz: empty line before import numpy

pygpu/tests/test_gpu_ndarray.py
```diff
@@ -41,15 +41,16 @@ def test_hash():
...
+        hash(g)
     except TypeError as e:
         exc = e
     assert exc is not None
 
 
 def test_bool():
     for data in [numpy.empty((0, 33)), [[1]], [[0]], [], [0], [1], 0, 1]:
-        assert bool(pygpu.asarray(data, context=ctx)) == bool(numpy.asarray(data))
+        assert bool(pygpu.asarray(data, context=ctx)) == \
```
nouiz: We try to remove that syntax (\ at the end of a line) in Theano. We use paranthesis instead. I think it is more readable and other agreed on that. Why do you do differently?

src/gpuarray_buffer_cuda.c
abergeron: No

pygpu/tools.py
abergeron: I don't know.  Because that is what was working.  Also, that cache is currently only used the python code for reduction which is hopefully going away soon.

I don't want to put much effort in it.

pygpu/tests/support.py
abergeron: ok

pygpu/tests/test_gpu_ndarray.py
abergeron: It was in use at numerous locations in the test, so I removed it everywhere.

pygpu/tools.py
nouiz: k

nouiz:
travis and jenskins fail.

pygpu/gpuarray.pyx
```diff
@@ -1374,7 +1374,7 @@ cdef GpuArray pygpu_empty_like(GpuArray a, ga_order ord, int typecode):
     cdef GpuArray res
 
     if ord == GA_ANY_ORDER:
-        if py_CHKFLAGS(a, GA_F_CONTIGUOUS) and \
+        if (py_CHKFLAGS(a, GA_F_CONTIGUOUS) and
                 not py_CHKFLAGS(a, GA_C_CONTIGUOUS):
```
nouiz: You didn't added the ) and the end here.

abergeron:
I also removed REQD_WG_SIZE() since it didn't mean the same thing for OpenCL and CUDA and fixing it was non-trivial.  It is completely unused to the best of my knowledge.

PR #456, opened 2017-06-19
Make math function names in OpenCL kernel compatible with CUDA

xiaoqie:
Names of CUDA and OpenCL fp32 math functions are different.


src/gpuarray_buffer_opencl.c
```diff
@@ -320,7 +320,64 @@ static const char CL_PREAMBLE[] =
...
+  "#define logbf(x) logb(x)\n"
+  "#define logf(x) log(x)\n"
+  "#define modff(x, iptr) modf(x, iptr)\n"
+  "#define nanf(tagp) NAN\n"
+  "#define nan(tagp) NAN\n"
+  "#define nextafterf(x, y) nextafter(x, y)\n"
+  "#define powf(x, y) pow(x, y)\n"
+  "#define remainderf(x, y) remainder(x, y)\n"
+  "#define remquof(x, y, quo) remquo(x, y, quo)\n"
+  "#define rhypotf(x, y) (1/hypot(x, y))\n"
```
abergeron: This might not respect tolerance specifications.

Again I prefer unimplemented to wrong.

src/gpuarray_buffer_opencl.c
```diff
@@ -320,7 +320,64 @@ static const char CL_PREAMBLE[] =
...
+  "#define ldexpf(x, exp) ldexp(x, exp)\n"
+  "#define lgammaf(x) lgamma(x)\n"
+  "#define log10f(x) log10(x)\n"
+  "#define log1pf(x) log1p(x)\n"
+  "#define log2f(x) log2(x)\n"
+  "#define logbf(x) logb(x)\n"
+  "#define logf(x) log(x)\n"
+  "#define modff(x, iptr) modf(x, iptr)\n"
+  "#define nanf(tagp) NAN\n"
+  "#define nan(tagp) NAN\n"
```
abergeron: This does not conform to the spec.  You are supposed to return a NAN value with the tag in the mantissa, which this doesn't do.

I would rather this be unsupported than wrong.

xiaoqie:
I'll close this and make kernel code in Theano use math functions without suffix.

PR #457, opened 2017-06-20
this missing import was preventing tests from running on my machine

jonathanstrong:


nouiz:
What error this is fixing? Which version of nose do you use?

jonathanstrong:
```
$ pip freeze | rg nose
nose==1.3.7
nose-parameterized==0.5.0
```
don't have the error handy but without this import I could not run the command `EVICE=cuda0 python -c "import pygpu; pygpu.test()"`

abergeron:
This will fix the immediate problem, but there is the alternative problem that it triggers a flake8 warning.

#459 fixes the problem and avoid a flake8 warning.

abergeron:
This was addressed in #459 which was merged, but thanks for the fix.

PR #460, opened 2017-06-22
(small fix) add global inline definition for MSVC 2008

notoraptor:
This fixes a compilation error on Windows for Python 2, as `inline` keyword is now used later in this file.

@nouiz @abergeron


nouiz:
I'm good with this. But this is already the 3rd file with this code. Should we do a windows_compat.h file and move that there? I think there is also other windows compat code elsewhere, so it could be moved there.

Mostly just an idea. I'm good with merging as is. If we can't do differently today, I think we should still merge this PR today and refactor later.

abergeron:
It should probably go in src/gpuarray/config.h and be included everywhere else from there.

notoraptor:
Updated: added a commit for factorization, @nouiz @lamblin . It compiles on my computer.

src/gpuarray/wincompat/util.h
```diff
@@ -0,0 +1,20 @@
...
+    #endif
+    #ifndef snprintf
+        #define snprintf _snprintf
+    #endif
+    #ifndef strdup
+        #define strdup _strdup
+    #endif
+    #ifndef alloca
+        #define alloca _alloca
+    #endif
```
abergeron: As far as I know, we never use alloca().

nouiz:
We used it in the old back-end.

I think it is very useful, as it is faster then malloc and you don't need
to free it. I think we could replace the use of malloc/free at many places
with it. So I would be to keep it there even if not currently used :)

notoraptor:
It is very weird, as I think I have seen this `alloca` this morning when refactoring, but I can't find it anymore now.

I will check the whole C code.

notoraptor:
NB: Found ! Here: https://github.com/Theano/libgpuarray/blob/master/src/private_config.h.in#L22

And a `grep alloca -wr src` gives the following result:
```
src/gpuarray/wincompat/util.h:    #ifndef alloca
src/gpuarray/wincompat/util.h:        #define alloca _alloca
src/gpuarray_blas_cuda_cublas.c:    float **T_l = alloca(sizeof(float *) * batchCount * 3);
src/gpuarray_blas_cuda_cublas.c:    double **T_l = alloca(sizeof(double *) * batchCount * 3);
src/gpuarray_blas_cuda_cublas.c:    float **T_l = alloca(sizeof(float *) * batchCount * 3);
src/gpuarray_blas_cuda_cublas.c:    double **T_l = alloca(sizeof(double *) * batchCount * 3);
src/gpuarray_blas_cuda_cublas.c:    float **T_l = alloca(sizeof(float *) * batchCount * 3);
src/gpuarray_blas_cuda_cublas.c:    double **T_l = alloca(sizeof(double *) * batchCount * 3);
src/private_config.h.in:#define alloca _alloca
```

I will `update private_config.h.in` (I think I made a mistake this morning: I have updated `private_config.h`, not `private_config.h.in`, but the former is auto-generated from the latter). @abergeron

notoraptor:
Updated.

src/gpuarray/wincompat/util.h
abergeron: Ok we do use it.

notoraptor:
@abergeron Some keywords like `inline` are used in public headers, e.g. `ga_float2half in gpuarray/utils.h`. So, the code for Windows compatibility must be in a public part, also.

Or, should we find a way to move any platform-dependent code from public headers to private part of libgpuarray ?

abergeron:
I'm ok with making the `inline` define public, but not the others.

notoraptor:
@abergeron Done. Useful definitions were already in private config, I have just added `inline` definition into public config.

It compiles well on my computer (Windows 8.1 64 bits, Python 2.7.13 64 bits).

PR #465, opened 2017-06-26
Test the build on windows

abergeron:
This will hopefully prevent us from shipping a version that doesn't build on windows every couple of releases.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,39 @@
+version: '0.6.7.{build}'
+
+image: Visual Studio 2015
+
+init:
+  - git config --global core.autocrlf input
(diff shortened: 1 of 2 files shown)


.appveyor.yml
```diff
@@ -0,0 +1,42 @@
+version: '0.6.7.{build}'
```
nouiz: should this number be increased due to the new release?

abergeron: It has been increased as part of the release.  But in reality it doesn't really matter since it's not used for anything outside of appveyor.

nouiz:
Did you updated the how to release to tell about that?

abergeron:
Yes

PR #476, opened 2017-07-07
Loop over supported CUDA versions to find installed CUDA on Windows and Mac.

### LINKED ISSUE #472: GpuArrayException: Could not load "nvrtc64_90.dll": Le module spécifié est introuvable.
Just to report a very strange issue about libgpuarray on Windows (8.1 64 bits). It now looks for a DLL named `nvrtc64_90.dll` but it can't find it on my computer:

```
(python2) HPPC@NOTORAPTOR-HP C:\Users\HPPC\mila\dev\git\theano
> set THEANO_FLAGS=device=cuda

(python2) HPPC@NOTORAPTOR-HP C:\Users\HPPC\mila\dev\git\theano
> python -c "import theano"

Tmp 9000 Major 9 Minor 0
ERROR (theano.gpuarray): Could not initialize pygpu, support disabled
  File "theano\gpuarray\__init__.py", line 179, in <module>
    use(config.device)
  File "theano\gpuarray\__init__.py", line 166, in use
    init_dev(device, preallocate=preallocate)
  File "theano\gpuarray\__init__.py", line 65, in init_dev
    sched=config.gpuarray.sched)
  File "pygpu\gpuarray.pyx", line 634, in pygpu.gpuarray.init (pygpu/gpuarray.c:9436)
  File "pygpu\gpuarray.pyx", line 584, in pygpu.gpuarray.pygpu_init (pygpu/gpuarray.c:9127)
  File "pygpu\gpuarray.pyx", line 1057, in pygpu.gpuarray.GpuContext.__cinit__ (pygpu/gpuarray.c:13429)
GpuArrayException: Could not load "nvrtc64_90.dll": Le module spécifié est introuvable.
```

Looking into NVIDIA bin folder (`C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v8.0\bin`), I found `nvrtc64_80.dll` but not `nvrtc64_90.dll`:

image

**NB**: it seems the problem comes from library version detection, based on the value returned by `cuDriverGetVersion()`:
https://github.com/Theano/libgpuarray/blob/master/src/gpuarray_buffer_cuda.c#L143
https://github.com/Theano/libgpuarray/blob/master/src/loaders/libnvrtc.c#L30

But it worked some days ago. I just reminded that I have updated my NVIDIA driver some days ago too, but it should not cause any problem, should it ?

@abergeron @nouiz @lamblin

notoraptor:
Fix #472 . Tested on my Windows 8.1 64 bits with Python 2.7.13 Anaconda 64 bits.

When compiled as Debug version:
```
(python2) HPPC@NOTORAPTOR-HP C:\Users\HPPC\mila\dev\git\theano
> echo %THEANO_FLAGS%
device=cuda0

(python2) HPPC@NOTORAPTOR-HP C:\Users\HPPC\mila\dev\git\theano
> python -c "import theano"
ERROR 18: Could not load "nvrtc64_90.dll": Le module spécifié est introuvable.

Detected CUDA 8.0.
Loading cuBLAS 8.0.
Using cuDNN version 6020 on context None
Mapped name None to device cuda0: GeForce 840M (0000:0A:00.0)
```

@nouiz @abergeron @lamblin .

### DIFF
--- a/src/gpuarray_buffer_cuda.c
+++ b/src/gpuarray_buffer_cuda.c
@@ -132,7 +132,6 @@ static int setup_lib(void) {
   const char *ver;
   CUresult err;
   int res, tmp;
-  int search_version = 0;
 
   if (!setup_done) {
     res = load_libcuda(global_err);
@@ -141,55 +140,24 @@ static int setup_lib(void) {
     err = cuInit(0);
     if (err != CUDA_SUCCESS)
       return error_cuda(global_err, "cuInit", err);
-    ver = getenv("GPUARRAY_CUDA_VERSION");
-    if (ver == NULL || strlen(ver) != 2) {
-      err = cuDriverGetVersion(&tmp);
-      if (err != CUDA_SUCCESS)
-        return error_set(global_err, GA_IMPL_ERROR, "cuDriverGetVersion failed");
-      major = tmp / 1000;
-      minor = (tmp / 10) % 10;
-      #if defined(WIN32) || defined(_WIN32) || defined(WIN64) || defined(_WIN64) || defined(__APPLE__)
-      /* We will dynamically search the right CUDA version only on Windows and Macintosh systems,
-      and only if user has not explicitely specified GPUARRAY_CUDA_VERSION. */
-      search_version = 1;
-      #endif
-    } else {
-      major = ver[0] - '0';
-      minor = ver[1] - '0';
-    }
-    /* NB: next line will cause problems if a CUDA 10.0 (or 9.11) is released in the future. */
-    if (major > 9 || major < 0 || minor > 9 || minor < 0)
-      return error_fmt(global_err, GA_VALUE_ERROR, "Invalid cuda version: %d.%d", major, minor);
-    if (!search_version) {
-      res = load_libnvrtc(major, minor, global_err);
-    } else {
-      /* First case in next array is reserved to eventually receive the version returned by cuDriverGetVersion(). */
-      int versions[] = {-1, 80, 75};
-      int versions_length = sizeof(versions) / sizeof(int);
-      int current_version = major * 10 + minor;
+    err = cuDriverGetVersion(&tmp);
+    if (err != CUDA_SUCCESS)
+      return error_set(global_err, GA_IMPL_ERROR, "cuDriverGetVersion failed");
+    major = tmp / 1000;
+    minor = (tmp / 10) % 10;
+    /* Let's try to load a nvrtc corresponding to detected CUDA version. */
+    res = load_libnvrtc(major, minor, global_err);
+    if (res != GA_NO_ERROR) {
+      /* Else, let's try to find a nvrtc corresponding to supported CUDA versions. */
+      int versions[][2] = {{8, 0}, {7, 5}, {7, 0}};
+      int versions_length = sizeof(versions) / (2 * sizeof(int));
       int i = 0;
-      for (i = 1; i < versions_length && versions[i] != current_version; ++i);
-      if (i == versions_length) {
-        /* Current version not found in the list of versions. We add it at top of the list. */
-        versions[0] = current_version;
-        /* We will iterate on versions from the first. */
-        i = 0;
-      } else {
-        /* Current version found in the list of known versions. No need to add it to the list. */
-        i = 1;
-      };
       do {
-        major = versions[i] / 10;
-        minor = versions[i] % 10;
+        major = versions[i][0];
+        minor = versions[i][1];
         res = load_libnvrtc(major, minor, global_err);
         ++i;
       } while(res != GA_NO_ERROR && i < versions_length);
-      #ifdef DEBUG
-      if (res == GA_NO_ERROR)
-        fprintf(stderr, "Detected CUDA %d.%d.\n", major, minor);
-      else
-        fprintf(stderr, "Unable to detect a CUDA version.\n");
-      #endif
     }
     if (res != GA_NO_ERROR)
       return res;
--- a/src/gpuarray_buffer_cuda.c
+++ b/src/gpuarray_buffer_cuda.c
@@ -150,7 +150,7 @@ static int setup_lib(void) {
     if (res != GA_NO_ERROR) {
       /* Else, let's try to find a nvrtc corresponding to supported CUDA versions. */
       int versions[][2] = {{8, 0}, {7, 5}, {7, 0}};
-      int versions_length = sizeof(versions) / (2 * sizeof(int));
+      int versions_length = sizeof(versions) / sizeof(versions[0]);
       int i = 0;
       do {
         major = versions[i][0];
--- a/src/gpuarray_buffer_cuda.c
+++ b/src/gpuarray_buffer_cuda.c
@@ -129,7 +129,6 @@ static int setup_done = 0;
 static int major = -1;
 static int minor = -1;
 static int setup_lib(void) {
-  const char *ver;
   CUresult err;
   int res, tmp;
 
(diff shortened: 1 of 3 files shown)


src/gpuarray_buffer_cuda.c
```diff
@@ -147,13 +148,49 @@ static int setup_lib(void) {
         return error_set(global_err, GA_IMPL_ERROR, "cuDriverGetVersion failed");
       major = tmp / 1000;
       minor = (tmp / 10) % 10;
+      #if defined(WIN32) || defined(_WIN32) || defined(WIN64) || defined(_WIN64) || defined(__APPLE__)
```
nouiz: why not also on linux?

src/gpuarray_buffer_cuda.c
```diff
@@ -147,13 +148,49 @@ static int setup_lib(void) {
...
+      /* We will dynamically search the right CUDA version only on Windows and Macintosh systems,
+      and only if user has not explicitely specified GPUARRAY_CUDA_VERSION. */
+      search_version = 1;
+      #endif
     } else {
       major = ver[0] - '0';
       minor = ver[1] - '0';
     }
+    /* NB: next line will cause problems if a CUDA 10.0 (or 9.11) is released in the future. */
```
nouiz: good catch. Make an issue about that. cuda 9 is annouced, so in 1 or 2 year we should have an cuda 10 I guess.

src/gpuarray_buffer_cuda.c
```diff
@@ -147,13 +148,49 @@ static int setup_lib(void) {
...
     if (major > 9 || major < 0 || minor > 9 || minor < 0)
       return error_fmt(global_err, GA_VALUE_ERROR, "Invalid cuda version: %d.%d", major, minor);
-    res = load_libnvrtc(major, minor, global_err);
+    if (!search_version) {
```
nouiz: I'll let @abergeron review this part.

src/gpuarray_buffer_cuda.c
notoraptor: Based on code in libnvrtc.c ( https://github.com/Theano/libgpuarray/blob/master/src/loaders/libnvrtc.c#L49 ), we just look for `libnvrtc.so` on Linux (without explicitely specifying a version).

src/gpuarray_buffer_cuda.c
notoraptor: Ok!

src/gpuarray_buffer_cuda.c
```diff
@@ -147,13 +148,49 @@ static int setup_lib(void) {
...
+      res = load_libnvrtc(major, minor, global_err);
+    } else {
+      /* First case in next array is reserved to eventually receive the version returned by cuDriverGetVersion(). */
+      int versions[] = {-1, 80, 75};
```
abergeron: We still support 7.0.

Also, since cuda 10 is not so far away, make this a list of pair of ints instead to explicitly list the majors/minors that we support.

src/gpuarray_buffer_cuda.c
abergeron: It is true that we only look at a specific file for windows, but I would rather not spread the assumption here.

We may devise a way to load a specific version of cuda on linux/mac later, and I would rather that the necessary components are in place at the upper levels.  Also this makes the code more complex for no benefit.

src/gpuarray_buffer_cuda.c
abergeron: If we can look for all supported versions, then we can probably drop the environment variable, which would solve the problem.

src/gpuarray_buffer_cuda.c
```diff
@@ -147,13 +148,49 @@ static int setup_lib(void) {
...
+      for (i = 1; i < versions_length && versions[i] != current_version; ++i);
+      if (i == versions_length) {
+        /* Current version not found in the list of versions. We add it at top of the list. */
+        versions[0] = current_version;
+        /* We will iterate on versions from the first. */
+        i = 0;
+      } else {
+        /* Current version found in the list of known versions. No need to add it to the list. */
+        i = 1;
+      };
```
abergeron: Can we just try a load with the minor/major from cuDriverGetVersion and then fallback to the list.  This seems overly complicated.

src/gpuarray_buffer_cuda.c
```diff
@@ -147,13 +148,49 @@ static int setup_lib(void) {
...
+        major = versions[i] / 10;
+        minor = versions[i] % 10;
+        res = load_libnvrtc(major, minor, global_err);
+        ++i;
+      } while(res != GA_NO_ERROR && i < versions_length);
+      #ifdef DEBUG
+      if (res == GA_NO_ERROR)
+        fprintf(stderr, "Detected CUDA %d.%d.\n", major, minor);
+      else
+        fprintf(stderr, "Unable to detect a CUDA version.\n");
```
abergeron: I would probably push that message back in the loader for nvrtc, similar to what you did for cublas.  The code in gpuarray_buffer_cuda is already kind of too large.

src/gpuarray_buffer_cuda.c
notoraptor: By `list of pair of ints`, do you mean something like
```C
int versions[][2] = {{8, 0}, {7, 5}, {7,  0}};
```
?

abergeron: If that works, yes.

src/gpuarray_buffer_cuda.c
notoraptor: I did that to avoid testing twice a same version that is not installed (e.g. if cuDriverGetVersion() returns 80, then 80 will already be in the list of supported versions).

Do you anyway want the fallback strategy ?

notoraptor:
@abergeron updated. This should be tested on Mac also. Check how I support >9 versions.

NB: for a CUDA version like `10.1`, current code will look for `nvrtc64_101.dll`. For a version like `10.12`, it will look for `nvrtc64_1012.dll` (pattern `%d%d`). Not sure if it is the right pattern to use (maybe should I consider that `10.1 => 1010` and `10.12 => 1012`, or rather `10.1 => 1001` and `10.12 => 1012` ?). I don't know if NVIDIA may release a minor version > 9.

src/gpuarray_buffer_cuda.c
```diff
@@ -140,20 +140,25 @@ static int setup_lib(void) {
...
+    if (err != CUDA_SUCCESS)
+      return error_set(global_err, GA_IMPL_ERROR, "cuDriverGetVersion failed");
+    major = tmp / 1000;
+    minor = (tmp / 10) % 10;
+    /* Let's try to load a nvrtc corresponding to detected CUDA version. */
     res = load_libnvrtc(major, minor, global_err);
+    if (res != GA_NO_ERROR) {
+      /* Else, let's try to find a nvrtc corresponding to supported CUDA versions. */
+      int versions[][2] = {{8, 0}, {7, 5}, {7, 0}};
+      int versions_length = sizeof(versions) / (2 * sizeof(int));
```
abergeron: rather than `2 * sizeof(int)` use `sizeof(versions[0])`

src/loaders/libcublas.c
```diff
@@ -41,21 +45,22 @@ int load_libcublas(int major, int minor, error *e) {
...
-    char libname[] = "cublas64_??.dll";
+    const char* libname_pattern = "cublas64_%d%d.dll";
+    char libname[64];
 
-    libname[9] = DIGITS[major];
-    libname[10] = DIGITS[minor];
+    #ifdef DEBUG
+    fprintf(stderr, "Loading cuBLAS %d.%d.\n", major, minor);
+    #endif
+    sprintf(libname, libname_pattern, major, minor);
```
abergeron: I object to sprintf.  Use snprintf at the minimum.

Also check for errors and report them.  We don't want to use a truncated name for loading.

Finally, have the format string in the arguments to snprintf().

abergeron:
We'll see about the name pattern for cuda 10 when it is released.  No need to worry too much about this now.

notoraptor:
@abergeron Updated.

abergeron:
Code was mostly fine, I was just tired of back and forth for little gain, so I fixed the few remaining style issues.

notoraptor:
Tests succeeded (also tested on my computer).

PR #480, opened 2017-07-18
Fix kernel cache

abergeron:
This should fix the problem that if you compile multiple kernels from the same source you will only get the first one back.


abergeron:
I'm waiting on a review of this for the 0.6.9 release.

lamblin:
Looks reasonable.

Issue #481, opened 2017-07-19
Build fails on 32-bit architectures


ghisvail:
In my effort to resurrect the packaging of `libgpuarray` / `pygpu` for Debian, I noticed the software fails to build on 32-bit architectures. See the following [page](https://buildd.debian.org/status/package.php?p=libgpuarray) for the logs. The errors are of the following nature:

```
/<<PKGBUILDDIR>>/src/util/integerfactoring.c:271:20: error: '__int128' is not supported on this target
  return ((unsigned __int128)a * (unsigned __int128)b) % m;
```

If 32-bit builds are not supported, then it may be wise to at least update the `Build Requirements` section of the docs appropriately, and have CMake fail early and not even attempt a build if the platform is 32-bit. Otherwise, please advise on a workaround for the build.

Cheers,

abergeron:
I will admit that 32-bits builds are sort of a second-tier platform for us.

The main reason is that such build are of questionable utility since there are no CUDA runtime for 32-bits platform.  However, there may be OpenCL builds so you could still do something useful.

I have no idea how to fix the build right now, but I'll look into it.

ghisvail:
> However, there may be OpenCL builds so you could still do something useful.

Indeed.

> I have no idea how to fix the build right now, but I'll look into it.

Would it be possible to make the inclusion of 128-bit arithmetic conditional on whether the host platform has supports it?

abergeron:
I'm not sure.  I'll take a look and figure it out.

ghisvail:
Please have a look at the following [bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=868877) report filed on the Debian BTS. There are potentially useful comments.

abergeron:
@obilaniu Fixed the problem with __int128 in #483 

Once that is merged we will probably do a 0.6.9 soon-ish.

obilaniu:
@ghisvail Why is my GCC inline assembler code broken? It certainly works when I try it. E.g.

    gaIMulMod(0xBEEFDEADBEEFDEAD, 0xBEEFDEADBEEFDEAD, 0xDEADBEEFDEADBEEF) == 9006955852451488416

And if it were done in 64 bits that would totally blow up.

[...]

PR #483, opened 2017-07-19
Trivial Fixes related to #481.

obilaniu:
A compiler error fix for #481, plus many errors reported when `-Wall -Wextra` is in use, including all the errors [here](https://buildd.debian.org/status/fetch.php?pkg=libgpuarray&arch=i386&ver=0.6.8-1&stamp=1500376646&raw=0).

Note that certain warnings appear only at `-O0` and some only at `-O2`, so it's important to build in both release and debug mode to see fresh warnings.

### DIFF
--- a/src/cache/lru.c
+++ b/src/cache/lru.c
@@ -114,15 +114,14 @@ struct _hash {
   size_t size;
 };
 
-static inline size_t roundup2(size_t s) {
+static inline unsigned long long roundup2(unsigned long long s) {
   s--;
   s |= s >> 1;
   s |= s >> 2;
   s |= s >> 4;
   s |= s >> 8;
   s |= s >> 16;
-  if (sizeof(size_t) >= 8)
-    s |= s >> 32;
+  s |= s >> 32;
   s++;
   return s;
 }
--- a/src/cache/twoq.c
+++ b/src/cache/twoq.c
@@ -124,15 +124,14 @@ struct _hash {
   size_t size;
 };
 
-static inline size_t roundup2(size_t s) {
+static inline unsigned long long roundup2(unsigned long long s) {
   s--;
   s |= s >> 1;
   s |= s >> 2;
   s |= s >> 4;
   s |= s >> 8;
   s |= s >> 16;
-  if (sizeof(size_t) >= 8)
-    s |= s >> 32;
+  s |= s >> 32;
   s++;
   return s;
 }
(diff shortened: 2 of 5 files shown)


src/cache/lru.c
```diff
@@ -114,15 +114,14 @@ struct _hash {
...
+static inline unsigned long long roundup2(unsigned long long s) {
   s--;
   s |= s >> 1;
   s |= s >> 2;
   s |= s >> 4;
   s |= s >> 8;
   s |= s >> 16;
-  if (sizeof(size_t) >= 8)
-    s |= s >> 32;
+  s |= s >> 32;
```
abergeron: Won't that trigger something about narrowing when size_t is 32 bits?  Maybe moving the condition to an ifdef would help, although I don't know if it will work.

obilaniu: @abergeron But it did trigger a warning, which is why I rewrote the code so it uses `unsigned long long`

abergeron: Would it be possible to use an `#ifdef`?

obilaniu: Not in a truly portable manner... GCC offers
```
> gcc -E -dM -m64 - < /dev/null | grep __SIZEOF
#define __SIZEOF_INT__ 4
#define __SIZEOF_POINTER__ 8
#define __SIZEOF_LONG__ 8
#define __SIZEOF_LONG_DOUBLE__ 16
#define __SIZEOF_SIZE_T__ 8
#define __SIZEOF_WINT_T__ 4
#define __SIZEOF_PTRDIFF_T__ 8
#define __SIZEOF_FLOAT__ 4
#define __SIZEOF_SHORT__ 2
#define __SIZEOF_INT128__ 16
#define __SIZEOF_WCHAR_T__ 4
#define __SIZEOF_DOUBLE__ 8
#define __SIZEOF_LONG_LONG__ 8
```

But MSVC doesn't offer those or any equivalents that I know. We could enumerate 64-bit architectures, like `__x86_64__`, but that list would be fragile, would never be exhaustive and could vary by compiler.

I would strongly prefer the `unsigned long long` solution. C99 requires it to be at least 64 bits wide, and even the one heretic compiler we care about that doesn't officially support C99 does support `unsigned long long` and does make it 64-bits wide.

abergeron: ok then

PR #494, opened 2017-08-01
Stop comparing to None or Ellipsis with '==' (or equivalent)

### LINKED ISSUE #493: Fix deprecation warnings for comparison of array == None
The behaviour changed with numpy 1.13.

I saw the deprecation warnings using `.__setitem__` on `gpuarray` objects.

abergeron:
Fixes #493

### DIFF
--- a/pygpu/gpuarray.pyx
+++ b/pygpu/gpuarray.pyx
@@ -40,6 +40,15 @@ cdef bytes _s(s):
         return s
     raise TypeError("Expected a string")
 
+cdef size_t countis(l, object val):
+    cdef size_t count
+    cdef size_t i
+    count = 0
+    for i in range(len(l)):
+        if l[i] is val:
+            count += 1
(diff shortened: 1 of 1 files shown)


pygpu/gpuarray.pyx
```diff
@@ -40,6 +40,15 @@ cdef bytes _s(s):
         return s
     raise TypeError("Expected a string")
 
+cdef size_t countis(l, object val):
```
nouiz: Is it useful to test numpy version and use the numpy count for old numpy to don't slow down setitem and getitem? Those 2 methods was optimized a few times, it would not be great that this new implementation slow them down.

abergeron: I think this will be faster, even on older numpy since it compares with `is`, not `==`.  It's also C code so it won't be that slow in any case.

PR #499, opened 2017-08-16
API-breaking changes planned for 0.7

abergeron:


src/gpuarray/buffer.h
```diff
@@ -550,13 +550,7 @@ GPUARRAY_PUBLIC gpucontext *gpukernel_context(gpukernel *k);
  */
 #define GA_CTX_PROP_DEVNAME  1
 
-/**
```
nouiz: why remove them? I think they are still needed.

abergeron: This is exactly the same as GA_CTX_PROP_MAXLSIZE0.  So I want to remove this alias since it may be confused for the maximum total local size.

src/cluda_opencl.h.c
```diff
@@ -0,0 +1,107 @@
+static const char cluda_opencl_h[] = {
```
nouiz: what is this?

src/cluda_opencl.h
```diff
@@ -0,0 +1,40 @@
+#define local_barrier() barrier(CLK_LOCAL_MEM_FENCE)
```
nouiz: I think it would be better to wrap this into something like:
```
#ifndef CLUDA_H
...
#define CLUDA_H
#endif
```

src/head.py
```diff
@@ -0,0 +1,35 @@
+def wrt(f, n, b):
```
nouiz: Add comment to tell what this file does and when it should be used.

src/gpuarray_buffer_cuda.c
```diff
@@ -322,7 +322,8 @@ cuda_context *cuda_make_ctx(CUcontext ctx, gpucontext_props *p) {
   }
   res->errbuf->flags |= CUDA_MAPPED_PTR;
   /* Prime the cache */
-  cuda_alloc((gpucontext *)res, p->initial_cache_size, NULL, 0);
+  if (p->initial_cache_size)
```
nouiz: I forgot the detail, but why this don't cause a memory leak? We want to just fill the cache, not use it.

tests/check_array.c
```diff
@@ -27,20 +27,20 @@ START_TEST(test_take1_ok) {
                               18, 19, 20, 21, 22, 23};
   uint32_t buf[12 * 24];
   const size_t data_dims[1] = {24};
-  ssize_t indexes[12];
```
nouiz: Why did you remove the ability to use those dtype on the GPU?

src/CMakeLists.txt
```diff
@@ -14,6 +14,22 @@ add_custom_command(
   WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
   DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/gen_types.py)
 
+add_custom_command(
```
nouiz: Why do you continue to keep them in the repo if you regenete them automatically? I would prefer to remove them from the repo.

pygpu/gpuarray.pxd
```diff
@@ -172,10 +172,6 @@ cdef extern from "gpuarray/array.h":
                           gpudata *data, size_t offset, int typecode,
                           unsigned int nd, const size_t *dims,
                           const ssize_t *strides, int writable)
-    int GpuArray_copy_from_host(_GpuArray *a,
```
nouiz: What was the bug? crash? bad values?

pygpu/reduction.py
```diff
@@ -37,7 +37,8 @@ def _ceil_log2(x):
 
 #define REDUCE(a, b) (${reduce_expr})
 
-KERNEL void ${name}(const unsigned int n, ${out_arg.decltype()} out
+KERNEL void ${name}(const unsigned int n, ${out_arg.decltype()} out,
```
nouiz: why the offset is needed on the output?

src/gpuarray_buffer_cuda.c
```diff
@@ -37,6 +37,9 @@ STATIC_ASSERT(sizeof(GpuArrayIpcMemHandle) == sizeof(CUipcMemHandle), cuda_ipcme
  */
 #define FRAG_SIZE (64)
 
+extern gpuarray_blas_ops cublas_ops;
```
abergeron: I don't want to expose it as a property anymore.

src/gpuarray_buffer_cuda.c
abergeron: Yes, I'll fix it.

src/gpuarray_buffer_cuda.c
```diff
@@ -758,7 +776,7 @@ static gpudata *cuda_alloc(gpucontext *c, size_t size, void *data, int flags) {
    * to a multiple of FRAG_SIZE.  This also ensures that if we split a
    * block, the next block starts properly aligned for any data type.
    */
-  if (!(ctx->flags & GA_CTX_DISABLE_ALLOCATION_CACHE)) {
+  if (ctx->max_cache_size != 0) {
```
abergeron: max_cache_size = SIZE_T_MAX (or anything bigger than the GPU memory), initial_cache_size = 0

src/gpuarray_buffer.c
```diff
@@ -37,15 +37,84 @@ int gpu_get_device_count(const char* name, unsigned int platform,
...
+  return GA_NO_ERROR;
+}
+
+int gpucontext_props_sched(gpucontext_props *p, int sched) {
+  if (sched == GA_CTX_SCHED_MULTI)
+    FLSET(p->flags, GA_CTX_MULTI_THREAD);
+  else
+    FLCLR(p->flags, GA_CTX_MULTI_THREAD);
+
+  switch (sched) {
```
abergeron: ok

src/gpuarray_buffer.c
```diff
@@ -37,15 +37,84 @@ int gpu_get_device_count(const char* name, unsigned int platform,
...
-  return res;
+  gpucontext *r;
+  if (ops == NULL) {
+    gpucontext_props_del(p);
+    return global_err->code;
+  }
+  if (p == NULL && gpucontext_props_new(&p) != GA_NO_ERROR)
+    return global_err->code;
+  r = ops->buffer_init(p);
+  gpucontext_props_del(p);
```
abergeron: This is an ease of use concern.  I wouldn't make that the default way objects are handled, but I think this is a reasonable exception.

You could say that once you pass the props object to gpucontext_init() it becomes the property of the context and it will manage its (very short) lifetime.

src/cluda_opencl.h.c
abergeron: This is to embed the source code in cluda_opencl.h in the binary so that we can use it for kernels.

src/cluda_opencl.h
abergeron: ok

tests/check_array.c
abergeron: That is groundwork for call32.  Also I didn't remove all uses of the type, just the ability to make arrays, because arrays of a type that can change size is a bad idea.

src/CMakeLists.txt
abergeron: This is intended for developers that will modify the .h files directly.  If you just build from the source, it shouldn't have to regenerate since the files are unmodified.

pygpu/gpuarray.pxd
abergeron: There was some bad values, but it's also an ill-concieved idea from the start.

The idea was to avoid copying data on the CPU during a transfer so it would figure out the start and the end of the passed-in data (using strides and shapes) and copy the whole block.  This would potentially copy large swaths of irrelevant data that just happen to sit between those boundaries.

It also bypassed the forced offset that is added in DEBUG and create some problems related to that.

pygpu/reduction.py
abergeron: Because it's an array and therefore it can have an offset.

src/gpuarray/buffer.h
nouiz: I got confused myself! So what is the name for the maximum total local size? Is it available?

src/gpuarray_buffer.c
nouiz: ok, but document that well as this is non-standard.

This type of thing can be good, but can also cause problems as some non-stardard stuff in the NumPy C API.

tests/check_array.c
nouiz: ok

src/head.py
nouiz: Can you add how to use that script in the comment? Put the command line you use yourself.

nouiz:
All CI failed

src/gpuarray/buffer.h
abergeron: There isn't one currently.  But we can expose it if required.

src/gpuarray_buffer_cuda.c
abergeron: The theano PR does this correctly now, since the problem was in pygpu.

I believe there is documentation about what max_cache_size and initial_cache_size mean, and if not, I should add it.  But I wouldn't explicitly document this combination since it follows naturally from the properties of both limits.

src/gpuarray_buffer.c
abergeron: I'll add some documentation about that.

src/head.py
abergeron: ok

src/gpuarray_buffer_cuda.c
abergeron: Ok there is no documentation, but I'll add some.

nouiz:
I'm not able to run libgpuarray tests:
```
.../repos/libgpuarray/Build$ DEVICE=cuda0 make test
Running tests...
No tests were found!!!
```

this was working before. This is taken from doc/installation.txt

src/gpuarray/buffer.h
nouiz: ok

src/gpuarray_buffer_cuda.c
```diff
@@ -322,8 +322,11 @@ cuda_context *cuda_make_ctx(CUcontext ctx, gpucontext_props *p) {
   }
   res->errbuf->flags |= CUDA_MAPPED_PTR;
   /* Prime the cache */
-  if (p->initial_cache_size)
-    cuda_alloc((gpucontext *)res, p->initial_cache_size, NULL, 0);
+  if (p->initial_cache_size) {
+    gpudata *tmp = cuda_alloc((gpucontext *)res, p->initial_cache_size, NULL, 0);
```
nouiz: IRL answer, opencl don't have a cache

nouiz:
There is still a usage of a function you removed:

tests/check_collectives.c:  err = GpuArray_copy_from_host(&Adev, ctx, A, GA_INT, ND, indims, instrds); \

nouiz:
@abergeron told that jenkins probably don't have MPI installed, so didn't compiled collective tests.

@slefrancois when you have time, add that to jenkins installation.

PR #502, opened 2017-08-21
Changes for CUDA 9.0 float16 support.

abergeron:
This integrates the relevant parts of #485 with the changes in #499.

#499 should be merged first.


borisfom:
@abergeron : looks good to me.

borisfom:
@abergeron : what is your setup? We also have collectives crash (garbage returned in the key on init) for a long time.

nouiz:
IRL discussion, he use cuda9. On cuda8 I have the crash too. As we can
reproduce it, we probably can make it work.

abergeron:
Apparently the buffer code was broken, so I just removed it.  It was just fancy code to be able to modify the value of the ID directly, but we can still get and set it as a bytearray from the `.comm_id` property so it's not a great loss.

borisfom:
Thanks @abergeron

borisfom:
Thanks @abergeron !

nouiz:
@tsirif what do you think about the crash fix?

We need to make the release, so merging. But if needed, this can be added back later.

tsirif:
No, I do not think it is needed either, probably I should not have put that in the first place.

nouiz:
thanks

PR #512, opened 2017-09-01
Better conda

abergeron:


abergeron:
This is where I'll stop for now.

I changed the appveyor build to make conda packages for python 2.7 and 3.6 of libgpuarray and pygpu.  They are kept as "build artifacts" on the appveyor build page.

There is no automatic upload yet, but it could be added in a subsequent PR.

Also kept for later: automatic package builds for mac and linux.

conda/pygpu/meta.yaml
```diff
@@ -19,14 +18,14 @@ requirements:
     - git            [not win]
     - python
     - cython >=0.25
-    - numpy x.x
+    - numpy 1.11
```
nouiz: Does it mean it will only support numpy 1.11? Why not bump that to 1.13 to support the more recent version?

abergeron: No, it was discovered that numpy is forward-compatible.  So it should work with numpy 1.12 and 1.13 too (and 1.14 whenever it comes out).

PR #515, opened 2017-09-05
Fix versionning for auto-generated conda packages

notoraptor:
Sub-module `pygpu.version` is required in Theano but does not exist anymore in pygpu. Pygpu now uses `versioneer` to manage versions, but this module seems to auto-generate only a version string, it does not provide itself some symbols like `major`, `minor`, `patch`, etc.

I suggest to re-introduce the module `pygpu.version`, but as a standalone module (ie. not auto-generated) which parses the actual version string computed by `versioneer`.

@nouiz @abergeron


notoraptor:
**NB (note)**: Current conda package for `pygpu` on Windows has a bad version string generated by `versioneer`, and I don't know why:

```
(test2) C:\Users\notoraptor\mila\dev>conda install -c mila-udem pygpu
Fetching package metadata .............
Solving package specifications: .

Package plan for installation in environment C:\Users\notoraptor\data\programs\Anaconda3\envs\test2:

The following NEW packages will be INSTALLED:

    pygpu: 0.7.1-py27hb1fb574_0 mila-udem

Proceed ([y]/n)?


(test2) C:\Users\notoraptor\mila\dev>
(test2) C:\Users\notoraptor\mila\dev>python -c "import pygpu; print(pygpu.__version__)"
0+unknown
```

On my computer (Windows 10 64 bits), version string is correctly generated when I compile libgpuarray+pygpu from master branch (`0.7.1+1.g78c7b1845`).

notoraptor:
I don't yet know why pygpu version is not well computed when building conda package, but I remark this error in appveyor output ( https://ci.appveyor.com/project/abergeron/libgpuarray-jv775/build/1.0.28-ttxrxnen/job/e78xf795w19igpib ) :
```
fatal: --dirty is incompatible with commit-ishes
```

Maybe should we make sure that appveyor runs with a branch not considered to be `dirty` ?

nouiz:
That would be great, but a second after having working conda packages.

When CI execute on a PR, they freuquently  merge the PR to the master. I
suppose this is what they consider "dirty". So maybe we should activate the
CI on the master branch and not just on PR?

nouiz:
I think the current Appveyer have access only to @abergeron. At least, I
didn't found how to use it. It don't seem to automatically give access to
via github access as travis does. So maybe you will need to make you an
appveyer account and configure it like it is needed there.

abergeron:
For the issue on AppVeyor: https://github.com/warner/python-versioneer/issues/122

notoraptor:
Ok, this should make Appveyor generate conda packages with right version. Let's see if it works.

To do that I modified `versioneer.py` according to first comment in the issue mentioned by @abergeron : https://github.com/warner/python-versioneer/issues/122 .

I don't know if changes applied to `versioneer.py` may break any other versionning functionnality (it seems not, currently). So I suggest we don't forget to find a better way to fix this problem (later, after we finish the releases).

@nouiz @abergeron

notoraptor:
Now it seems to work ! https://ci.appveyor.com/project/abergeron/libgpuarray-jv775/build/1.0.28-odhhtqfs/job/a1c0ho3yoxl5jafl

notoraptor:
Updated, workaround restricted to Windows only @nouiz  .

notoraptor:
Appyevor passed, version still right ! https://ci.appveyor.com/project/abergeron/libgpuarray-jv775/build/1.0.28-vcnjgqcq

PR #518, opened 2017-09-08
Try to fix theano error related to sgerBatch.

notoraptor:
On my computer, this seems to fix errors reported here: http://earlgrey.iro.umontreal.ca:8080/job/Theano_pr_win/44/

Parameter `A` for `sgerBatch` should be the one that is read and written, but current code call `cuda_record` to that parameter with flag `CUDA_WAIT_READ`. This PR fixes it.

NB: I have not applied the fix to the other similar functions (e.g. `dgerBatch`, etc.), I prefer post the PR first, so that we can discuss about the problems. If it is really the right error, then I will fix it in the other functions.

@nouiz @abergeron @lamblin


notoraptor:
Updated for `dgerBatch`. `sger` and `dger` seems already right.

PR #519, opened 2017-09-11
Test another possible fix for versionning

notoraptor:
Let's see if appveyor generates the right version number for pygpu with this fix.

@abergeron

### DIFF
--- a/conda/pygpu/meta.yaml
+++ b/conda/pygpu/meta.yaml
@@ -13,9 +13,6 @@ build:
 
 requirements:
   build:
-    - m2-git         [win]
-    - m2-filesystem  [win]
-    - git            [not win]
     - python
     - cython >=0.25
     - numpy 1.11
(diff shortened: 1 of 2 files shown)


notoraptor:
This fix seems to work ! https://ci.appveyor.com/project/abergeron/libgpuarray-jv775/build/1.0.30-tlhouddp

@nouiz @abergeron

conda/pygpu/meta.yaml
```diff
@@ -13,9 +13,6 @@ build:
 
 requirements:
   build:
-    - m2-git         [win]
```
nouiz: why this fix the problem?

Also, we need to make the build fail if it can't generate a good version.

abergeron: Because it is that git that doesn't like --dirty and gets used for the setup.py version.

Also we don't need it since we grab the source from the existing checkout.

abergeron:
We can ignore the travis failure, that's fixed in another PR.

PR #521, opened 2017-09-11
Upload packages for release automatically.

abergeron:


abergeron:
I think I have the script down ok, however the only real way to test is to merge the PR and push a tag.

This is because the "secure" environment variables are not available during PR build so we can't make a real test run there.

abergeron:
Since this is packaging only, I'll merge this

Issue #486, opened 2017-07-21
pygpu.gpuarray.GpuArrayException: (b'malloc: No such file or directory', 6)


Kay1794:
Hi all,
Just like the title said I could not pass the pygpu.test() when I ran in to those errors.
I don't know what happened and it seems there's no one has met the same problems. All the errors I met are in the same type.
Any help will be so appreciated!

`
ERROR: pygpu.tests.test_blas.test_rgemmBatch_3d(16, 16, 9, 16, 'float32', ('f', 'f', 'c'), (False, False), False, 1, True, True, 0.6, 0.6)

Traceback (most recent call last):
File "/home/rcf-proj/yz6/miniconda3/envs/corossion/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
self.test(*self.arg)
File "/home/rcf-proj/yz6/miniconda3/envs/corossion/lib/python3.6/site-packages/pygpu/tests/support.py", line 44, in f
func(*args, **kwargs)
File "/home/rcf-proj/yz6/miniconda3/envs/corossion/lib/python3.6/site-packages/pygpu/tests/test_blas.py", line 229, in rgemmBatch_3d
trans_b=trans[1], overwrite_c=overwrite)
File "pygpu/blas.pyx", line 209, in pygpu.blas.gemmBatch_3d (pygpu/blas.c:4366)
File "pygpu/blas.pyx", line 66, in pygpu.blas.pygpu_blas_rgemmBatch_3d (pygpu/blas.c:2327)
pygpu.gpuarray.GpuArrayException: (b'malloc: No such file or directory', 6)`

abergeron:
Which version of pygpu do you have?

Also which device are you using to test with?

Kay1794:
Hi, thanks for replying.
I use the newest version on a tesla k40.
The device works well when I am using pytorch so the cuda should already be set properly right?

abergeron:
Please tell me the exact version, "newest version" can mean different things depending on where you install it from and when you did your install.

Also a couple of other questions:
 - What is the version of cuda that you have installed?
 - Are you using the k40 through cuda or opencl?
 - Is the error reproducible?

shawntan:
I'm facing a similar problem as well. 
Version: 0.7.1+8.g342016c

I first came across this error using batch_tensordot, and I followed the problem all the way back to libgpuarray.
```
======================================================================
[...]

PR #522, opened 2017-09-12
Fixing src/gpuarray_array_blas.c

dendisuhubdy:
fixing gpuarray_array_blas.c, C_OFFSET value comparison with NULL, fixing GpuArrayException: ('malloc: Resource temporarily unavailable', 6) @shawntan @obilaniu


lamblin:
This should fix #486.

PR #526, opened 2017-09-13
Fix some memory leaks in libgpuarray

notoraptor:
Related to https://github.com/Theano/Theano/issues/6162 .

@nouiz @abergeron

### DIFF
--- a/pygpu/gpuarray.pyx
+++ b/pygpu/gpuarray.pyx
@@ -1428,6 +1428,7 @@ cdef GpuArray pygpu_reshape(GpuArray a, unsigned int nd, const size_t *newdims,
             tot *= newdims[i]
     cdims = <size_t *>calloc(nd, sizeof(size_t))
     if cdims == NULL:
+        free(cdims)
         raise MemoryError, "could not allocate cdims"
 
     cdef size_t d
@@ -1437,10 +1438,12 @@ cdef GpuArray pygpu_reshape(GpuArray a, unsigned int nd, const size_t *newdims,
             d = a.size // tot
 
             if d * tot != a.size:
+                free(cdims)
                 raise GpuArrayException, "..."
         cdims[i] = d
 
     array_reshape(res, a, nd, cdims, ord, nocopy)
+    free(cdims)
     return res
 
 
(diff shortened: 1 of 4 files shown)


pygpu/gpuarray.pyx
```diff
@@ -1428,6 +1428,7 @@ cdef GpuArray pygpu_reshape(GpuArray a, unsigned int nd, const size_t *newdims,
             tot *= newdims[i]
     cdims = <size_t *>calloc(nd, sizeof(size_t))
     if cdims == NULL:
+        free(cdims)
```
abergeron: If cdims is NULL then it can't be allocated.  This is useless.

notoraptor: Indeed.

notoraptor: Fixed.

pygpu/gpuarray.pyx
```diff
@@ -1437,10 +1437,12 @@ cdef GpuArray pygpu_reshape(GpuArray a, unsigned int nd, const size_t *newdims,
             d = a.size // tot
 
             if d * tot != a.size:
+                free(cdims)
                 raise GpuArrayException, "..."
         cdims[i] = d
 
     array_reshape(res, a, nd, cdims, ord, nocopy)
+    free(cdims)
```
abergeron: This whole free thing should probably be in a try: .. finally since there could be an exception at basically any point that will take us out of the function.

notoraptor: @abergeron I made a new commit to use try-finally block: https://github.com/Theano/libgpuarray/pull/526/commits/714f3a54e69e86ad7e71e22439bf0b06a328a2e9

pygpu/gpuarray.pyx
```diff
@@ -1430,18 +1430,21 @@ cdef GpuArray pygpu_reshape(GpuArray a, unsigned int nd, const size_t *newdims,
...
-    for i in range(nd):
-        d = newdims[i]
-        if i == caxis:
-            d = a.size // tot
-
-            if d * tot != a.size:
-                raise GpuArrayException, "..."
-        cdims[i] = d
+    try:
+        cdef size_t d
```
abergeron: This needs to stay outside of the try..finally

notoraptor: Fixed.

PR #528, opened 2017-09-14
CUDA race conditions 

dendisuhubdy:
HostFromGpu -> GpuArray_read() -> gpudata_read() -> cuda_read() -> cuMemcpyDtoHAsync()`returns 700 and then takes an exit path

```
Okay so after 3h of debugging, we know that in @shawntan's minimal testcase, `HostFromGpu -> GpuArray_read() -> gpudata_read() -> cuda_read() -> cuMemcpyDtoHAsync()` returns 700 and then takes an exit path that causes `GpuArray_read()` to return 3.

By @obilaniu

@shawntan @dendisuhubdy

Hey, just a random guess, should be easy to test.

Could you please, in current `master`, file `src/gpuarray_blas_cuda_cublas.c`, insert

    `GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(Ta, CUDA_WAIT_ALL));`

_immediately before_ the call to `cublasSgemmBatched()` at line 855 and `cublasDgemmBatched()` at line 982?

I can see a potential failure to upload the pointers "in time" on stream `mem_s`, and such a failure would be consistent with INVALID_ADDRESS being generated in cuBLAS code and asynchronously reported in HostFromGpu.
```

### DIFF
--- a/src/gpuarray_blas_cuda_cublas.c
+++ b/src/gpuarray_blas_cuda_cublas.c
@@ -835,8 +835,7 @@ static int sgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
       return ctx->err->code;
     }
 
-	// added cuda_wait
-	GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(Ta, CUDA_WAIT_ALL));
+    GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(Ta, CUDA_WAIT_READ));
 
     err = cublasSgemmBatched(h->h,
                              convT(transA), convT(transB),
@@ -965,8 +964,7 @@ static int dgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
       return ctx->err->code;
     }
 
-	// added cuda_wait
-	GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(Ta, CUDA_WAIT_ALL));
+    GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(Ta, CUDA_WAIT_READ));
 
     err = cublasDgemmBatched(h->h,
                              convT(transA), convT(transB),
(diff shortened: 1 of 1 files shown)


dendisuhubdy:
jenkins test this please

nouiz:
jenkins test this please

Can you confirm if this fix the user error or not?

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -835,6 +835,9 @@ static int sgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
       return ctx->err->code;
     }
 
+	// added cuda_wait
+	GA_CUDA_EXIT_ON_ERROR(ctx, cuda_wait(Ta, CUDA_WAIT_ALL));
```
abergeron: This should be WAIT_READ

src/gpuarray_blas_cuda_cublas.c
```diff
@@ -962,6 +965,9 @@ static int dgemmBatch(cb_order order, cb_transpose transA, cb_transpose transB,
...
```
abergeron: Same.

shawntan:
@nouiz Different errors are being seen on different machines now, but I'm unsure if they are caused by related bugs. 
One thing I've observed since patching was this: `pygpu.gpuarray.GpuArrayException: cuStreamWaitEvent(s, a->wev, 0): CUDA_ERROR_ILLEGAL_ADDRESS: an illegal memory access was encount`, and I'm not the only one.

Unfortunately these bugs are super hard to replicate and don't always happen consistently across machines.

src/gpuarray_blas_cuda_cublas.c
dendisuhubdy: I just did the change

src/gpuarray_blas_cuda_cublas.c
dendisuhubdy: Mem sous

dendisuhubdy:
So it looks like our potential patch fixed Stan Jastrzebski's problem but not @shawntan 's problem.

nouiz:
Note, I just merged in the master of libgpuarray another update that can help. Can @shawntan try it?

shawntan:
Yup. Seems to work so far, at least on bart15.

PR #529, opened 2017-09-14
Adding feature python setup.py clean

dendisuhubdy:
adding feature of python setup.py clean where it will delete the designated build files stated in .clean which is Build/ build/ and others


versioneer.py
```diff
@@ -1647,6 +1647,25 @@ def make_release_tree(self, base_dir, files):
...
+                    for filename in glob.glob(wildcard):
+                        try:
+                            os.remove(filename)
+                        except OSError:
+                            shutil.rmtree(filename, ignore_errors=True)
+
+            # It's an old-style class in Python 2.7...
+            distutils.command.clean.clean.run(self)
+    cmds["clean"] = cmd_clean
+
```
abergeron: Don't put that in versioneer, it's going to the lost the next time we update it.

.clean
```diff
@@ -0,0 +1,21 @@
...
+*.pyo
+*.egg-info
+dist
+setuptools*egg
+setuptools.pth
+distribute*egg
+distribute*tar.gz
+*.so
+*.o
+*.log
```
abergeron: Could this list be built into the code.  I'm not sure I like the idea of an external file list.

dendisuhubdy: Yes I could but it'll be ugly, one other choice is to load .gitignore than .clean

versioneer.py
dendisuhubdy: The thing is that the setup call at the end of `setup.py` uses cmds list from versioneer.

abergeron: You can get the cmdclass from versioneer, add the custom clean command and then pass it to setup().

Issue #541, opened 2017-09-29
Elemwise kernels on discontiguous arrays sometimes produce garbage


kohr-h:
(This may be due to my lack of understanding the internals of elemwise kernels.)
When elemwise kernels receive all of the following:
- more than 3 many arguments
- floating point constants
- discontiguous arrays

then the results are garbage. I created [a script](https://gist.github.com/kohr-h/d6e7de79eaef2ee1c12b014a85b50935) to check various scenarios.

For example:
```python
def lico(a, x, b, y, z):
    a = np.asarray(a)
    b = np.asarray(b)
    a_arg = pygpu.elemwise.as_argument(a, 'a', read=True)
    x_arg = pygpu.elemwise.as_argument(x, 'x', read=True)
    b_arg = pygpu.elemwise.as_argument(b, 'b', read=True)
    y_arg = pygpu.elemwise.as_argument(y, 'y', read=True)
    z_arg = pygpu.elemwise.as_argument(z, 'z', read=True, write=True)
    args = [z_arg, a_arg, x_arg, b_arg, y_arg]
    oper = 'z = a * x + b * y'
    k = pygpu.elemwise.GpuElemwise(y.context, oper, args,
                                   convert_f16=True)
    k(z, a, x, b, y, broadcast=False)
    return z


a = 1.1
b = 3
x = pygpu.array(np.arange(12).reshape(3, 4))
y = pygpu.array(np.ones((3, 4)))
z = pygpu.array(2 * np.ones((3, 4)))
x, y, z = x[:, ::2], y[:, ::2], z[:, ::2]
npy_res = a * np.asarray(x) + b * np.asarray(y)
print('before: z=')
print(z)
lico(a, x, b, y, z)
print('after: z=')
print(z)
print('numpy result:')
print(npy_res)
```
produces
```
before: z=
[[ 2.  2.]
 [ 2.  2.]
 [ 2.  2.]]
after: z=
[[  1.48219694e-323   2.20000000e+000]
 [  4.40000000e+000   6.60000000e+000]
 [  8.80000000e+000   1.10000000e+001]]
numpy result:
[[  3.    5.2]
 [  7.4   9.6]
 [ 11.8  14. ]]
```

Another curious and possibly related issue occurs when the arguments to the kernel are reordered, and here the error is triggered already for 3 arguments (not for 2 though). Consider this variant
```python
def axpy(a, x, y):
    a = np.asarray(a)
[...]

PR #545, opened 2017-10-04
Fix wrong index in the generation code for scalars.

abergeron:


abergeron:
Fixes #541

nouiz:
As this fix wrong values, I think we should make a release rapidly.

Is there other pending wrong values issues?

Also, I forgot that this PR miss test. Can you add that in another PR?

PR #546, opened 2017-10-05
Test for the fix in #545

abergeron:


tests/check_elemwise.c
```diff
@@ -434,6 +434,60 @@ START_TEST(test_basic_scalar) {
...
+  static const int32_t data1[4] = {0, 1, 2, 3};
+  static const float data2[4] = {2.0, 2.0, 2.0, 2.0};
+  float data3[4];
+
+  size_t dims[2] = {2, 2};
+
+  gpuelemwise_arg args[3] = {{0}};
+  void *rargs[3];
+
+  ga_assert_ok(GpuArray_empty(&x, ctx, GA_UINT, 2, dims, GA_C_ORDER));
```
nouiz: data1 is declared int32_t, while here you set uint.

abergeron: Fixed.

PR #553, opened 2017-10-26
add CUDA 9 version when trying to load libnvrtc

pambros:
With the last CUDA SDK and cuDriverGetVersion, I get major=9 and minor=1 but nvrtc64_91.dll does not exist. It is still nvrtc64_90.dll. So the code fails to load the lib and look for other versions and CUDA 9 was not in the list. Now it is.

### DIFF
--- a/src/gpuarray_buffer_cuda.c
+++ b/src/gpuarray_buffer_cuda.c
@@ -176,7 +176,7 @@ static int setup_lib(void) {
     res = load_libnvrtc(major, minor, global_err);
     if (res != GA_NO_ERROR) {
       /* Else, let's try to find a nvrtc corresponding to supported CUDA versions. */
-      int versions[][2] = {{8, 0}, {7, 5}, {7, 0}};
+      int versions[][2] = {{9, 0}, {8, 0}, {7, 5}, {7, 0}};
       int versions_length = sizeof(versions) / sizeof(versions[0]);
       int i = 0;
       /* Skip versions that are higher or equal to the driver version */


nouiz:
I don't find cuda 9.1 on the web. Where did you took that version?

pambros:
With cuda_9.0.176_windows.exe and cuDriverGetVersion(&tmp), I get tmp=9010.
So in the code:
```
major = tmp / 1000;
minor = (tmp / 10) % 10;
```
I have major=9 and minor=1
I don't know if we should have instead minor = (tmp / 100) % 10; but right now the code looks for nvrtc64_91.dll and it does not exist.

TalhaAsmal:
When compiled with the debug flag, trying to initialize the GPU using PyGPU results in the following error:

```
Using Theano backend.
Loading nvrtc 9.1.
ERROR 18: Could not load "nvrtc64_91.dll": The specified module could not be found.

Loading nvrtc 8.0.
ERROR 18: Could not load "nvrtc64_80.dll": The specified module could not be found.

Loading nvrtc 7.5.
ERROR 18: Could not load "nvrtc64_75.dll": The specified module could not be found.

Loading nvrtc 7.0.
ERROR 18: Could not load "nvrtc64_70.dll": The specified module could not be found.

ERROR (theano.gpuarray): Could not initialize pygpu, support disabled
```

I confirm the proposed change fixes this and allows the GPU to be initialized successfully. After making the proposed change, the debug output is as follows:

```
Using Theano backend.
Loading nvrtc 9.1.
ERROR 18: Could not load "nvrtc64_91.dll": The specified module could not be found.

Loading nvrtc 9.0.
Loading cuBLAS 9.0.
```

nouiz:
I think we should make a new release with this. Especially as we should
make Theano 1.0 final tomorow. More people will update and we could see
more of this problem happening. I prefer fewer questions.

PR #558, opened 2017-12-14
Add 9.1 to the list of versions that windows looks for.

abergeron:
This is the only change required to support cuda 9.1.  It's mostly a bugfix release with regards to 9.0.


nouiz:
travis failed. Otherwise, I'm good with this PR.

abergeron:
Travis is drunk.  I think we can safely ignore it.

nouiz:
I think the PR is good, but merging it as is will make people think it isn't stable due to the failure. Can you look at fixing it or removing it?

Issue #560, opened 2017-12-19
Better error message for driver too old with CUDA 9.1


nouiz:
To help us lower user support we need a better error message. Currently with driver good enought for CUDA 9.0 (384.90), you get this error when tring to use CUDA 9.1:

```
$ THEANO_FLAGS=device=cuda python -c "import theano"
Using cuDNN version 7002 on context None
ERROR 11: cublasCreate: (cublas) Library not initialized.
ERROR (theano.gpuarray): Could not initialize pygpu, support disabled
Traceback (most recent call last):
  File "/u/bastienf/repos/theano/gpuarray/__init__.py", line 227, in <module>
    use(config.device)
  File "/u/bastienf/repos/theano/gpuarray/__init__.py", line 214, in use
    init_dev(device, preallocate=preallocate)
  File "/u/bastienf/repos/theano/gpuarray/__init__.py", line 159, in init_dev
    pygpu.blas.gemm(0, tmp, tmp, 0, tmp, overwrite_c=True)
  File "pygpu/blas.pyx", line 149, in pygpu.blas.gemm (pygpu/blas.c:3538)
  File "pygpu/blas.pyx", line 47, in pygpu.blas.pygpu_blas_rgemm (pygpu/blas.c:2028)
GpuArrayException: ('cublasCreate: (cublas) Library not initialized.', 11)

$ nvidia-smi 
Tue Dec 19 14:57:15 2017       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 384.90                 Driver Version: 384.90                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  GeForce GT 520      Off  | 00000000:01:00.0 N/A |                  N/A |
| N/A   42C    P0    N/A /  N/A |    347MiB /   964MiB |     N/A      Default |
+-------------------------------+----------------------+----------------------+
|   1  GeForce GTX TIT...  Off  | 00000000:02:00.0 Off |                  N/A |
| 22%   37C    P8    14W / 250W |      2MiB / 12207MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
[...]

PR #561, opened 2018-01-08
Add possible cause for the Library not initialized error message from cublas

abergeron:
Fix for #560


src/gpuarray_blas_cuda_cublas.c
```diff
@@ -53,7 +53,8 @@ static const char *estr(cublasStatus_t err) {
 
 static inline int error_cublas(error *e, const char *msg, cublasStatus_t err) {
   return error_fmt(e, (err == CUBLAS_STATUS_ARCH_MISMATCH) ? GA_DEVSUP_ERROR : GA_BLAS_ERROR,
-                   "%s: %s", msg, estr(err));
+                   "%s: %s %s", msg, estr(err),
```
nouiz: This add an extra space when the old message is used. Can you remove the space on this line and add it in the string bellow?

abergeron: done

PR #563, opened 2018-01-10
Small doc update and re-raise an error.

nouiz:


pygpu/gpuarray.pyx
```diff
@@ -634,15 +634,22 @@ def init(dev, sched='default', single_stream=False, kernel_cache_path=None,
...
+
         if kernel_cache_path:
             kernel_cache_path_b = _s(kernel_cache_path)
             gpucontext_props_kernel_cache(p, <const char *>kernel_cache_path_b)
-        gpucontext_props_alloc_cache(p, initial_cache_size, max_cache_size)
+
+        err = gpucontext_props_alloc_cache(p, initial_cache_size,
+                                           max_cache_size)
+        if err != GA_NO_ERROR:
+            raise get_exc(err), gpucontext_error(NULL, err)
```
abergeron: I'm pretty sure that those two methods will never return any error, but the code checking for it doesn't hurt.

PR #571, opened 2018-04-05
pygpu/tests: Skip tests that require CUDA context if the context is not CUDA

### LINKED ISSUE #491: Skip CUDA specific tests if DEVICE is not CUDA compatible
I have successfully enabled integration tests for the packaged version of `libgpuarray` on Debian. The `pygpu` test suite is run on `pocl`, a CPU implementation of OpenCL, with most tests [passing](http://debomatic-amd64.debian.net/distribution#unstable/libgpuarray/0.6.9-2/autopkgtest).

However, the test suite attempts to run CUDA specific tests although DEVICE is set to `opencl0:0`, resulting in many errors of the form:
```
TypeError: This is for CUDA arrays.
```

Please consider adapting the test suite so that those CUDA tests aren't performed if the selected device has got no CUDA compute. This should be achievable via something like `pytest` [skipif](https://docs.pytest.org/en/latest/skipping.html#id1).

jvesely:
Fixes #491
Signed-off-by: Jan Vesely <jan.vesely@rutgers.edu>

### DIFF
--- a/pygpu/tests/test_gpu_ndarray.py
+++ b/pygpu/tests/test_gpu_ndarray.py
@@ -195,7 +195,7 @@ def asfortranarray(shp, dtype, offseted_outer, offseted_inner, sliced, order):
 
     # numpy upcast with a view to 1d scalar.
     if gpu.flags['F_CONTIGUOUS']:
-        assert b.gpudata == gpu.gpudata
+        assert ctx.kind != b'cuda' or b.gpudata == gpu.gpudata
     elif (sliced != 1 or shp == () or (offseted_outer and len(shp) > 1) or
           (order != 'f' and len(shp) > 1)):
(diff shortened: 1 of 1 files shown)


jvesely:
@nouiz, are you interested in bug fixes?

abergeron:
Yes we are, however I would fix this differently.  The only part of those tests that require a cuda context is the .gpudata access.  You could make that part conditional on the context kind and the rest should pass.

nouiz:
@abergeron this PR was updated. Can you review it again?
@jvesely when you update a commit and force push, this do not send a notification. So if you can make  comment, it will make us see the change faster.

jvesely:
thanks. I didn't know new commits were silent.

PR #576, opened 2018-10-20
add float64 support to tril triu

wonghang:
add float64 support to pygpu tril() triu()
they are used in https://github.com/Theano/Theano/blob/master/theano/gpuarray/linalg.py

### DIFF
--- a/pygpu/tests/test_basic.py
+++ b/pygpu/tests/test_basic.py
@@ -7,31 +7,33 @@ import numpy
 
 
 def test_tril():
-    for shape in [(10, 5), (5, 10), (10, 10)]:
-        for order in ['c', 'f']:
-            for inplace in [True, False]:
-                ac, ag = gen_gpuarray(shape, 'float32',
-                                      order=order, ctx=context)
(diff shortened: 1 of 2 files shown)


pygpu/tests/test_basic.py
```diff
@@ -7,31 +7,33 @@
...
-                else:
-                    assert numpy.all(ac == ag)
+    for dtype in ['float32','float64']:
+        for shape in [(10, 5), (5, 10), (10, 10)]:
+            for order in ['c', 'f']:
+                for inplace in [True, False]:
+                    ac, ag = gen_gpuarray(shape, dtype,
+                                          order=order, ctx=context)
+                    result = triu(ag, inplace=inplace)
+                    assert numpy.all(numpy.triu(ac) == result)
```
nouiz: Maybe assert that results have the right dtype? The current test could pass if the value have the wrong dtype I think. Otherwise, it loooks good. Thanks.

abergeron:
I think we can do better tests later, for now the code uses the right dtype so I'm fine with it.

PR #580, opened 2018-11-06
Support more recent cuda versions 

abergeron:
Also make sure to not always report errors for nvrtc64_70

### DIFF
--- a/src/gpuarray_buffer_cuda.c
+++ b/src/gpuarray_buffer_cuda.c
@@ -176,7 +176,7 @@ static int setup_lib(void) {
     res = load_libnvrtc(major, minor, global_err);
     if (res != GA_NO_ERROR) {
       /* Else, let's try to find a nvrtc corresponding to supported CUDA versions. */
-      int versions[][2] = {{9, 1}, {9, 0}, {8, 0}, {7, 5}, {7, 0}};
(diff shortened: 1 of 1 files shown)


abergeron:
This should get some testing on windows before it's merged.

nouiz:
jenkins test windows

abergeron:
Jenkins is fine, but I meant test on a windows machine that has cuda10 installed.

nouiz:
jenkins didn't ran... It have problems.
I'll test locally here tomorrow.

src/gpuarray_buffer_cuda.c
```diff
@@ -190,7 +190,8 @@ static int setup_lib(void) {
       } while (res != GA_NO_ERROR && i < versions_length);
     }
     if (res != GA_NO_ERROR)
-      return res;
+      // Return the error from the original attempt
+      return load_libnvrtc(major, minor, global_err);
```
nouiz: This is not the original major/minor that you try again to generate the original error. But the guessed major/minor.

abergeron: That's true.

abergeron: Should be fixed now.

nouiz:
as jenkins isn't working, I tested locally and it work. So merging.

nouiz:
Note, a new release could lower the number of questions we have.

Issue #581, opened 2018-12-24
gpuarray crashes with ARM Mali GPU


Kishwar:
Hello Theano team,

I am trying to load Theano/Keras model, I am getting crash on gpuarray. It happens also even when I use GPU for training. Same code works OK for CPU.

odroid@odroid:~/Desktop/share/thMNIST$ sudo THEANO_FLAGS=device=opencl0:1 python3 th__mnist.py
**Using Theano backend.
Mapped name None to device opencl0:1: Mali-T628** 
Traceback (most recent call last):
  File "th__mnist.py", line 50, in <module>
    model.fit(X_train, y_train, validation_data=(X_test, y_test), epochs=10, batch_size=200)
  File "/usr/local/lib/python3.6/dist-packages/keras/engine/training.py", line 1010, in fit
    self._make_train_function()
  File "/usr/local/lib/python3.6/dist-packages/keras/engine/training.py", line 519, in _make_train_function
    **self._function_kwargs)
  File "/usr/local/lib/python3.6/dist-packages/keras/backend/theano_backend.py", line 1397, in function
    return Function(inputs, outputs, updates=updates, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/keras/backend/theano_backend.py", line 1383, in __init__
    **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/theano/compile/function.py", line 317, in function
    output_keys=output_keys)
  File "/usr/local/lib/python3.6/dist-packages/theano/compile/pfunc.py", line 486, in pfunc
    output_keys=output_keys)
  File "/usr/local/lib/python3.6/dist-packages/theano/compile/function_module.py", line 1841, in orig_function
    fn = m.create(defaults)
  File "/usr/local/lib/python3.6/dist-packages/theano/compile/function_module.py", line 1715, in create
    input_storage=input_storage_lists, storage_map=storage_map)
  File "/usr/local/lib/python3.6/dist-packages/theano/gof/link.py", line 699, in make_thunk
    storage_map=storage_map)[:3]
  File "/usr/local/lib/python3.6/dist-packages/theano/gof/vm.py", line 1091, in make_all
    impl=impl))
  File "/usr/local/lib/python3.6/dist-packages/theano/gof/op.py", line 952, in make_thunk
    compute_map=compute_map, impl='c')
  File "/usr/local/lib/python3.6/dist-packages/theano/gpuarray/elemwise.py", line 2519, in prepare_node
[...]

PR #582, opened 2019-01-13
Fix 581 (crash on 32bit) + spelling fixes

rebecca-palmer:
cl_property uses a size_t to request some clGetDeviceInfo memory sizes that are actually cl_ulong (always 64 bit).  On systems where size_t is 32 bit, this fails with CL_INVALID_VALUE (#581 reports this from ARM, but it can also happen on 32 bit x86).

This fix clips the values to fit in a size_t to avoid breaking ABI, but it might actually be better to switch to returning a cl_ulong.

The other commit is some spelling errors mostly found by Debian's lintian.


abergeron:
I'm not sure how relevant the support for 32-bit is, but I agree with the fix, so ok.

PR #584, opened 2019-01-27
Miscellaneous bug fixes (including a stack smashing crash)

rebecca-palmer:
Two crashes/exceptions, an intermittent test failure, a typo, and a code to user-friendly error table that wasn't actually being used: see commit messages for details.

PR #585, opened 2019-01-27
Give a user-friendly error when double precision is not supported

rebecca-palmer:
i.e. one that actually says that, not a generic "Could not initialize C GpuElemwise instance" or a code that means nothing until looked up.

Also extend "unsupported = skip, not fail" to all tests that use double precision.

There are two places where this reads error message text to determine the error type because the error code isn't passed up the stack: making it be so passed would be more robust, but more interface-breaking.

Motivation / testing note: Intel integrated GPUs are probably the most readily available "no double precision" hardware (the [beignet-opencl-icd](https://packages.debian.org/unstable/beignet-opencl-icd) driver disables it by default on all of them, I think Ivybridge/Haswell lack it at the hardware level).  Unfortunately, they may crash in the BLAS tests (with or without these changes) due to clMathLibraries/clblas#322 and/or clMathLibraries/clblas#321: skip those tests if necessary.

(This is branched off #584 because they touch the same line of code so would merge conflict: they are semantically independent.)

### DIFF
--- a/pygpu/_elemwise.pyx
+++ b/pygpu/_elemwise.pyx
@@ -15,6 +15,11 @@ cdef bytes to_bytes(s):
       return <bytes>(<unicode>s).encode('ascii')
   raise TypeError("Can't convert to bytes")
 
+cdef extern from "gpuarray/buffer.h":
+    ctypedef struct gpucontext:
+        pass
+    char *gpucontext_error(gpucontext *ctx, int err)
+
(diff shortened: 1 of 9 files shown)


abergeron:
This looks good, thanks.

abergeron:
By the way @rebecca-palmer are you interested in commit access to this repo?

You've done a number of great maintenance/bugfix PRs so I would be glad to have you working on this.

